PR #1, opened 2020-04-22
[Spike] Design System

ethax-ross:
Screenshot 2020-04-22 at 15 04 02

### Context

The Get into Learning website will be a different design to the main GOV UK website, therefore we will not be leveraging the GOV UK design system. Instead, we are opting to create a new design system for the website that will give clarity and ease of use to the content designers when publishing content.

Initially, we have opted to keep the whole website inside this single Rails app, however we can envisage a time further down the line where it may make sense to split the design system out into it's own repository/application (and with it moving the majority of website pages to a static website generator). For that reason, the design system should not be tightly coupled to the Rails ecosystem.

The content designers will be working exclusively in Markdown; where it makes sense they should be able to leverage view components using custom Markdown formatting (similar to [GovSpeak](https://github.com/alphagov/govspeak) and ideally using the exact same Markdown syntax where possible).

Adds a design system living under `/lib/design_system`. The view components are made up of:

- _style.scss - styling, using class names consistent with the GOV UK design system
- template.haml - a partial containing the button Haml (HTML)
- button_controller.js - StimulusJS controller for handling dynamic interactions (can be omitted if not necessary)
- helper.rb - ruby helpers to render the component partial (providing a clean interface for the component and a way to render the component in place of it's Markdown tag)
- README.html.md - description of the component, usage examples, code snippets, etc

Haml is chosen over HTML simply because it's more concise when you start to apply conditional logic to HTML tags (such as disabling, class names, data attributes etc). The helper provides a way of defining the interface for a component, and something to call when we are mapping a bit of Markdown to a certain component. 

The component README contains examples, descriptions, use cases, etc and is designed to be pulled into a larger 'design system website' for browsing. The default Markdown renderer has been extended to support syntax highlighting with highlight.js (the GOV UK design system appears to use this). At the minute I'm pulling in all the READMEs and serving the design system 'website' as part of Rails - it's a lightweight controller/views and we could them to a Sinatra app or something if the design system was to be pulled out of Rails later on.

### Guidance to review

- Extending Markdown with custom tags is going to be quite time consuming (I've done a very simple button extension in this PR, but others are going to be complex). I wonder if we should try and get all the GovSpeak ones for free by mimicking their CSS class names and just using their Markdown extension directly? (we could also wrap it and transform the class names to continue using our own `git-` prefix). If we are writing custom Markdown extensions I would also opt to move the parsing logic into the individual component directory as well to be consistent.
- I haven't added tests for the design system components, but I would expect to use RSpec (or similar) for the helpers and Jest for the Stimulus controllers. We could also look at snapshot testing, though I don't know what the support for this is like in the Ruby world.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -24,6 +24,8 @@ module GovukRailsBoilerplate
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 5.2
 
+    config.eager_load_paths << Rails.root.join("lib")
+
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,7 @@ gem "canonical-rails"
 
 gem 'redcarpet'
 gem 'emd'
+gem 'haml'
 
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -90,6 +90,9 @@ GEM
     foreman (0.87.1)
     globalid (0.4.2)
       activesupport (>= 4.2.0)
+    haml (5.1.2)
+      temple (>= 0.8.0)
+      tilt
     i18n (1.8.2)
       concurrent-ruby (~> 1.0)
     jaro_winkler (1.5.4)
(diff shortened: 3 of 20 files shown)


config/application.rb
```diff
@@ -24,6 +24,8 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 5.2
 
+    config.eager_load_paths << Rails.root.join("lib")
```
jebw, line 27: One really fussy nitpick might be to just eager-load files under `lib/design_system` instead of `lib`

ethax-ross: Yeah agreed; though when I did this earlier it broke my name spacing, I think I would have to remove `DesignSystem::` from the namespaces for it to work that way, but I'll take another look 👍

jebw: Ah, see your point - might take some lateral thinking

PR #3, opened 2020-05-04
Re org

DrLazer:
Moved a few bits and pieces into their own components

Navigation for desktop
Navigation for mobile


app/views/layouts/_footer.html.erb
```diff
@@ -0,0 +1,21 @@
+<%# <footer>
```
ethax-ross, line 1: Can we just delete this layout instead of commenting it out, or is it coming to be making a comeback soon?

app/views/layouts/application.html.erb
```diff
@@ -10,69 +11,20 @@
     <%= tag :link, rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' %>
     <%= stylesheet_pack_tag 'application', 'data-turbolinks-track': 'reload', media: 'all' %>
     <%= javascript_pack_tag 'application', 'data-turbolinks-track': 'reload', defer: true %>
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
```
ethax-ross, line 14: Is it worth bundling this as an npm dependency instead?

app/views/layouts/_footer.html.erb
DrLazer: yeah, like within 4 hours

PR #7, opened 2020-05-05
Footer

DrLazer:
footer section

desktop and mobile


DrLazer:
@ethax-ross 100% that's the plan.
I blasted this out as fast as possible so they had something available for Wednesday 2pm (cut-off for UR)
But I've got a lot of tidying up to do now

PR #8, opened 2020-05-06
Initial home=page tweaks

DrLazer:
quick fixes and updates to the design

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -1,11 +1,13 @@
 <nav class="navbar">
     <div class="navbar-right">
         <ul>
-            <li><a href="/">Life as a teacher</a></li>
+            <li><a href="/">Home</a></li>
             <li><a href="/">Steps to become a teacher</a></li>
             <li><a href="/">Funding your training</a></li>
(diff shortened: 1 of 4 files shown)


app/views/layouts/_header.html.erb
```diff
@@ -15,13 +17,22 @@
...
+            <li><a href="/">Steps to become a teacher</a></li>
+            <li><a href="/">Funding your training</a></li>
+            <li><a href="/">Teaching as a career</a></li>
+            <li><a href="/">Salaries and Benefits</a></li>
+            <li><a href="/">Find an event near you</a></li>
+            <li><a href="/">Talk to us</a></li>
         </div>
     </div>
+    <div class="navbar-mobile-right navbar-mobile-right-label">
+        <a href="javascript:void(0);" class="icon" onclick="navToggle()">
```
ethax-ross, line 30: Again, one for later - but I think we should use StimulusJS for sprinkles of JS. I've done a spike on how we could put together the design system inside the Rails app; might be worth a look at some point (https://github.com/DFE-Digital/get-into-teaching-app/pull/1). There's also a write up of that spike in a Google Docs under the `BETA` folder somewhere.

PR #19, opened 2020-05-11
Container 1200

DrLazer:
white content sections to have a max size of 1200px
introduction of separated sass files
colour pallete
replace map image
add a map marker pin
tweaks to the header


app/webpacker/styles/application.scss
```diff
@@ -14,6 +14,9 @@
     font-style: normal;
 }
 
+@import './colours.scss';
```
ethax-ross, line 17: Can we use the American spelling? 🙈

PR #21, opened 2020-05-12
Fe clean up

DrLazer:
a lot of reorganisation of sass for reuse


app/views/layouts/_footer.html.erb
```diff
@@ -1,12 +1,12 @@
 <footer>
-    <div class="footer-strip-one">
+    <div class="footer-top">
         <a href="/">
-            <div class="logo">
+            <div class="logo__image">
                 <img src="<%= asset_pack_path 'media/images/getintoteachinglogo.svg' %>" />
-                <div class="tagline">Every Lesson shapes a life.</div>
+                <div class="logo__image__tagline">Every Lesson shapes a life.</div>
```
ethax-ross, line 6: What convention are you using for class names? It looks like a bit of a mix between these:

```
logo__image__tagline
footer-bottom__logo-container
call-to-action-button
```

DrLazer: BEM
hyphens are part of a block name
__ double underscores separate elements
-- double hyphens are modifiers, but there aren't that many of them

ethax-ross: Ah I see, the single hyphen was throwing me but you're just using that for block/element names with multiple words in 👍

ethax-ross:
Looks good, though CI is failing and it may be worth fixing up your commits while you're at it?

PR #27, opened 2020-05-18
Events listing design

DrLazer:
design changes and tweaks to the events listing pages
plus incorporation of custom icons
replaced the hamburger FA icons with custom ones

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -18,9 +18,7 @@
     </div>
     <div class="navbar__mobile">
         <a href="javascript:void(0);" class="icon" onclick="navToggle()">
-            <span class="navbar__mobile__hamburger">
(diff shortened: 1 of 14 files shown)


app/webpacker/styles/text.scss
```diff
@@ -75,7 +84,7 @@ a {
     }
 
     .strapline-prominent {
-        font-size: 20px;
+        font-size: 19px;
```
ethax-ross, line 87: Might want to use `rem` for font sizes instead of `px` for accessibility

app/views/layouts/_header.html.erb
```diff
@@ -55,15 +53,15 @@
 <script>
     function navToggle() {
```
ethax-ross, line 54: We should move this into a Stimulus controller at some point

PR #29, opened 2020-05-19
Feature/hero path

DrLazer:
helper method to determine hero path

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -6,7 +6,11 @@ class PagesController < ApplicationController
     render template: "pages/events"
   end
 
+  def content
+    render template: "pages/content"
+  end
+
   def show
-    render template: "content/#{params[:page]}"
+    render template: "content/#{params[:page]}", layout: 'layouts/content'
   end
 end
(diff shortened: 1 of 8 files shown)


app/helpers/application_helper.rb
```diff
@@ -16,4 +16,13 @@ def header_image
     # FIXME needs to parse frontpatter
     "media/images/main-header.jpg"
   end
+
+  def hero_image_path(imgpath)
+    if imgpath =~ %r(^/assets/)
+        imgpath
```
jebw, line 22: Guessing the linter may complain about the extra whitespace?

config/routes.rb
```diff
@@ -7,5 +7,6 @@
 
   get "/events", to: "pages#events", via: :all
 
+  #get "*page", to: "pages#content", as: :page
```
jebw, line 10: This can be dropped I guess

app/controllers/pages_controller.rb
```diff
@@ -6,7 +6,11 @@ def events
     render template: "pages/events"
   end
 
+  def content
```
jebw, line 9: This is now redundant isn't it?

app/helpers/application_helper.rb
jebw: Turns out the linter doesn't care :)

DrLazer: lol

PR #32, opened 2020-05-21
Feature/individual events page

DrLazer:
individual events page and componentised eventbox listings

### DIFF
--- a/app/views/components/_eventbox.html.erb
+++ b/app/views/components/_eventbox.html.erb
@@ -10,8 +10,6 @@
         <p>
             <%= description %>
         </p>
-
-        
         <div style="<%=  traintoteach ? '' : 'display:none'%>"><h3><div class="event-resultbox__content__icon icon-train-to-teach " ></div> Train to teach event</h3></div>
         <div style="<%=  online ? '' : 'display:none'%>"><h3><div class='event-resultbox__content__icon icon-online'></div> Online event</h3></div>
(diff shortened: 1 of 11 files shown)


app/views/components/_eventbox.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    </div>
+    <div class="event-resultbox__datetime">
+        <%= datetime %>
+    </div>
+    <div class="<%=  traintoteach ? 'event-resultbox__greenstripe' : 'event-resultbox__bluestripe' %>"></div>
+    <div class="event-resultbox__content">
+        <p>
+            <%= description %>
+        </p>
+        <div style="<%=  traintoteach ? '' : 'display:none'%>"><h3><div class="event-resultbox__content__icon icon-train-to-teach " ></div> Train to teach event</h3></div>
```
ethax-ross, line 13: These types of conditional styles could be more concise, e.g. `<%= "display: none" unless traintoteach %>`

app/webpacker/styles/components/eventbox.scss
```diff
@@ -0,0 +1,60 @@
+
+.event-resultbox {
```
ethax-ross, line 2: Looks like there's a fair amount of duplication with eventboxshort, it may be cleaner as a modifier just overwriting the different attributes? `eventbox--short`

DrLazer: yep, absolutely, but this design is to be replaced so was just a quick hack.
if it remains ill change it

PR #34, opened 2020-05-22
google analytics added

goodviber:
### JIRA ticket number
Request from Steven Fawcett
Adds google analytics (gtag) to app
### Guidance to review
Check is working via Google Analytics


.travis.yml
```diff
@@ -0,0 +1,26 @@
+language: ruby
```
jebw, line 1: I don't think this file should be being added? I'd previously removed a travis config because it was breaking CI.

 Is there a reason to add it back in or is this just a merging problem?

app/views/layouts/application.html.erb
```diff
@@ -11,6 +21,7 @@
     <%= tag :link, rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' %>
     <%= stylesheet_pack_tag 'application', 'data-turbolinks-track': 'reload', media: 'all' %>
     <%= javascript_pack_tag 'application', 'data-turbolinks-track': 'reload', defer: true %>
+    <%= javascript_pack_tag 'google_analytics' %>
```
jebw, line 24: I don't think this needs to be off in its own pack file? It will add another roundtrip to the server with associate latency for 7 lines of JS? Could it go in the application pack file instead?

app/views/layouts/application.html.erb
```diff
@@ -2,6 +2,16 @@
...
 
   <head>
+  <% if Rails.env.production? %>
+  <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-159214613-2"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+      gtag('config', 'UA-159214613-2');
```
jebw, line 12: Can the token be replaced with a config value, probably from outside the app - maybe an env var? Would be worth talking through with @sfawcett123 how best to handle this?

The `<% if Rails.env.production?` could then be replaced with something like `<% if ENV['GA_ID'].present?` and could be test in development by setting the appropriate env var?

sfawcett123:
I have wrapped the PAAS Manifest.yml as a jinja2 template, and can now pass secrets through as environment variables or GitHub secrets, so we can store them as GitHub secrets foreach project.

The current issue I am trying to consider is how to set environment variables that differer from environment to environment, i.e. Test / Production 

But I thought I might be able to solve that using Azure Secrets Management  now we have a an Azure account for the project.  But that’s a little down the road

app/views/layouts/application.html.erb
goodviber: Sounds good

.travis.yml
goodviber: It was a merge, I just did the minimum and commented out the failing task to get the build to pass, Craig did suggest we could just sack the whole file.

PR #43, opened 2020-05-29
Feature/event registration

DrLazer:
- markup for the remaining steps of the sign up for an event process
- installation of the gov frontend to use those controls on the forms


config/routes.rb
```diff
@@ -8,6 +8,9 @@
   get "/eventschool", to: "pages#eventschool", via: :all
   get "/event", to: "pages#event", via: :all
   get "/event/register/1", to: "pages#eventregistration", via: :all
+  get "/event/register/2", to: "pages#eventregistration2", via: :all
```
ethax-ross, line 11: This could be changed to a single route/action:

```
get "/event/register/:step_number", to: "pages#eventregistration", via: all
```
You could then do something like this in the action:

```
def eventregistration
  render template: "pages/events/registration/step#{params[:step_number]}"
end
```

app/views/pages/events/registration/step2.html.erb
```diff
@@ -0,0 +1,25 @@
...
+    <div class="event-reg-main__content container-1000">
+        <div class="event-reg-main__content__left">
+            <h1>Sign up for this event:</h1>
+            <h2>Train to Teach Coventry online event</h2>
+            <p></p>
+
+            <div class="govuk-form-group">
+                <div class="fieldlabel">Phone number (optional)</div>
+                <p class="fieldhelp">This helps us contact you with updates to this event and other important information.</p>
+                <input class="govuk-input" id="phone-number" name="phone-number" type="text">
```
ethax-ross, line 11: May be worth checking if you can use specific inputs, like input type `tel` here as it'll have a nicer experience especially on mobile it should show the number pad.

config/routes.rb
DrLazer: i needed this info, but couldnt find it :)

app/views/pages/events/registration/step2.html.erb
DrLazer: ill make a note and check the gov style docs

PR #66, opened 2020-06-09
imitate the content style layout on the events page

DrLazer:
imitate the content style layout on the events page

### DIFF
--- a/app/webpacker/styles/events.scss
+++ b/app/webpacker/styles/events.scss
@@ -1,49 +1,44 @@
 .types-of-event {
 
+    vertical-align: top;
     text-align: left;
-    padding-top: 20px;
-    padding-bottom: 10px;
+    width: 100%;
 
-    &__header {
-        margin-top: 30px;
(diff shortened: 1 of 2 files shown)


app/webpacker/styles/events.scss
```diff
@@ -1,49 +1,44 @@
...
-    &__header {
-        margin-top: 30px;
-        margin-bottom: 0px;
-        margin-left:20px;
-    }
+    &__left {
 
-    &__content {
+        box-sizing: border-box;
         padding-left: 40px;
```
ethax-ross, line 10: Can be one for later -- you could combine these into a singe `padding` statement -- might be worth sticking a prettier and a linter to take care of these for you

DrLazer: well ... personally i like them explicitly written out.  almost everyone I ask is unable to tell me what order they are in.  Top, Right, Bottom, Left.  Is it? :)

PR #73, opened 2020-06-15
GITPB-245 Base classes for event wizard

jebw:
### JIRA ticket number

GITPB-245

### Context

Within the Get into Teaching app we have a couple of wizard processes, Apply for an Event and the Join Newsletter.

These classes add a basic framework for implementing the wizard

1. Added base classes for the data store, the wizard and the steps
2. Added an implementation for the first screen of the event application wizard
3. Added a common route and controller for the Event Application wizard

### Design notes

The `Store` class is abstracted out, even though it currently just wraps session to allow it to be replaced with alternatives in the future - eg use Redis as the backing store.

The `Step` 's implement each of the screens, are are effectively basic ActiveModels. 

The `Wizard` simply controls the flow between screens. 

A single controller is used to access a wizard with `params[:id]` being used to determine the step and load the appropriate form from the view.

Adding branching should not affect controllers because the flow logic is held within the `Step` and `Wizard` classes.

#### Future work

**Conditionals**

`Step`'s do not yet implement logic for whether they step should be being shown (ie conditional) but it is intended to add a `active?` method to `Step`, defaulting to `true`. 

This will allow implementing a Wizard as a linear flow but with steps that can be skipped. The actual logic for `active?` will have access to the full `Store` hash so can do things like 'this is active if the user said yes to a question 2 screens back`.

`#active?` will also allow determining whether to include the step in the serialization at the end when uploading data to the API.

Currently `Wizard`'s `#next_step`, `#previous_step` operate in a simple linear manner but the PR for the `#active?` method on the `Step` class will update `#next_step`/`#previous_step` to take `#active?` into account.

**API Integration**

This does not currently implement API integration support - the intention is to replace the current call to `#save` in the controller with separate `#valid?` and `#persist` calls, with an `Step#call_api` method or equivalent between those steps. This does mean the API call is held within the Step itself but does not tie it into the 'save` lifecycle as using callbacks would. It would also be a no-op for most steps apart from the identification ones at the start of the wizard.

**Users manually skipping steps**

Currently there is nothing to enforce that users follow the correct flow - whilst all steps are required this is only necessary at the end of the journey but will be implemented in the abstract `Wizard` class when conditionals are implemented.

### DIFF
--- a/app/models/wizard/base.rb
+++ b/app/models/wizard/base.rb
@@ -1,0 +1,40 @@
+module Wizard
+  class Base
+    class_attribute :steps
+
+    class << self
+      def indexed_steps
+        @indexed_steps ||= steps.index_by(&:key)
+      end
+
+      def step(key)
+        indexed_steps[key] || raise(UnknownStep)
+      end
+
+      def step_index(key)
+        steps.index step(key)
+      end
+    end
+
+    delegate :steps, :step, :step_index, :indexed_steps, to: :class
+
+    def initialize(store)
+      @store = store
+    end
+
+    def find(key)
+      step(key).new @store
+    end
+
+    def previous_step(key)
+      index = step_index(key)
+      index.positive? ? steps[index - 1]&.key : nil
+    end
+
+    def next_step(key)
+      steps[step_index(key) + 1]&.key
+    end
+
+    class UnknownStep < RuntimeError; end
+  end
+end
--- a/app/models/wizard/base.rb
+++ b/app/models/wizard/base.rb
@@ -1,4 +1,6 @@
 module Wizard
+  class UnknownStep < RuntimeError; end
+
   class Base
     class_attribute :steps
 
@@ -47,7 +49,5 @@ module Wizard
     def next_step(key = current_step)
       steps[step_index(key) + 1]&.key
     end
-
-    class UnknownStep < RuntimeError; end
   end
 end
--- a/app/models/wizard/base.rb
+++ b/app/models/wizard/base.rb
@@ -14,24 +14,37 @@ module Wizard
       def step_index(key)
         steps.index step(key)
       end
+
+      def step_keys
+        indexed_steps.keys
+      end
     end
 
-    delegate :steps, :step, :step_index, :indexed_steps, to: :class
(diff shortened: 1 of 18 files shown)


spec/controllers/event_steps_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  describe "#update" do
+    let(:key) { Events::Steps::PersonalDetails.model_name.param_key }
+    before { patch step_path, params: { key => contact_params } }
+    subject { response }
+
+    context "with valid data" do
+      let(:contact_params) { attributes_for(:events_personal_details) }
+      it { is_expected.to redirect_to root_path }
```
ethax-ross, line 20: Does this need another test case to cover if there is a next step? this just seems to test the last step, unless I'm mis-reading

spec/controllers/event_steps_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    context "with invalid data" do
+      let(:contact_params) { { "first_name" => "test" } }
+      it { is_expected.to have_http_status :success }
```
ethax-ross, line 25: Worth testing that the current path doesn't change here?

app/models/wizard/base.rb
```diff
@@ -0,0 +1,40 @@
...
+      @store = store
+    end
+
+    def find(key)
+      step(key).new @store
+    end
+
+    def previous_step(key)
+      index = step_index(key)
+      index.positive? ? steps[index - 1]&.key : nil
```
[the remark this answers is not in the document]
jebw: Good point - I'd added the `&.` later on from the ternary

spec/controllers/event_steps_controller_spec.rb
jebw: Yeh, this particular wizard implementation only has a single step at present - it'll gain more in the PR for the apply for an event wizard

app/models/wizard/base.rb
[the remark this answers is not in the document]
jebw: Just revisited this - It needs to stay in place

`steps[-1]` will return the last item in the array rather than `nil`

PR #82, opened 2020-06-22
GITPB-245 Apply for an event wizard

jebw:
### JIRA ticket number

GITPB-245

### Context

The Apply an Event wizard should capture and validate the steps for the Apply for an Event process

1. Integrated GovUK Form Builder
2. Added email address validation to Personal Details step
3. Added Contact Details step
4. Added Further Details step


app/helpers/events_helper.rb
```diff
@@ -0,0 +1,14 @@
+module EventsHelper
+  def format_event_date(event)
+    return if event.startDate.blank?
```
ethax-ross, line 3: The camel case here makes me cringe - do you want me to update the API to spit out snake case? I think it's just a serialiser setting

spec/controllers/event_steps_controller_spec.rb
```diff
@@ -1,8 +1,18 @@
 require "rails_helper"
```
ethax-ross, line 1: I don't feel strongly about it, but I would expect request specs to live in `/spec/requests/`

jebw: Yeh, good call - I'll move all of them.

app/helpers/events_helper.rb
jebw: Agreed, it was done to maintain parity with the API. I'm in two minds whether the appropriate option is to amend the case here or in the API. It is orthoganol to these changes though so I'd suggest doing that in a separate PR?

ethax-ross: I would vote for changing the API - it's more than likely going to be consumed by Rails apps predominantly so I think it makes sense to change it there; I'll take a look into it

spec/controllers/event_steps_controller_spec.rb
jebw: Ok moved.

PR #92, opened 2020-06-23
Feature/talk to us telephone link

DrLazer:
talk to us telephone link

### DIFF
--- a/app/views/sections/_talk-to-us.html.erb
+++ b/app/views/sections/_talk-to-us.html.erb
@@ -35,7 +35,7 @@
         <div class="talk-to-us__inner__freephone">
             <p>
                 If you’d prefer, you can call us about teaching or teacher training on Freephone 0800 389 2500, Monday-Friday between 8.30am and 5pm.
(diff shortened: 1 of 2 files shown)


app/views/sections/_talk-to-us.html.erb
```diff
@@ -35,7 +35,7 @@
         <div class="talk-to-us__inner__freephone">
             <p>
                 If you’d prefer, you can call us about teaching or teacher training on Freephone 0800 389 2500, Monday-Friday between 8.30am and 5pm.
-                <span class="telephone-number">0800 389 2500</span>.
+                <a href="tel:0800 389 2500" class="telephone-number">0800 389 2500</span>.
```
ethax-ross, line 38: You should drop the spaces in the href as it's a URL:

>However, even though ITU-T E.123 [E.123] recommends the use of space
   characters as visual separators in printed telephone numbers, "tel"
   URIs MUST NOT use spaces in visual separators to avoid excessive
   escaping.

https://tools.ietf.org/html/rfc3966

PR #151, opened 2020-07-10
Feature/cookie acceptance

DrLazer:
cookie acceptance

### DIFF
--- a/app/webpacker/controllers/cookie-acceptance_controller.js
+++ b/app/webpacker/controllers/cookie-acceptance_controller.js
@@ -1,0 +1,24 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+
+    static targets = ["overlay"]
+
+    connect() {
+        this.checkforCookie();
+    }
+
+    checkforCookie() {
+        this.showDialog();
+    }
+
+    accept(event) {
+        event.preventDefault();
+        this.overlayTarget.style.display = "none";
+    }
+
+    showDialog() {
+        this.overlayTarget.style.display = "flex";
+    }
+
+}
--- a/app/webpacker/controllers/cookie-acceptance_controller.js
+++ b/app/webpacker/controllers/cookie-acceptance_controller.js
@@ -10,11 +10,8 @@ export default class extends Controller {
 
     checkforCookie() {
         var cookie = document.cookie;
-        var cookieFrags = cookie.split(';');
-        for(var i=0;i<cookieFrags.length;i+=1) {
-            if(cookieFrags[i].indexOf('GiTBetaCookie=Accepted') > -1) {
-                return;
-            }
+        if(cookie.indexOf('GiTBetaCookie=Accepted') > -1) {
+            return;
         }
         this.showDialog();
     }
--- a/app/webpacker/controllers/cookie-acceptance_controller.js
+++ b/app/webpacker/controllers/cookie-acceptance_controller.js
@@ -9,12 +9,20 @@ export default class extends Controller {
     }
 
     checkforCookie() {
+        var cookie = document.cookie;
+        var cookieFrags = cookie.split(';');
+        for(var i=0;i<cookieFrags.length;i+=1) {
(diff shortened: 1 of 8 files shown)


app/webpacker/controllers/cookie-acceptance_controller.js
```diff
@@ -0,0 +1,32 @@
...
+
+export default class extends Controller {
+
+    static targets = ["overlay"]
+
+    connect() {
+        this.checkforCookie();
+    }
+
+    checkforCookie() {
```
ethax-ross, line 11: I would opt for using a cookie library - doing this manually is verbose and error prone:

https://github.com/js-cookie/js-cookie

app/webpacker/controllers/cookie-acceptance_controller.js
```diff
@@ -0,0 +1,32 @@
...
+                return;
+            }
+        }
+        this.showDialog();
+    }
+
+    accept(event) {
+        event.preventDefault();
+        this.overlayTarget.style.display = "none";
+        document.cookie = "GiTBetaCookie=Accepted; expires=Fri, 31 Dec 2021 12:00:00 UTC";
```
ethax-ross, line 25: Why are we setting it to an absolute date?

app/webpacker/controllers/cookie-acceptance_controller.js
```diff
@@ -0,0 +1,32 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
```
ethax-ross, line 3: Can we add a test for this controller?

app/webpacker/controllers/cookie-acceptance_controller.js
DrLazer: I disagree with this one, we don't plan to have any additional cookie functionality beyond .... is one set at the moment
so bringing in an external dependency for that is a bit overkill

app/webpacker/controllers/cookie-acceptance_controller.js
DrLazer: Discussed with Myles and Lizzie, at the moment it's as requested

app/webpacker/controllers/cookie-acceptance_controller.js
DrLazer: Yes, my next task is to test all the stimulus controllers

app/webpacker/controllers/cookie-acceptance_controller.js
ethax-ross: Ok, I don't feel strongly about bringing in a library, but I don't think the for loop is required, why not just:

```
var cookie = document.cookie
if (cookie.includes('GiTBetaCookie=Accepted'))
{
    return;
}
```

You may want to use `indexOf` instead of polyfilling `includes`...

app/webpacker/controllers/cookie-acceptance_controller.js
ethax-ross: Ok, I guess it's getting removed after private beta anyway

app/webpacker/controllers/cookie-acceptance_controller.js
ethax-ross: It would make sense to add it into this PR unless you're under pressure to get this shipped for a demo?

app/webpacker/controllers/cookie-acceptance_controller.js
DrLazer: theoretically if we used a second cookie value for something on the same domain then we would need to loop them to find the one which contains the value we are looking for.  Plus it's on the client, so the client could edit them.

it already uses indexOf

app/webpacker/controllers/cookie-acceptance_controller.js
DrLazer: yeah, I considered like ... 3 months from todays date for example but ... I guess it doesn't matter so long as this all comes out when it goes public.

PR #169, opened 2020-07-17
Feature/284 send mailing list sign up request

ethax-ross:
### JIRA ticket number

[GITPB-284](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?issueParent=18255%2C20451%2C19124&selectedIssue=GITPB-284)

### Context

When a candidate completes the mailing list form we want to send the data to the API so the CRM can be updated.

- Rename more_info to callback_information

Make the attribute name consistent with the API so that pre-filling the form works automatically.

- Send mailing list add member request

### DIFF
--- a/app/models/mailing_list/steps/degree_stage.rb
+++ b/app/models/mailing_list/steps/degree_stage.rb
@@ -1,7 +1,7 @@
 module MailingList
   module Steps
     class DegreeStage < ::Wizard::Step
-      attribute :degree_status_id
+      attribute :degree_status_id, :integer
       validates :degree_status_id,
                 presence: true,
                 inclusion: { in: :degree_status_option_ids, allow_nil: true }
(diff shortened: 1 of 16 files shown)


app/models/mailing_list/steps/degree_stage.rb
```diff
@@ -15,7 +15,7 @@ def degree_status_options
       end
 
       def degree_status_option_ids
```
goodviber, line 17: This pattern is repeated a few times, maybe share it?

ethax-ross: I'll see if I can DRY it up in a follow up PR 👍

app/models/mailing_list/steps/degree_stage.rb
```diff
@@ -1,7 +1,7 @@
 module MailingList
   module Steps
     class DegreeStage < ::Wizard::Step
-      attribute :degree_status_id
+      attribute :degree_status_id, :integer
       validates :degree_status_id,
```
goodviber, line 5: look a little odd, presence, inclusion, allow_nil ?

PR #188, opened 2020-07-24
Clean up styling around events and sign up forms

ethax-ross:
### JIRA ticket number

[GITPB-510](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?assignee=5e997e0e6b01320c41b6d21c&selectedIssue=GITPB-510)
[GITPB-493](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?assignee=5e997e0e6b01320c41b6d21c&selectedIssue=GITPB-493)

- Tidy up eventbox/short partial usage. Tie class names to the type of event.
- Consolidate event/mailing list registration styles into single stylesheet.
- Update date formatting to match designs.


app/services/get_into_teaching_api/constants.rb
```diff
@@ -1,5 +1,12 @@
 module GetIntoTeachingApi
   module Constants
+    EVENT_TYPES =
+      {
+        "Train to Teach Event" => 222_750_001,
+        "School and University Event" => 0o000000000,
```
goodviber, line 6: is that correct?

ethax-ross: It's just dummy data for now; the event types are still up in the air so we will need to revisit this once they've been worked out...

PR #198, opened 2020-07-28
Remove hard coded links to content

ethax-ross:
### JIRA ticket number

[GITPB-523](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?assignee=5e997e0e6b01320c41b6d21c&selectedIssue=GITPB-523)

### Context

We want to be able to dynamically render the navigation links (and index pages for other content), avoiding the need to hardcode links to the content pages.

- Move markdown_pages into template_handlers

We are going to add another template handler that will provide a sitemap to all ERB pages; shuffling the existing template handler into a namespace that can be used for both going forward.

- Add ERB template handler to inject sitemap

Subclass the ActionView ERB template handler to inject a `@sitemap` instance variable that contains the details and front matter of all the content pages. This will enable us to build lists of content, navigation menus, etc. Takes inspiration from the [Middleman Sitemap](https://middlemanapp.com/advanced/sitemap/).

- Build navigation menus from front matter

The content repo is being updated so that pages include a `navigation` front matter attribute containing the position of the page in the header/footer navigation menus. This commit adds a helper to find pages that should appear in the navigation and updates the static links to be built out dynamically.

### DIFF
--- a/lib/template_handlers/erb.rb
+++ b/lib/template_handlers/erb.rb
@@ -1,0 +1,46 @@
+module TemplateHandlers
+  class ERB < ActionView::Template::Handlers::ERB
+    CONTENT_DIR = Rails.root.join("app/views/content").freeze
+
+    def call(template, source = nil)
+      # inspect objects to Ruby strings which can be eval'd
+      %(@sitemap = #{self.class.sitemap.inspect}; #{super}.html_safe)
+    end
+
+    class << self
+      attr_writer :sitemap
+
+      def sitemap
+        @sitemap ||= build_sitemap
+      end
+
+      def build_sitemap
+        resources = Dir.glob(content_pattern).map do |file|
+          {
+            path: path(file),
+            front_matter: front_matter(file),
+          }
+        end
+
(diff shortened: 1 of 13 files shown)


lib/template_handlers/erb.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      def build_sitemap
+        resources = Dir.glob(content_pattern).map do |file|
+          {
+            path: path(file),
+            front_matter: front_matter(file),
+          }
+        end
+
+        { resources: resources }
```
jebw, line 25: Resources feels like quite an ambiguous name? Unless its part of a bigger picture this may make more sense as something like specific like `markdown_pages`, or `markdown_content`.

ethax-ross: Agreed, I took the naming from the Middleman sitemap but it's not a good fit or representation of what this is for us. I'll merge this down then follow it up with another PR to rename it as I want to get the changes out early doors to avoid any disruption.

PR #208, opened 2020-07-30
generate and render some static events

DrLazer:
generate and render some static events

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -6,18 +6,18 @@ class EventsController < ApplicationController
     categories = ["Train to Teach events", "Online events"]
     @events_by_category = categories.map do |category|
       events = 5.times.collect do
-      {
-        :category => category,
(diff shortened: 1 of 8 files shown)


app/controllers/events_controller.rb
```diff
@@ -3,7 +3,22 @@ def index
...
-    @show_categorised_events = true
+    categories = ["Train to Teach events", "Online events"]
+    @events_by_category = categories.map do |category|
+      events = 5.times.collect do
+      {
+        :category => category,
+        :title => "Train to teach Coventry Online event",
+        :datetime => "11 May 2020 at 17:00 - 18:30",
+        :description => "This online event specifically for anybody looking to train to teach in Coventry will give you the opportunity to ask our experts questions about teacher training.",
+        :type => GetIntoTeachingApi::Constants::EVENT_TYPES["Train to Teach Event"],
```
ethax-ross, line 14: We could pass through `category` here as the type with a bit of tweaking I think?

PR #216, opened 2020-08-04
GITPB-524 Add third party stimulus controllers

jebw:
### JIRA ticket number

GITPB-524

### Context

We should be able to report events from the application to the analytics platforms used by the marketing team.

1. Added a 'base' controller to share the majority of the logic for the analytics platforms.
   1. this will init the service if it hasn't already been initialised
   2. send the requested event defined by the data attributes
   3. If cookies have not yet been accepted, than the event will instead be fired (and the service initialised) when cookies are accepted.
2. Added separate controllers for Facebook, Snapchat, Pinterest, Google Tag Manager
3. The body tag contains the ids in the data attributes so added a helper method to share this across the layout files
4. Fire an event from the Cookies Acceptance controller to notify this controller of the acceptance of cookies
5. Added tests for the controllers

### Guidance to review

Ids can be set in `.env.local` - I'd recommend doing testing in a private browser window

```
PINTEREST_ID=2613344414675
SNAPCHAT_ID=c7147501-f387-420a-b498-7134ee247bbd
FACEBOOK_ID=241400573208008
GOOGLE_TAG_MANAGER_ID=UA-159214613-2
```

Currently only snapchat and facebook are set to fire events but you'll be able to check the scripts are loaded (or not) by inspecting the <HEAD> tag of a page and looking for the appropriate `<script>` tags for a particular service


app/helpers/application_helper.rb
```diff
@@ -1,4 +1,27 @@
 module ApplicationHelper
+  def analytics_body_tag(attributes = {}, &block)
```
ethax-ross, line 2: Worth a test here? You could probably dynamically load the controller names in a spec and assert they appear in the `controller` attribute so it fails if someone adds a new analytics controller but forgets to include it here/ends up adding it manually instead of using the helper

spec/javascript/controllers/gtm_controller_spec.js
```diff
@@ -0,0 +1,67 @@
...
+  ` ;
+
+  // window appears to not be getting redefined between runs, so remove manually
+  afterEach(() => { delete window.gtag }) ;
+
+  describe("with cookie already set", () => {
+    beforeEach(() => {
+      Object.defineProperty(window.document, 'cookie', {
+        writable: true,
+        value: 'GiTBetaCookie=Accepted; expires=Fri, 31 Dec 2021 12:00:00 UTC'
```
ethax-ross, line 17: This test will start failing on that date, we could always create a `Date` in the future in place of this hard coded string

spec/javascript/controllers/pinterest_controller_spec.js
```diff
@@ -0,0 +1,67 @@
+import { Application } from 'stimulus' ;
+import PinterestController from 'pinterest_controller' ;
+
+describe('PinterestController', () => {
```
ethax-ross, line 4: These are fairly repetitive, you could maybe include the core tests in another module and require them in each controller spec to DRY it up (passing in the params for that particular controller)

spec/javascript/controllers/gtm_controller_spec.js
jebw: Yeh, good call - took this from one of the existing controller tests without thinking about it.

ethax-ross:
Looks good, but there are a couple of seemingly unrelated spec failures - one could be flakey because `ENV` is being mocked - maybe add a light abstraction around `ENV` and mock that? (though I can't see why it would be a problem from looking at the code, it's different env vars being mocked out):

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/spec/features/password_protection_spec.rb#L5-L7
https://github.com/DFE-Digital/get-into-teaching-app/pull/216/files#diff-31c50b16a4c9f23be71c49d30afcf7edR11-R15

The other error has been going on since the new error pages were added - looks like a body attribute was removed:

https://github.com/DFE-Digital/get-into-teaching-app/pull/209

ethax-ross:
Turns out these were both unrelated, I've raised PRs to fix them:

https://github.com/DFE-Digital/get-into-teaching-app/pull/219
https://github.com/DFE-Digital/get-into-teaching-app/pull/220

PR #236, opened 2020-08-10
Show newlines in privacy policy content

jebw:
### Context

Currently the privacy policy content is shown as a single block and it is assumed to be safe HTML content

1. Escape unexpected HTML content
2. Use Rails' simple_format helper to process new lines into breaks, and double new lines into paragraphs

### DIFF
--- a/spec/helpers/text_formatting_helper_spec.rb
+++ b/spec/helpers/text_formatting_helper_spec.rb
@@ -1,0 +1,17 @@
+require "rails_helper"
+
+describe TextFormattingHelper, type: :helper do
+  describe "#safe_format" do
+    subject { safe_format content }
+
+    context "with new lines" do
+      let(:content) { "hello\r\n\r\nworld" }
+      it { is_expected.to eql "<p>hello</p>\n\n<p>world</p>" }
+    end
+
+    context "with html content" do
+      let(:content) { "<strong>hello</strong> <em>world</em>" }
+      it { is_expected.to eql "<p>hello world</p>" }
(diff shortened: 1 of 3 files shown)


spec/helpers/text_formatting_helper_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  describe "#safe_format" do
+    subject { safe_format content }
+
+    context "with new lines" do
+      let(:content) { "hello\r\n\r\nworld" }
+      it { is_expected.to eql "<p>hello</p>\n\n<p>world</p>" }
+    end
+
+    context "with html content" do
+      let(:content) { "<strong>hello</strong> <em>world</em>" }
```
ethax-ross, line 13: Do we need to configure `ActionView::Base.sanitized_allowed_tags` to allow the tags used in the privacy policy (I actually thought it allowed `strong` and `em` by default)? It looks like `ul` and `li` will be needed, but if we're including some of the content in different sections as well we may need to allow headings too?

jebw: Hah - yes, good call - the actual helper I'd taken from School Experience - forgot I'd added that bit.

jebw: At present there's not richer formatting anyway - its all just plain text, if its decided to use HTML we could switch out `strip_tags` with `sanitize` but when I talked to Myles earlier it just sounds like it needs the white space to be shown in the output

PR #240, opened 2020-08-10
Feature/stories fixes

sigmaben:
update featured story style. Updated som other styling issues. Also added new assets to content repo.

### DIFF
--- a/app/webpacker/styles/stories.scss
+++ b/app/webpacker/styles/stories.scss
@@ -1,17 +1,19 @@
 .stories-feature {
-    display: flex;
+
+    display: table;
     width: 100%;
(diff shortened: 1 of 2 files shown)


app/webpacker/styles/stories.scss
```diff
@@ -1,19 +1,17 @@
 .stories-feature {
-
-    display: table;
+    display: flex;
```
jebw, line 2: Is it worth setting an ie11 fallback since that doesn't support `display: flex`, eg 

```css
display: -ms-flexbox;
display: flexbox;
```

A lot of DfE staff will view the pages using the default browser on their computers which is (certainly used to be) IE11

PR #252, opened 2020-08-13
swap application layout back to content

DrLazer:
swap application layout back to content


ethax-ross:
We're going to revert the original PR instead of just the layout change.

PR #257, opened 2020-08-14
Layout, copy and design tweaks for event pages

ethax-ross:
### JIRA ticket number

[GITPB-580](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?selectedIssue=GITPB-580)

- Group Get into Teaching events into single section

The Get into Teaching event types (train to teach, application workshop and online events) should all be grouped under the same "Organised by Get into Teaching" heading on the find an event landing page. Other event types should be in a category labelled by their event type, such as "Organised by School or University events".

- Fix layout issues in featured event boxes

If the boxes had less than 3 events in them the width of the eventbox was wrong.
 
- Correct colouring and icons for event boxes

- Add copy to the featured event boxes

- Update copy for each event category page

We should look at moving this copy into the content repo in a future PR.


app/helpers/events_helper.rb
```diff
@@ -47,11 +47,11 @@ def name_of_event_type(type_id)
     type&.value || ""
   end
 
-  def train_to_teach_event_type?(type_id)
-    train_to_teach_types = [
-      GetIntoTeachingApiClient::Constants::EVENT_TYPES["Train to Teach Event"],
-      GetIntoTeachingApiClient::Constants::EVENT_TYPES["Online Event"],
-    ]
-    train_to_teach_types.include?(type_id)
+  def event_type_color(type_id)
```
goodviber, line 50: perhaps a case for a case?

ethax-ross: I like what you did there. Fixed up with a case statement 👍

PR #261, opened 2020-08-17
Display featured events in the correct order

ethax-ross:
### JIRA ticket number

[GITPB-601](https://dfedigital.atlassian.net/browse/GITPB-601)

### Context

Currently the events will be displayed in the order that they are returned from the API, which is relative to their start date. This means the order of the featured boxes will change over time. Instead, they should be explicitly ordered as:

```
Train to Teach Events
Online Events
Application Workshops
Schools or University Events
```

Display featured events in the correct order

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -37,7 +37,7 @@ private
   end
 
   def categorise_events
-    @events_by_category = @events.each_with_object({}) do |event, hash|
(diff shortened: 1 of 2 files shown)


ethax-ross:
Yeah they're in `GetIntoTeachingApiClient::Constants::EVENT_TYPES` - if someone changes them without realising the implications the request spec will fail as I've hard-coded the ordering there.

PR #262, opened 2020-08-17
Feature/page titles

DrLazer:
**Trello ticket**
https://trello.com/c/D3UY8jwT/261-static-page-title-throughout-the-site

page titles were missing throughout the app everywhere except content driven pages.
modified the steps wizard to include a title and added a title where required in controller methods

### DIFF
--- a/app/controllers/event_steps_controller.rb
+++ b/app/controllers/event_steps_controller.rb
@@ -4,7 +4,7 @@ class EventStepsController < ApplicationController
   include WizardSteps
   self.wizard_class = Events::Wizard
 
-  before_action :set_page_title
+  before_action :set_page_title, except: %i[index completed resend_verification]
 
 private
 
@@ -27,7 +27,6 @@ private
   end
 
   def set_page_title
-    @page_title = "Sign up for #{@event.name}"
-    @page_title += ", #{@current_step.title.downcase} step" if @current_step
+    @page_title = "Sign up for #{@event.name}, #{@current_step.title.downcase} step"
   end
 end
--- a/app/controllers/event_steps_controller.rb
+++ b/app/controllers/event_steps_controller.rb
@@ -25,8 +25,9 @@ private
   end
 
   def set_page_title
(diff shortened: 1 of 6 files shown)


app/controllers/event_steps_controller.rb
```diff
@@ -1,9 +1,11 @@
 class EventStepsController < ApplicationController
-  before_action :load_event
+  before_action :load_event
```
jebw, line 2: Please enabled `.editor-config` support in your editor - it should then start stripping white space at the end of lines

app/models/wizard/step.rb
```diff
@@ -8,9 +8,14 @@ class << self
       def key
         name.split("::").last.underscore
       end
+
+      def title
```
jebw, line 12: Can a spec be added to ensure this is correct

app/controllers/mailing_list/steps_controller.rb
```diff
@@ -3,6 +3,8 @@ class StepsController < ApplicationController
     include WizardSteps
     self.wizard_class = MailingList::Wizard
 
+    before_action :set_page_title
```
jebw, line 6: I don't think this action will work on pages which don't have the current step variable set so it should be excluding some of the actions - see

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/controllers/concerns/wizard_steps.rb#L6

app/controllers/event_steps_controller.rb
```diff
@@ -1,9 +1,11 @@
 class EventStepsController < ApplicationController
-  before_action :load_event
+  before_action :load_event 
 
   include WizardSteps
   self.wizard_class = Events::Wizard
 
+  before_action :set_page_title
```
jebw, line 7: I think the variables this relies upon wont be available on every action in the controller, I think you'll need to `except` some of the actions - see

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/controllers/concerns/wizard_steps.rb#L6

PR #269, opened 2020-08-19
Updated the export of data to the API only include valid data

jebw:
### JIRA ticket number

GITPB-517

### Context

We should allow wizard steps to decide how to export their data instead of shipping the entirety of the Wizard store. This will allow 1. Separating the data sent to the API from the UI design of the steps
2. Not sending data for steps which have been skipped

1. Allow steps to export their own data - default in base class is to exclude their data if the step is skipped
2. Only export appropriate fields for authenticate step

### DIFF
--- a/app/models/wizard/steps/authenticate.rb
+++ b/app/models/wizard/steps/authenticate.rb
@@ -1,6 +1,8 @@
 module Wizard
   module Steps
     class Authenticate < ::Wizard::Step
+      IDENTITY_ATTRS = %i[email first_name last_name date_of_birth].freeze
+
       attribute :timed_one_time_password
 
       validates :timed_one_time_password, presence: true, length: { is: 6, message: :invalid },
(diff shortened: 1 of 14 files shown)


app/models/wizard/steps/authenticate.rb
```diff
@@ -26,6 +28,12 @@ def timed_one_time_password=(value)
         super(value)
       end
 
+      def candidate_identity_data
+        @store.fetch(IDENTITY_ATTRS).transform_keys do |k|
```
ethax-ross, line 32: Would it make more sense to go via the `@wizard` here to avoid recreating the logic to map to API attribute keys? something like:

```
IDENTITY_ATTRS = %i[email firstName lastName dateOfBbirth].freeze

def candidate_identity_data
  @wizard.export_camelized_hash.fetch(IDENTITY_ATTRS)
end
```

ethax-ross (approved):
One minor suggestion, happy for you to take it or leave it

app/models/wizard/steps/authenticate.rb
jebw: This was my first inclination - and actually to put it in the api support mixin but the resend action doesn't have a wizard step setup?

ethax-ross (approved):

app/models/wizard/steps/authenticate.rb
jebw: Dependancy on wizard causes other issues so leaving as is

PR #273, opened 2020-08-19
Added additional hero images for mobile if alternatives are set.

jebw:
### JIRA ticket number

555

### Context

When on mobile devices we wish to show different hero images with different aspect ratios

1. Added additional hero sections for mobile
2. These show when narrower than 550px, and the desktop versions are hidden
3. If no mobile versions are set then the desktop versions show always

### Guidance to review

Test various screens and check it looks right - frontmatter attribute is `mobileimage` - works the same as the normal `image` attribute


peteryates (approved):
Looks good. I think this is another good candidate for a component.

PR #274, opened 2020-08-20
Namespace cache keys for Redis

jebw:
### Context

We wish to share a single Redis instance between both apps. We should namespace the keys to ensure isolation.

1. Namespace all keys written to/read from cache under the `GIT` namespace


sfawcett123 (approved):
Looks good

PR #275, opened 2020-08-20
Added user research environment to Rails app

jebw:
### Context

We will be adding an additional user research instance. This will require separate configuration from the other environments

1. Added a new Rails environment called `userresearch`
2. Added a credentials store for the new environment, this includes new Sentry keys and Secret Key Base keys


sfawcett123 (approved):
where you put  
`config.x.git_api_endpoint =     "https://get-into-teaching-api-dev.london.cloudapps.digital/api"`
could this be an environment variable?
it is prone to change per environment 
I was thinking that it maybe possible for terraform to set the environment variable, as it would more likely know where the API would be

Thats probably one for the future though

jebw:
Its in a per-environment config file already, thats why I'm adding a separate rails environment for the new UR instances

PR #283, opened 2020-08-24
Update logo alt text to show it goes to homepage

sigmaben:


DrLazer (changes requested):
Looking at the DAC document I think they were after .... tell the user what it does ... kind of description
So "Go to Home page" or something similar

DrLazer (approved):

PR #285, opened 2020-08-24
Pass correct step to re-authenticate

ethax-ross:
This was probably a copy/paste error; the behaviour doesn't actually change as both steps have a key of `authenticate`, but this will protect against that potentially changing in the future and breaking the mailing list resend verification flow.


jebw (approved):
Good spot!

PR #287, opened 2020-08-25
GITPB-629 UTM handling

jebw:
### JIRA ticket number

GITPB-629

### Context

We should be passing UTM codes across between applications

1. Add mechanism to disable page caching
2. Pass the username and password through in the link
3. Record the UTM codes in the session and pass them through in the link to the TTA site

### Guidance to review

1. It currently lacks tests - it has been manually tested and this is an expendiancy issue, I will get add tests for this behaviour tomorrow

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -69,4 +69,10 @@ module ApplicationHelper
 
     link_to text, path, **options
   end
+
+  def tta_service_link(opts = nil, &block)
+    return nil if ENV["TTA_SERVICE_URL"].blank?
+
+    link_to ENV["TTA_SERVICE_URL"], opts, &block
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -73,6 +73,11 @@ module ApplicationHelper
   def tta_service_link(opts = nil, &block)
     return nil if ENV["TTA_SERVICE_URL"].blank?
 
-    link_to ENV["TTA_SERVICE_URL"], opts, &block
+    url = ENV["TTA_SERVICE_URL"]
+    if Rails.application.config.x.utm_codes && session[:utm]
+      url += "?" + session[:utm].to_param
+    end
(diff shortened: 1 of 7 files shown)


spec/helpers/application_helper_spec.rb
```diff
@@ -78,4 +78,31 @@
...
+      it { is_expected.to be_nil }
+    end
+
+    context "with username and password set" do
+      before do
+        allow(ENV).to receive(:[]).with("HTTPAUTH_USERNAME").and_return "user"
+        allow(ENV).to receive(:[]).with("HTTPAUTH_PASSWORD").and_return "pass"
+      end
+
+      it { is_expected.to have_css "a[href=\"https://user:pass@tta-service/\"]" }
```
ethax-ross, line 105: Is this to avoid the user being prompted for the basic auth creds when navigating to the TTA service? Presumably the browser history will then contain these for others to see, unless it's filtered out by modern browsers? is that a problem?

app/helpers/application_helper.rb
```diff
@@ -69,4 +69,22 @@ def back_link(path = :back, text: "Back", **options)
 
     link_to text, path, **options
   end
+
+  def tta_service_link(opts = nil, &block)
```
ethax-ross, line 73: Take it or leave it:

```
def tta_service_link(opts = nil, &block)
  url = ENV["TTA_SERVICE_URL"].to_s

  return nil if url.blank?

  if [ENV["HTTPAUTH_USERNAME"], ENV["HTTPAUTH_PASSWORD"]].all?(&:present?)
    url.sub!("https://", "https://#{ENV['HTTPAUTH_USERNAME']}:#{ENV['HTTPAUTH_PASSWORD']}@")
  end

  if Rails.application.config.x.utm_codes && session[:utm]
    url += "?" + session[:utm].to_param
  end

  link_to url, opts, &block
end
```

ethax-ross (approved):

PR #290, opened 2020-08-26
GITPB-633 Redirect for TTA service

jebw:
### JIRA ticket number

GITPB-633

### Context

We need to link from the GiT site to the TTA service. The link needs to be different in different environments. The static markdown pages in the content repo cannot have condiitonal logic within them. 

1. Add redirecting path across to the TTA service - this provides a consistent internal link, which will redirect across to the appropriate TTA service for the environment the GiT site is running within. This includes any UTM codes in the users session.
2. Replaced the earlier solution to this problem of a rails helper to dynamicly generate the url

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,4 +1,6 @@
 class PagesController < ApplicationController
+  rescue_from ActionView::MissingTemplate, with: :rescue_missing_template
+
   def scribble
     @page_title = "Scribble Test"
     render template: "pages/scribble"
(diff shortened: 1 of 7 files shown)


spec/requests/pages_controller_spec.rb
```diff
@@ -46,4 +46,27 @@
       end
     end
   end
+
+  describe "redirect to TTA site" do
+    context "with /tta-service url" do
+      include_context "stub env vars", "TTA_SERVICE_URL" => "https://tta-service/"
```
ethax-ross, line 52: I think some of the setup could be moved into the parent `describe` block:

```
describe "redirect to TTA site" do
  include_context "stub env vars", "TTA_SERVICE_URL" => "https://tta-service/"
  subject { response }

  context "with /tta-service url" do
    before { get "/tta-service" }
    it { is_expected.to redirect_to "https://tta-service/" }
  end

  context "with /tta url" do
    before { get "/tta" }
    it { is_expected.to redirect_to "https://tta-service/" }
  end

  context "with utm params" do
    before { get "/tta-service?utm_test=abc&test=def" }
    it { is_expected.to redirect_to "https://tta-service/?utm_test=abc" }
  end
end
```

app/controllers/pages_controller.rb
```diff
@@ -1,4 +1,6 @@
 class PagesController < ApplicationController
+  rescue_from ActionView::MissingTemplate, with: :rescue_missing_template
```
ethax-ross, line 2: Not sure if this has a request spec around it already, but may be worth adding if you're refactoring it?

ethax-ross (approved):

app/controllers/pages_controller.rb
jebw: Think its already covered but will check

jebw: Wasn't covered so added a spec

spec/requests/pages_controller_spec.rb
jebw: good call

ethax-ross (approved):

PR #295, opened 2020-08-27
Feature/webchat

DrLazer:
hide the webchat controls by default
use JS to enable them
launch chat in a new window

### DIFF
--- a/app/webpacker/controllers/talk-to-us_controller.js
+++ b/app/webpacker/controllers/talk-to-us_controller.js
@@ -2,7 +2,7 @@ import { Controller } from "stimulus"
 
 export default class extends Controller {
 
-    static targets = ["chat", "launch", "tta"];
+    static targets = ["chat", "tta"];
 
     connect() {
         this.showChatControls();
--- a/app/webpacker/controllers/talk-to-us_controller.js
+++ b/app/webpacker/controllers/talk-to-us_controller.js
@@ -1,0 +1,24 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+
+    static targets = ["chat", "launch", "tta"];
+
+    connect() {
+        this.showChatControls();
+    }
+
+    showChatControls() {
+        this.chatTarget.style.display = 'inline-block';
+        this.ttaTarget.style.width = "48%";
+    }
+
+    startChat(e) {
+        e.preventDefault();
+        var windowFeatures = "menubar=no,location=yes,resizable=yes,scrollbars=no,status=yes,width=340,height=480";
+        window.open("https://gov.klick2contact.com/v03/launcherV3.php?p=DfE&d=971&ch=CH&psk=chat_a1&iid=STC&srbp=0&fcl=0&r=Static&s=https://gov.klick2contact.com/v03&u=&wo=&uh=&pid=82&iif=0",
+        "Get into teaching: Chat online",
+        windowFeatures);
+    }
+
+}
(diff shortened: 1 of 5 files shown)


app/webpacker/controllers/talk-to-us_controller.js
```diff
@@ -0,0 +1,24 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+
+    static targets = ["chat", "launch", "tta"];
```
ethax-ross, line 5: You don't need the `launch` target, it's just the action being used

app/webpacker/controllers/talk-to-us_controller.js
```diff
@@ -0,0 +1,24 @@
...
+
+    connect() {
+        this.showChatControls();
+    }
+
+    showChatControls() {
+        this.chatTarget.style.display = 'inline-block';
```
ethax-ross, line 12: Would it be nicer to have this in CSS and just toggle a class name?

ethax-ross (commented):
Looks good, but can we add a test for the Stimulus controller

app/webpacker/controllers/talk-to-us_controller.js
DrLazer: nice one

app/webpacker/controllers/talk-to-us_controller.js
DrLazer: debatable one.  if we end up adding more attribute with the action then yes.
but I dont think thats the plan.
its not a good idea to have like ...

.hidden { display: none }
.block { display: block }

etc, as you are then micro-abstracting the css logic

spec/javascript/controllers/talk-to-us_controller_spec.js
```diff
@@ -0,0 +1,55 @@
...
+describe('TalkToUsController', () => {
+
+    document.body.innerHTML = 
+    `
+    <section id="talk-to-us" class="talk-to-us" data-controller="talk-to-us">
+
+        <div class="talk-to-us__inner__table__column" data-target="talk-to-us.chat" id="1">
+            <p>
+                <b>Chat online</b> <br/>
+                If you have questions about getting into teaching, we can help you get the answers you need with our one-to-one live online chat service,
```
ethax-ross, line 13: Might be worth removing the copy/unnecessary tags from the HTML fixture to make it easy to read

spec/javascript/controllers/talk-to-us_controller_spec.js
```diff
@@ -0,0 +1,55 @@
...
+
+    </section>
+    `;
+
+    const application = Application.start() ;
+    application.register('talk-to-us', TalkToUsController) ;
+
+    describe("when first loaded", () => {
+
+        it("should make the chat section visible", () => {
```
ethax-ross, line 43: I think you could test that `window.open` is called with something like:

```
 global.open = jest.fn();
button.click();
expect(global.open).toBeCalledWith(...);
```

I've found these sorts of tests are a bit hit and miss with Jest though; trying to test `window.x` doesn't always work.

ethax-ross (approved):
Looks good - couple of comments you can take or leave.

spec/javascript/controllers/talk-to-us_controller_spec.js
DrLazer: ha, I had a go at that but didnt get it working.
I was trying spy, and window.spy

tbh its not testing much other than window.open does open a window, which is probably out of scope for the test

PR #301, opened 2020-09-02
Feature/footer social

sigmaben:
Add brand icons font variant. 
Add social links.
Update Footer styling to accommodate links and bring in line with design. 
Update talk section stylie to fix bugs from chat integration

### DIFF
--- a/app/views/sections/_footer.html.erb
+++ b/app/views/sections/_footer.html.erb
@@ -2,23 +2,23 @@
     <%= render "sections/talk-to-us" %>
     <div class="footer-top">
         <div class="footer-top__social">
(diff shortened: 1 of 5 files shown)


jebw (commented):
How necessary is it to implement this with flexbox? I'm aware IE11 doesn't support that very well and that's what most of DfE uses as a default browser?

sigmaben:
@jebw Id prefer to use flexbox over table as a display property as it serves most users better. I'll add fallback support for IE11 and add to the PR.

jebw (dismissed):
Looks good but can you add whitespace before the new attributes

jebw (approved):
LGTM

PR #304, opened 2020-09-03
GITPB-612 Extra event step

jebw:
### JIRA ticket number

GITPB-612

### Context

We now give candidates the opportunity to subscribe to the newsletter as part of the signup process.

1. Dropped the future events field
2. Moved the postcode field to a new screen
3. Added fields for Degree Status, Journey Stage and Preferred Teaching subject
4. Updated the API integration to submit the new fields
5. Updated the completion screen to confirm if the user has subscribed to the newsletter
6. Added a javascript controller to update the label on the submit button dependant on whether they are subscribing to the newsletter

### DIFF
--- a/spec/models/events/steps/personalised_updates_spec.rb
+++ b/spec/models/events/steps/personalised_updates_spec.rb
@@ -27,4 +27,23 @@ describe Events::Steps::PersonalisedUpdates do
       expect(subject.address_postcode).to be_nil
     end
   end
+
+  context "skipped?" do
+    let(:mailing_list) { nil }
+    let(:backingstore) { { "subscribe_to_mailing_list" => mailing_list } }
+
+    context "with mailing_list question not answered" do
+      it { is_expected.to be_skipped }
+    end
+
+    context "with mailing_list question answered as yes" do
+      let(:mailing_list) { true }
+      it { is_expected.not_to be_skipped }
+    end
+
+    context "with mailing_list question answered as no" do
+      let(:mailing_list) { false }
+      it { is_expected.to be_skipped }
+    end
+  end
 end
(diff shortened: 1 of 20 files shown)


spec/models/events/steps/personalised_updates_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    it { is_expected.not_to allow_value("12345").for :consideration_journey_stage_id }
+
+    it { is_expected.to allow_value(subject.teaching_subject_option_ids.first).for :preferred_teaching_subject_id }
+    it { is_expected.to allow_value(subject.teaching_subject_option_ids.last).for :preferred_teaching_subject_id }
+    it { is_expected.not_to allow_value(nil).for :preferred_teaching_subject_id }
+    it { is_expected.not_to allow_value("12345").for :preferred_teaching_subject_id }
+  end
+
+  context "data cleaning" do
+    it "cleans the postcode" do
```
ethax-ross, line 40: One for later, but I wrote a small shared example to deal with this for the TTA app that we could pull in here as well:

https://github.com/DFE-Digital/get-teacher-training-adviser-service/blob/master/spec/support/shared_examples/sanitize_input.rb

Then this becomes:

```
include_context "sanitize fields", %i[address_postcode]
```

spec/models/events/steps/personalised_updates_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  context "attributes" do
+    it { is_expected.to respond_to :address_postcode }
+    it { is_expected.to respond_to :degree_status_id }
+    it { is_expected.to respond_to :consideration_journey_stage_id }
+    it { is_expected.to respond_to :preferred_teaching_subject_id }
+  end
+
+  context "validations" do
+    it { is_expected.to allow_value("TE571NG").for :address_postcode }
```
ethax-ross, line 17: These may be easier to read/more concise if we grouped some of them into single blocks:

```
it { is_expected.to allow_values("TE571NG", "TE57 1NG", " TE57 1NG ", "").for :address_postcode }
... etc ...
```

ethax-ross (approved):
Looks good ✅

PR #306, opened 2020-09-04
Remove sign up/how to attend from closed events

ethax-ross:
### JIRA ticket number

[GITPB-578](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?selectedIssue=GITPB-578)

### Context

If an event has a status of 'Closed' then it should not be possible to sign up or enquire about attendance. Also, if someone attempts to access the sign up page for a closed event we should redirect them and inform them that they cannot sign up.

- Remove sign up/how to attend from closed events

This commit hides the 'Sign up for this event' buttons from the event detail page if the event accepts online registration and also removes the 'How to attend' section if the event is closed.

- Prevent sign up to closed event

If an event is closed we don't want to let users proceed through the sign up journey. Instead, they will be redirected to the event page which has been updated to include a message when the event is closed for new attendees.

### Guidance to review

Screenshot 2020-09-04 at 13 15 44

### DIFF
--- a/app/controllers/event_steps_controller.rb
+++ b/app/controllers/event_steps_controller.rb
@@ -1,10 +1,17 @@
 class EventStepsController < ApplicationController
   before_action :load_event
+  before_action :redirect_closed_events, only: %i[show update] # rubocop:disable Rails/LexicallyScopedActionFilter
 
   include WizardSteps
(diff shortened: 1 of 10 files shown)


jebw (approved):
LGTM

PR #308, opened 2020-09-07
Add brakeman

jebw:
### Context

Add Brakeman to our CI pipeline to automatically scan for vulnerabilities

1. Added Brakeman gem
2. Run brakeman as part of CI run
3. Filter page template requests
4. Replaced a dynamic redirect

### DIFF
--- a/app/controllers/stories_controller.rb
+++ b/app/controllers/stories_controller.rb
@@ -1,4 +1,6 @@
 class StoriesController < ApplicationController
+  include StaticPages
+
   def show
     new_stories = ["becoming-a-mum-sparked-my-interest-in-teaching"]
     is_new_story = new_stories.any? { |story| params[:story].include?(story) }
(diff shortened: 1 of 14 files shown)


app/controllers/stories_controller.rb
```diff
@@ -18,4 +18,14 @@ def show
...
   end
+
+private
+
+  def stories_template
+    "content/life-as-a-teacher/my-story-into-teaching/#{filtered_story_page}"
+  end
+
+  def filtered_story_page
+    params[:story].to_s.gsub(%r{[^a-z_\-/]}, "")
```
ethax-ross, line 29: Is it worth extracting this into a constant to describe what it's matching? Also, does the forward slash not need escaping and will this match a URL containing numbers? I'm thinking if they put something in like:

`stories/my-top-10-teaching-moments`

spec/requests/event_steps_controller_spec.rb
```diff
@@ -93,6 +93,10 @@
       get resend_verification_event_steps_path(readable_event_id, redirect_path: "redirect/path")
       response
     end
-    it { is_expected.to redirect_to("redirect/path") }
+
+    it do
+      is_expected.to redirect_to \
```
ethax-ross, line 98: Much nicer 💯

app/controllers/stories_controller.rb
jebw: Good call

ethax-ross (approved):

PR #311, opened 2020-09-08
Add feedback banner

ethax-ross:
### Context

We want to show a banner prompting for feedback on all pages. If a user opts to hide the banner it should only re-appear in another session (when they close the browser - so no expiry on the cookie). If the user clicks on 'tell us what you think' they should be taken to the feedback form in a new tab/window:

https://docs.google.com/forms/d/1Fp6At8Jrfg9umL42jT9iM4faS0rtWEA29dLrFbuKPXk/viewform?ts=5f464d3a&edit_requested=true

We also want to ask 'was this page helpful' on all the content pages (except home page/index pages).

The functionality for what happens if someone clicks on yes/no in response to the question 'is this page helpful?' will be implemented in a following PR.

- Add phase banner/page question partials
- Add Stimulus controllers to hide the banners and persist that state for the browser session
- Update link to feedback survey
- Enable hiding the 'is this page helpful?' question via front matter

### Guidance to review

Screenshot 2020-09-08 at 11 45 23

Screenshot 2020-09-08 at 11 45 26


DrLazer (approved):
cool!

PR #313, opened 2020-09-09
Send "is this page helpful?" responses to Prometheus

ethax-ross:
### JIRA ticket number

[GITPB-574](https://dfedigital.atlassian.net/browse/GITPB-547)

### Context

We currently as the user on various content pages 'is this page helpful?' but the links for yes/no are not connected to any storage mechanism. 

- We want to send these responses through to Prometheus so we can display them in a Grafana dashboard. 
- When a user has sent a response we want to display the text "Thank you for your feedback" and hide the yes/no buttons. 
- If there is a problem when sending the response we should show an error message to the user, prompting them to try again.
- If the user has Javascript disabled the feedback box should not display at all. 

- Add Prometheus::Client

Add the `Prometheus::Client` for aggregating and exporting metrics on the `/metrics` endpoint.

Configured to use `DirectFileStore` as this supports Puma forking additional processes (by default metrics will be summed together; if we use a gauge metric this may not be desirable).

- Add FeedbackController to increment :page_helpful metric

Introduces the `FeedbackController` which will increment a `:page_helpful` counter metric, tagging it with labels for the URL and the 'answer' (yes or no).

This will be called when a user hits the yes/no links in response to the question 'Is this page useful?'

- Expose form authenticity token to Javascript

We will need to make a POST ajax request from a Stimulus controller in order to submit the feedback 'page helpful' answer. This exposes the form authenticity token to JS so that we can use it when making the request.

- Add fetch polyfill

So that we can support `fetch` in IE.

- Add PageHelpfulController for sending responses

Adds a Stimulus controller to send feedback responses to the application when the user clicks yes/no. We display a thank you message to the user when their submission has been sent. If there was an error we inform the user that their submission was not successful.

### Guidance to review

Best reviewed by-commit.

### DIFF
--- a/app/views/sections/_head.html.erb
+++ b/app/views/sections/_head.html.erb
@@ -8,5 +8,4 @@
   <%= tag :link, rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' %>
   <%= stylesheet_pack_tag 'application', 'data-turbolinks-track': 'reload', media: 'all' %>
   <%= javascript_pack_tag 'application', 'data-turbolinks-track': 'reload', defer: true %>
-  <%= javascript_tag "window._token = '#{form_authenticity_token}'" %>
 </head>
--- a/app/views/sections/_head.html.erb
+++ b/app/views/sections/_head.html.erb
@@ -8,4 +8,5 @@
   <%= tag :link, rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' %>
   <%= stylesheet_pack_tag 'application', 'data-turbolinks-track': 'reload', media: 'all' %>
   <%= javascript_pack_tag 'application', 'data-turbolinks-track': 'reload', defer: true %>
+  <%= javascript_tag "window._token = '#{form_authenticity_token}'" %>
 </head>
--- a/config/application.rb
+++ b/config/application.rb
@@ -15,9 +15,22 @@ require "action_view/railtie"
 # require "sprockets/railtie"
 require "rails/test_unit/railtie"
 
+require "prometheus/client/data_stores/direct_file_store"
+
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
+
+PROMETHEUS_DIR = "/tmp/prometheus".freeze
+
+# Needs to clear before initializing.
+Dir["#{PROMETHEUS_DIR}/*.bin"].each do |file_path|
+  File.unlink(file_path)
+end
+
+# The DirectFileStore is the only way to aggregate metrics across processes.
+file_store = Prometheus::Client::DataStores::DirectFileStore.new(dir: PROMETHEUS_DIR)
+Prometheus::Client.config.data_store = file_store
 
 module GovukRailsBoilerplate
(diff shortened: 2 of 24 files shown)


config/application.rb
```diff
@@ -15,10 +15,23 @@
...
 
+PROMETHEUS_DIR = "/tmp/prometheus".freeze
+
+# Needs to clear before initializing.
+Dir["#{PROMETHEUS_DIR}/*.bin"].each do |file_path|
+  File.unlink(file_path)
+end
+
+# The DirectFileStore is the only way to aggregate metrics across processes.
+file_store = Prometheus::Client::DataStores::DirectFileStore.new(dir: PROMETHEUS_DIR)
```
jebw, line 32: I'm not understanding this - but a FileStore looks like a problem? The comment references aggregating across processes, but we'll need to aggregate across nodes. The file store will also be lost on deployment?

jebw (changes requested):
We need to handle the form submission part differently I think

app/views/sections/_head.html.erb
```diff
@@ -8,4 +8,5 @@
   <%= tag :link, rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' %>
   <%= stylesheet_pack_tag 'application', 'data-turbolinks-track': 'reload', media: 'all' %>
   <%= javascript_pack_tag 'application', 'data-turbolinks-track': 'reload', defer: true %>
+  <%= javascript_tag "window._token = '#{form_authenticity_token}'" %>
```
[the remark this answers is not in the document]
ethax-ross: Well spotted, I've added a commit to move the CRSF token to a separate endpoint and it's now called from the Stimulus controller prior to submitting the feedback.

config/application.rb
ethax-ross: Aggregating across nodes should be done by the Prometheus server using jobs:

https://prometheus.io/docs/concepts/jobs_instances/

Restarting the application should only lose a small sample of metrics (from the time of the last scrape to the time the app restarts); if we scrape frequently enough the impact of this should be negligible. Granted, we're sending a potentially low-frequency metric for the feedback responses, but again the amount we miss from this edge case will be small (I'll raise it with PM to make sure they're happy with this).

The `DirectFileStore` is the recommended approach for applications with pre-fork servers/multi-processors, but there are caveats:

- When scraped by Prometheus the store reads all the files into memory (one file per metric per process), which can bump up the memory usage a fair amount. I don't anticipate this being an issue for us yet/soon given the amount of metrics we're collecting is small.

- The metric files need to be cleared out each time the application starts to avoid a new instance exporting metrics from a previous one; I've done this programatically on start but since we're dockerized it probably won't matter anyway.

- A large number of files can be created; there's no real solution for this one in the client library. If we're on the way to running out of disk space we should be alerted by infrastructure monitoring, but I imagine worst-case scenario is the container runs out of memory, blows up and a new one spins up to take it's place. It would only result in a problem if it happens on all instances at the same time - if we're worried about that then we could schedule periodic restarts of the instances to keep the usage low.

ethax-ross:
We're storing this in GA instead, closing for now but we'll want to pick up the Prometheus integration separately at some point.

PR #315, opened 2020-09-09
Prevent API check when already authenticated

ethax-ross:
### JIRA ticket number

[GITPB-692](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?selectedIssue=GITPB-692)

### Context

When the wizard is about to be completed we check for any invalid steps; this was resulting in another API call to verify the verification code in the `Authenticate` step. If the user takes too long completing the form their code may not be valid by the time they reach they come to submit it, resulting in them being redirected back to the `Authenticate` step.

This PR sets an `authenticated` flag in the store; if this is present the `Authenticate` step will not try and re-verify their code.

- Prevent API check when already authenticated


jebw (approved):
LGTM

PR #318, opened 2020-09-10
Prompt for feedback when closing window

ethax-ross:
### JIRA ticket number

[GITPB-659](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?selectedIssue=GITPB-659)

### Context

When a user moves their mouse to close the window/tab, we want to prompt them to leave feedback.

- The dialog should only appear once per browser session.
- If actioned, the dialog should not appear for another 12 months.

- Prompt for feedback when closing window

Add `FeedbackPromptController` to intercept when the mouse leaves the top of the viewport and display feedback dialog partial. Set a cookie when the prompt has been seen so we only show once per browser session. Set a cookie when the prompt has been actioned so that we don't show for another 12 months.

Consolidate the cookie acceptance CSS to make it a more generic `dialog.scss`.

### Guidance to review

Screenshot 2020-09-10 at 08 35 08


app/views/sections/_feedback-prompt.html.erb
```diff
@@ -0,0 +1,18 @@
...
+      <h3>We'd like your feedback</h3>
+      <p>
+        To help us improve the Get Into Teaching website, we'd like to ask you a few short questions about your experience today.
+      </p>
+      <a href="https://docs.google.com/forms/d/e/1FAIpQLSfIBfaYUk07Fb6dxUKAokiPoGGL73ZYVGHM1mSNyA-K-6QRbA/viewform" data-action="feedback-prompt#disable" class="call-to-action-button">
+        Yes, I'll answer the questions <i class="fas fa-chevron-right"></i>
+      </a>
+      <br />
+      <a class="secondary-link" data-action="feedback-prompt#close #feedback-prompt#disable">
+        No thank you <i class="fas fa-chevron-right"></i>
```
DrLazer, line 14: This needs a screen reader annotation to work out of context
"I don't want to answer the questions"

DrLazer (changes requested):
just the annotation

DrLazer (approved):

PR #320, opened 2020-09-10
Add Google map to event page

ethax-ross:
### JIRA ticket number

[GITPB-246](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?assignee=5e997e0e6b01320c41b6d21c&selectedIssue=GITPB-246)

### Context

We want to display a Google map of the venue location when viewing an event.

- Add Google map to event page

Lift/shift of the Google map code from Schools Experience, tweaking it to work with addresses instead of lat/lng coordinates as the API for GiT only geocodes the postcode, which isn't accurate enough to plot on a map.

### Guidance to review

The Stimulus `MapController` could do with some tests, but it's so dependent on the Google Maps APIs that it's probably not going to be worth it at the moment (there's no tests in the Schools Experience repo where this is lifted from, so that may be the reason).

JS enabled map:

Screenshot 2020-09-14 at 20 33 42

JS disabled map:

Screenshot 2020-09-14 at 20 36 29


config/environments/rolling.rb
```diff
@@ -5,4 +5,6 @@
   # Override any production defaults here
   config.x.git_api_endpoint = \
     "https://get-into-teaching-api-dev.london.cloudapps.digital/api"
+  config.x.google_maps_key = ENV["GOOGLE_MAPS_KEY"].presence || \
```
jebw, line 8: This is unneeded since its getting set via the inheritance from the `production.rb`

config/environments/userresearch.rb
```diff
@@ -5,7 +5,8 @@
   # Override any production defaults here
   config.x.git_api_endpoint = \
     "https://get-into-teaching-api-dev.london.cloudapps.digital/api"
-
+  config.x.google_maps_key = ENV["GOOGLE_MAPS_KEY"].presence || \
```
jebw, line 8: This is unneeded since its getting set via the inheritance from the `production.rb`

config/environments/preprod.rb
```diff
@@ -5,4 +5,6 @@
   # Override any production defaults here
   config.x.git_api_endpoint = \
     "https://get-into-teaching-api-dev.london.cloudapps.digital/api"
+  config.x.google_maps_key = ENV["GOOGLE_MAPS_KEY"].presence || \
```
jebw, line 8: This is unneeded since its getting set via the inheritance from the `production.rb`

jebw (approved):
LGTM but would suggest removing the duplicate lines in `preprod.rb`/`rolling.rb`/`userresearch.rb`

jebw (approved):

PR #331, opened 2020-09-14
Add support for on-campus mailing list sign ups

ethax-ross:
### JIRA ticket number

[GITPB-611](https://dfedigital.atlassian.net/jira/software/projects/GITPB/boards/61?assignee=5e997e0e6b01320c41b6d21c&selectedIssue=GITPB-611)

### Context

We want to enable on-campus mailing list sign ups where the type of sign up is determined by the `ChannelId` sent to the CRM. The sign up URLs should be:

`/mailinglist/signup?channel=:channel_id`

We also need to display a tracking pixel on the sign up completion page, so that the sign up can be attributed to an agent at the event (via BAM).

- Include query parameters in index redirect

When we land on the index page for a wizard we get redirected to the first step; this ensures that any query parameters present are also passed through.

We will need this to ensure the `channel_id` for on-campus events is available in the mailing list wizard: `/mailinglist/sign_up?channel_id=123

- Set channel_id on first step of mailing list sign up

If a `channel` is passed as part of the query string we want to persist this in the candidate's session so that when they come to submit their completed form it contains the `channel_id` as part of the payload.

This will be used to track on-campus mailing list sign ups via URLs:

`/mailinglist/signup?channel=<channel_id>`

- Update API client dependency

- Add BAM tracking pixel to mailing list completion page

Add a BAM tracking pixel to the mailing list completion page. The pixel is of the image variety (as opposed to JS) so we need to trigger it with a `fetch` from JS, ensuring it only fires when the various cookie preferences are set.

### DIFF
--- a/.env.preprod
+++ b/.env.preprod
@@ -4,5 +4,6 @@ SNAPCHAT_ID=c7147501-f387-420a-b498-7134ee247bbd
 FACEBOOK_ID=241400573208008
 TWITTER_ID=o0wo4
 GOOGLE_TAG_MANAGER_ID=GTM-K7TK4WM
+BAM_ID=1
 TTA_SERVICE_URL="https://beta-adviser-getintoteaching.education.gov.uk/"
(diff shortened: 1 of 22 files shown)


jebw (approved):
SGTM

PR #334, opened 2020-09-14
Moved OWASP Scanner to Application

sfawcett123:
Moved the OWASP CVE scanner from Content to the code base, to increase performance of the Delivery

### DIFF
--- a/.zap/rules.tsv
+++ b/.zap/rules.tsv
@@ -1,0 +1,3 @@
+10038	IGNORE	(Content Security Policy (CSP) Header Not Set)
+10063	IGNORE	(Feature Policy Header Not Set)
+40025	IGNORE	(Proxy Disclosure)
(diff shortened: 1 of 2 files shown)


.zap/rules.tsv
```diff
@@ -0,0 +1,3 @@
+10038	IGNORE	(Content Security Policy (CSP) Header Not Set)
```
jebw, line 1: This line can go - we've got a CSP now

jebw (changes requested):
Looks good but needs the CSP ignore removing

sfawcett123:
The application never runs as a URL on the Gov PaaS and therefore getting a HTTP response is difficult, The alternative is simply to schedule CVE tests after a deployment of CONTENT but not cause the deployment to be dependant on it

PR #337, opened 2020-09-14
GITPB-694 Cookie preferences

jebw:
### JIRA ticket number

GITPB-694

### Context

As a user I expect to be able to manage my cookie preferences.

1. Add a static page with a form to manage cookie preferences
2. Add a CookiePreferences js object to handle persisting users selection, and emitting events with changes when the user accepts new categories of cookies
3. Added CookiePreferencesController to make the form in (1) interactive, uses CookiePreferences object to manage cookies
4. Updated various analytics controllers to only fire if their category of cookies has been enabled.

### DIFF
--- a/app/controllers/cookie_preferences_controller.rb
+++ b/app/controllers/cookie_preferences_controller.rb
@@ -1,0 +1,3 @@
+class CookiePreferencesController < ApplicationController
+  def show; end
+end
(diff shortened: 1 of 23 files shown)


ethax-ross (commented):
Looks good - a couple of take it or leave it comments, but there's also a few other areas in the app that need updating to reference the new `CookiePreference` object:

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/webpacker/controllers/feedback_prompt_controller.js
https://github.com/DFE-Digital/get-into-teaching-app/blob/078aa5c5a6aff67bc6bf5e67153e4bd025a12b40/app/webpacker/controllers/banner_controller.js

ethax-ross (approved):
LGTM

PR #341, opened 2020-09-15
GITPB-583 Only load in page video player if the user has accepted cookies

jebw:
### JIRA ticket number

GITPB-583 

### Context

We should only be loading youtube into the page if the user has accepted the non-functional category of cookies

1. Don't load youtube by default in the iframe src
2. Change the activate code to only fire if cookies have been accepted
3. Change the playback code to only fire if the activate code has fired
4. Add an event handler for when the cookies have changed to activate the in page player if appropriate


app/webpacker/controllers/video_controller.js
```diff
@@ -1,15 +1,41 @@
...
+      }
+    }
+
+    get playerAllowedByCookies() {
+      const cookiePrefs = new CookiePreferences() ;
+      return cookiePrefs.allowed('non-functional')
     }
     
+    cookiesAcceptedChecker(event) {
+    console.log("I FIRED ") ;
```
ethax-ross, line 23: Stray `console.log`

jebw: Darn - good catch

ethax-ross (approved):

PR #351, opened 2020-09-16
Added twitter tracking pixel support

jebw:
### JIRA ticket number

GITPB-667

### Context

We require the ability to integrate with twitters tracking pixels

1. Added support for twitters tracking pixels


DrLazer (approved):
yep, looks good also

PR #360, opened 2020-09-22
Remove leading / from page templates

ethax-ross:
The leading slash is not part of the allowed template format that we enforce prior to rendering a page.


jebw (approved):
LGTM

PR #362, opened 2020-09-22
Add another BAM tracking pixel

ethax-ross:
They will be using this pixel to test the tracking in the pre-prod environment.


jebw (approved):
LGTM

PR #365, opened 2020-09-23
Fix BAM tracking pixel

ethax-ross:
BAM have reported issues with their tracking links, I think because we're making the requests using `fetch` and `no-cors` so their cookie is not being sent. This PR switches out `fetch` to instead append `img` tags to the  DOM, so the browser makes the requests.


jebw (approved):
LGTM

PR #367, opened 2020-09-23
Check Redis and API connection in healthcheck.json

jebw:
### Context

We should be able to see if we're getting connectivity issues to either API or Redis

1. Extend the /healthcheck.json end point to show if connections to API and Redis are working

### DIFF
--- a/app/models/healthcheck.rb
+++ b/app/models/healthcheck.rb
@@ -9,10 +9,27 @@ class Healthcheck
     read_file "/etc/get-into-teaching-content-sha"
   end
 
+  def test_api
+    GetIntoTeachingApiClient::TypesApi.new.get_teaching_subjects
+    true
+  rescue Faraday::Error
+    false
+  end
+
+  def test_redis
+    return nil unless ENV["REDIS_URL"]
+
+    Redis.current.ping == "PONG"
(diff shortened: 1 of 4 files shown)


app/models/healthcheck.rb
```diff
@@ -9,10 +9,27 @@ def content_sha
     read_file "/etc/get-into-teaching-content-sha"
   end
 
+  def test_api
+    GetIntoTeachingApiClient::TypesApi.new.get_teaching_subjects
```
ethax-ross, line 13: You could hit the API health check endpoint here:

```
GetIntoTeachingApiClient::OperationsApi.new.health_check
```

spec/models/healthcheck_spec.rb
```diff
@@ -28,15 +29,27 @@
   include_examples "reading git shas", "app_sha", "/etc/get-into-teaching-app-sha"
   include_examples "reading git shas", "content_sha", "/etc/get-into-teaching-content-sha"
 
+  before do
```
ethax-ross, line 32: Worth testing the negative cases as well for `test_redis` and `test_api`?

ethax-ross (approved):

PR #378, opened 2020-09-24
Added spec for all pages checker

jebw:
### Context

We would like to check pages at least render before deploying 

1. Added feature spec to check the pages render

### Guidance to review

This spec will then also get run in the `-content` repos CI


peteryates (approved):
Looks good. `#strip_parents` though! :sweat_smile:

PR #380, opened 2020-09-24
Update event category copy

ethax-ross:
We have been provided with updated copy for the event category descriptions (both for the featured boxes and the category landing pages).

Removes use of `html_safe`.


jebw (approved):
Worth bearing in mind that you can also rename `key` to `key_html` in locales and it will automatically be marked html_safe

ethax-ross:
Huh, TIL! I'll merge this in for now but pick that up as some housekeeping the next time I'm making changes.

PR #381, opened 2020-09-24
Remove uses of html_safe

ethax-ross:
We should be using `simple_format` to safely allow whitelisted HTML elements and attributes.


jebw (approved):
LGTM :+1:

I've got a real dislike for `.html_safe` - way too much of a foot gun

PR #389, opened 2020-09-25
Point preprod environment to test API instance

ethax-ross:
The CRM team want to check the integration with the test CRM instance.


sfawcett123 (approved):
LGTM

PR #397, opened 2020-09-29
GITPB-723 VWO redirection

jebw:
### Context

We will be using VWO to handle redirecting traffic from the original site to the Private Beta site

1. Add the VWO smartcode - note this is an extended version that suggested which runs Analytics scripts only after VWO has completed
2. Updated all the analytics to only fire after VWO has completed and decided its not redirecting away

### DIFF
--- a/spec/javascript/controllers/analytics_spec_helper.js
+++ b/spec/javascript/controllers/analytics_spec_helper.js
@@ -102,47 +102,55 @@ export default class {
   }
 
   static describeWhenEventFires(name, controller, serviceFunctionName, cookieCategory) {
-    describe("when cookie:accepted event first", () => {
+    describe("when cookies have not yet been accepted", () => {
       beforeEach(() => {
         delete window.willRedirectionOccurByVWO ;
         Cookies.remove(CookiePreferences.cookieName)
         this.initApp(name, controller, "1234") ;
       })
 
-      describe("VWO has completed", () => {
+      describe("than VWO completes", () => {
         beforeEach(() => { window.willRedirectionOccurByVWO = false }) ;
 
-        it("Should not register the service", () => {
-          expect(typeof(window[serviceFunctionName])).toBe("undefined") ;
-        }) ;
-
-        it("Should register service when Cookie event is emitted", () => {
-          (new CookiePreferences).setCategory(cookieCategory, true) ;
-          expect(typeof(window[serviceFunctionName])).toBe("function") ;
-        }) ;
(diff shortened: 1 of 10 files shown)


spec/javascript/controllers/analytics_spec_helper.js
```diff
@@ -15,53 +15,135 @@ export default class {
...
+        })
+      }) ;
     })
   }
 
   static describeWhenEventFires(name, controller, serviceFunctionName, cookieCategory) {
-    beforeEach(() => { Cookies.remove(CookiePreferences.cookieName) })
-
-    describe("without cookie set yet", () => {
+    describe("when cookie:accepted event first", () => {
```
ethax-ross, line 105: I had to do a double take on this, I think it would be clearer as:

> when cookies have not yet been accepted

Also found specs on line 131 and 139 a little hard to follow, could we do anything to make the intention more explicit in the spec names, for example instead of:

> VWO then completes and emits event

something like:

> registers the service only after cookies are accepted and VWO completes

and:

> VWO then completes but is navigating away

could be:

> does not register the service if cookies are accepted and VWO redirects

spec/javascript/controllers/analytics_spec_helper.js
```diff
@@ -15,53 +15,135 @@ export default class {
...
-        expect(typeof(window[serviceFunctionName])).toBe("function") ;
+        it("Should register service when Cookie event is emitted", () => {
+          (new CookiePreferences).setCategory(cookieCategory, true) ;
+          expect(typeof(window[serviceFunctionName])).toBe("function") ;
+        }) ;
       }) ;
-    }) ;
+
+      describe("VWO has not yet completed", () => {
+        it("Should not register the service", () => {
```
ethax-ross, line 126: > Should not register the service when cookie event is emitted

ethax-ross (commented):
Looks good, just a couple of comments around some of the specs I found hard to follow

spec/javascript/controllers/analytics_spec_helper.js
jebw: Good call - this was me worrying more about figuring out the sequence last night and not paying enough attention to naming

spec/javascript/controllers/analytics_spec_helper.js
jebw: :eyes:

ethax-ross (dismissed):

PR #399, opened 2020-09-29
Bypass password for healthcheck

jebw:
We don't need to password protect the healthcheck endpoints.

Also made the password protection spec no longer depend upon the healthcheck endpoint


sfawcett123 (approved):
LGTM

PR #401, opened 2020-09-29
GITPB-689 LogIt integration

jebw:
### JIRA ticket number

GITPB-689

### Context

We should be generating logs which are useful to LogIt - initially that means single line logs, or logs where the multiple parts of a request can be grouped togther

1. Make STDOUT logs synchronous
2. Only load Semantic logging if either SEMANTIC_LOGGING or CF_INSTANCE_GUID are set
3. Make puma reopen the log output when if it forks a worker if running with semantic logging

### Notes

1. I've tested to make sure param filtering is honoured - visiting `http://localhost:3000/?first_name=Joe` shows as `[FILTERED]`

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -43,7 +43,7 @@ Rails.application.configure do
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :debug
+  config.log_level = :info
 
   # Prepend all log lines with the following tags.
   config.log_tags = [:request_id]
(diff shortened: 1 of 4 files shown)


ethax-ross (dismissed):

config/environments/production.rb
```diff
@@ -76,11 +76,11 @@
   # require 'syslog/logger'
   # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new 'app-name')
 
-  if ENV["RAILS_LOG_TO_STDOUT"].present?
-    logger           = ActiveSupport::Logger.new(STDOUT)
-    logger.formatter = config.log_formatter
-    config.logger    = ActiveSupport::TaggedLogging.new(logger)
-  end
+  # if ENV["RAILS_LOG_TO_STDOUT"].present?
```
ethax-ross, line 79: Is this intentionally commented out? can we just delete it?

config/environments/production.rb
```diff
@@ -110,4 +110,20 @@
...
   config.x.google_maps_key = ENV["GOOGLE_MAPS_KEY"].presence || \
     Rails.application.credentials.google_maps_key.presence
+
+  # Configure Semantic Logging for production environments
+  # This cannot be conditionally loaded so we use it all the time in production
+  # like environments.
+  #
+  # The rails_semantic_logger gem overwrites the log initializer code and by
+  # this point its too late to monkey patch that
+  STDOUT.sync = true
```
ethax-ross, line 120: Why does this need to be synchronous?

jebw: My experience on other stdout logging is that there can be a significant lag flushing. If the log msg relates to a problem and the service is restarted/node recycled than you never end up with the log info? This is stdout so I wasn't anticipating it being a significant block?

No strong feelings though?

ethax-ross: Cool, no I'm fine with that, I was just interested in the reason

ethax-ross (approved):

PR #413, opened 2020-10-02
Expose ActiveSupport metrics to Prometheus

ethax-ross:
- Add Prometheus::Client

Add the `Prometheus::Client` for aggregating and exporting metrics on the `/metrics` endpoint.

Configured to use `DirectFileStore` as this supports Puma forking additional processes (by default metrics will be summed together; if we use a gauge metric this may not be desirable). The DirectFileStore is the recommended approach for applications with pre-fork servers/multi-processors, but there are caveats:

1. When scraped by Prometheus the store reads all the files into memory (one file per metric per process), which can bump up the memory usage a fair amount. I don't anticipate this being an issue for us yet/soon given the amount of metrics we're collecting is small.

2. The metric files need to be cleared out each time the application starts to avoid a new instance exporting metrics from a previous one; I've done this programatically on start but since we're dockerized it probably won't matter anyway.

3. A large number of files can be created; there's no real solution for this one in the client library. If we're on the way to running out of disk space we should be alerted by infrastructure monitoring, but I imagine worst-case scenario is the container runs out of memory, blows up and a new one spins up to take it's place. It would only result in a problem if it happens on all instances at the same time - if we're worried about that then we could schedule periodic restarts of the instances to keep the usage low.

- Add ActiveSupport::Notifications instrumentation

Add metrics for recording notifications exposed from ActiveSupport, such as request durations, view rendering runtimes and cache reads.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,8 @@ gem "dotenv-rails"
 
 gem "govuk_design_system_formbuilder"
 
+gem "prometheus-client"
+
 gem "sentry-raven"
 
 gem "get_into_teaching_api_client_faraday", github: "DFE-Digital/get-into-teaching-api-ruby-client", require: "api/client"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -162,6 +162,7 @@ GEM
     parallel (1.19.1)
     parser (2.7.1.3)
       ast (~> 2.4.0)
+    prometheus-client (2.1.0)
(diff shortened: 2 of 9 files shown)


jebw (approved):
Looks good to me now that Prometheus accesses the end points directly.

Would be nice to add CPU, Ram and Disk to the metrics but those could be a subsequent PR?

PR #415, opened 2020-10-05
Remove "No Preference" and "Other" from preferred teaching subject options

ethax-ross:
- Remove No Preference/Other from MailingList::Steps::Subject
- Remove No Preference/Other from Event::Steps::PersonalisedUpdates

We no longer want "No Preference" or "Other" to be available values when selecting your preferred teaching subject.


DrLazer (approved):
cool, looks good to me

peteryates (commented):
Looks good.

PR #416, opened 2020-10-05
change cta banner colour

sigmaben:
Change to green as this is activation colour. from Branding feedback.


peteryates (approved):
Looks good

PR #417, opened 2020-10-05
Add extra copy to provider events

peteryates:
### JIRA ticket number

N/A

### Context

Small copy change for events of type 'School and University Event'

School or University events are now considered as though they can't be signed up for online. Instead of displaying any information in the 'How to apply' section the following text directs them to the event information section where they _should_ be given all the required info.

> To register for this event, follow the instructions in the event information section.

The 'Sign up for this event' button is also hidden when the event is a 'School or University Event'.

### Guidance to review

Does it look sane? I'm not sure the additional helper is an optimal solution here, it really should be a method on the event object (or a wrapper of it).

Screenshot from 2020-10-05 13-01-43

### DIFF
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -26,7 +26,7 @@ module EventsHelper
   end
 
   def can_sign_up_online?(event)
-    event.web_feed_id && event_status_open?(event)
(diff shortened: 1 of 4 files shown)


DrLazer (approved):
Looks good to me

PR #420, opened 2020-10-06
Prefix metrics with "app_" to distinguish them from other apps

ethax-ross:
All of our metrics are scraped by the same Prometheus instance, so we need to differentiate the metric names to be able to easily reference them in Grafana. The API metrics are prefixed with "api_" so it makes sense to prefix the app metrics with "app_".


peteryates (approved):
:+1:

PR #421, opened 2020-10-06
Rescue on GetIntoTeachingApiClient::ApiError

ethax-ross:
The Api Client rescues errors and exposes them as an ApiError, we need to rescue on this as well as the underlying Faraday error.


jebw (changes requested):
Should rescue from both I think?

ethax-ross:
I don't think it can throw a Faraday error as it swallows them, but I'll double check 👀

jebw (approved):

PR #423, opened 2020-10-06
Only render the first 9 events per category on the events index

peteryates:
### JIRA ticket number

N/A

### Context

Now there's been an influx of online events the index page is a bit unwieldy. After a discussion, doing this in the frontend (versus the API) was deemed the most straightforward approach for the time being.

Only show the first 9 events per category

### Guidance to review

Is this the best approach? The associated spec being a request spec didn't feel quite right but provisioning the variables for a view spec seemed fairly complex.


app/controllers/events_controller.rb
```diff
@@ -70,4 +76,8 @@ def event_search_params
     (params[Events::Search.model_name.param_key] || defaults)
       .permit(:type, :distance, :postcode, :month)
   end
+
+  def cap_results?
+    params.dig("events_search").blank?
```
ethax-ross, line 81: Does this return true or false when someone does an 'empty search'? i.e. if they search but using the default selections of all types and nationwide? I _think_ we would want to cap in this scenario, but may be worth double checking with Myles. Either way it would be good for a test to clarify I think?

jebw: Just seen this - hadn't really seen this controller code before - would it make sense to either move it to the Search model, or better create a Categories PORO model - that could be initialised from an API response. And then the logic of the categorisation can get tested at unit test level?

peteryates: @jebw - think that might be phase two, I'm definitely planning on revisiting this controller with regards to splitting out some of the more complex functionality and generally thinning it out a bit. This seemed like a high priority fix to prevent the events page from being exceedingly long.

peteryates: @ethax-ross - yeah, Myles agreed. Just pushed with the `#cap_results?` logic updated. Thanks

ethax-ross (approved):

PR #424, opened 2020-10-06
Added logging to a second GTM container

jebw:
### Context

We may potentially need to integrate with a second Gtm Container

1. Reformatted and generalised some of the GTM code
2. Added a Gtm2 controller which inherits from the original Gtm controller
3. Updated the helpers to initialise that


peteryates (approved):
Overall this looks fine to me.

I think numbering them `1` and `2` could probably make things a little clearer still, but I guess this is more of a speculative PR.

PR #429, opened 2020-10-07
Combine regular and short event boxes

peteryates:
### JIRA ticket number

N/A

### Context

Combine the two event box components and allow the 'short' one to be toggled via the `condensed` param. Also take an opportunity to tidy the HTML a bit and fix a couple of display/alignment bugs.

The main point of these changes was to make the event box easier to deal with by
* only having one component rather than two very similar ones,
* to tidy the HTML and CSS slightly
* to fix some problems caused by long text appearing alongside icons
* change the divider `<div>` elements to `<hr>`

| Description | Before | After |
| --------------- | ---------- | ------- |
| Event box as displayed on `/events` | Screenshot from 2020-10-07 13-48-51 | Screenshot from 2020-10-07 13-49-15
| Event icon label alignment fix (note the B in 'Birmingham' compared to the 'O' in 'Online' above) | Screenshot from 2020-10-07 13-50-43 | Screenshot from 2020-10-07 13-50-55 |
| Event icon label overflow fix | Screenshot from 2020-10-07 13-52-12 | Screenshot from 2020-10-07 13-52-23 |

### Guidance to review

General sense check, have I taken the right approach here? Is there a better option than `display: table` for icon alignment (without going down the `grid`/`subgrid` route plus fallbacks for users of lesser browers!).


sigmaben (changes requested):
Align icons left with the text above.

peteryates:
Screenshot from 2020-10-08 14-12-50

sigmaben (approved):

PR #433, opened 2020-10-08
Prefix the classes on main footer with site-footer to ease selecting

peteryates:
Previously they were all contained in just a `<footer>` tag which made using footers on other elements more difficult. Now the main footer and everything related to it has a class (or prefix) of `site-footer`.


jebw (approved):

ethax-ross (approved):
Looks good - can be one for later but I believe we're trying to follow BEM naming convention, so it would be `site-footer__bottom`, I think (though there are lots of other places this needs changing, it's not consistent atm)

PR #435, opened 2020-10-08
element.getElementById causing console errors in chrome

DrLazer:
element.getelementById is causing errors in chrome
not sure how this call managed to get in there by replacing it with document.getelementById retains functionality, all tests still pass and also fixes the console errors.

added in an additional check to make sure the element exists also

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,5 @@
 #  DFE-Digital Get into Teaching website
 
-
-
-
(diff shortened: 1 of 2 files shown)


jebw (approved):
LGTM

peteryates (approved):
Looks good.

PR #436, opened 2020-10-08
Allow HTML in event descriptions

ethax-ross:
We want to allow basic HTML tags in event descriptions; this commit runs the description through `safe_html_format` instead of `safe_format` (which only converts line breaks to paragraphs).


peteryates (approved):
:+1:

PR #437, opened 2020-10-08
Point app to production API

ethax-ross:
**DO NOT MERGE** - pending CRM schema updates

Updates the URL and shared secret credential to point to the production API instance.


peteryates (approved):
:+1:

PR #439, opened 2020-10-09
Revert "Point app to production API"

ethax-ross:
Reverts DFE-Digital/get-into-teaching-app#437 - we're going to do this a little later so we can get content changes into prod first


peteryates (approved):
:+1:

PR #440, opened 2020-10-09
Update home featured section display to allow video popups

sigmaben:
add style so play icon displays correctly.


peteryates (approved):
Looks good to me

PR #442, opened 2020-10-09
Add an Events::GroupPresenter to remove complexity from controller

peteryates:
### JIRA ticket number

N/A

### Context

The `EventsController` was a little logic-heavy. This has been moved out to the new `Events::GroupPresenter` class that takes care of grouping, sorting and capping events.

* Move logic from the events controller to the events presenter
* Remove the row-by-row rendering of the events list and instead render it inline, using CSS to position the event cards
* Drop the event titles from `<h2>` to `<h4>` for a better document structure

### Guidance to review

There should be no noticeable changes to the app from a user perspective. Hopefully the categorising and grouping logic is easier to follow and everything makes sense!

### Preview

Screenshot from 2020-10-09 16-00-06

[Full page](https://user-images.githubusercontent.com/128088/95599261-c2611f00-0a48-11eb-81a7-b40453a977b5.jpg)

### DIFF
--- a/app/presenters/events/group_presenter.rb
+++ b/app/presenters/events/group_presenter.rb
@@ -1,0 +1,40 @@
+module Events
+  class GroupPresenter
+    attr_accessor :all_events
+
+    INDEX_PAGE_CAP = 9
+
+    def initialize(events, cap: false)
+      @all_events = events.sort_by { |e| [event_type_name(e.type_id), e.start_at] }
+      @cap        = cap
+    end
+
+    def get_into_teaching_events
+      group_by_type_id(all_events.select { |event| event.type_id.in?(get_into_teaching_type_ids) })
+    end
+
+    def school_and_university_events
+      group_by_type_id(all_events.reject { |event| event.type_id.in?(get_into_teaching_type_ids) })
+    end
+
+  private
+
+    def event_type_name(type_id)
+      GetIntoTeachingApiClient::Constants::EVENT_TYPES.values.index(type_id)
+    end
+
+    def group_by_type_id(events)
+      groups = events.group_by(&:type_id)
+
+      return groups unless @cap
+
+      groups.each.with_object({}) do |(type_id, events_in_group), capped_groups|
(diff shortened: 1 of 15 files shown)


peteryates:
Ignore the `<i>` thing that SonarCloud reports, I've added it to my list :bookmark_tabs:

peteryates:
I improved the look in older browsers that don't support CSS-grid by adding a Flexbox fallback. This looks good in IE10 and IE11. IE9 and older will fall back to a list view, which is still perfectly-useable. This page still works fine in IE5.

## IE10 and 11

Screenshot from 2020-10-10 15-30-36

## IE9 and below

Tested and it works all the way back to IE5!

Screenshot from 2020-10-10 15-32-46

## Mobile views

If there's still anyone using Windows mobile, here's how it should look.

Screenshot from 2020-10-10 15-52-55

## Edge (latest)

Screenshot from 2020-10-10 15-44-25

app/presenters/events/group_presenter.rb
```diff
@@ -0,0 +1,40 @@
...
+    def event_type_name(type_id)
+      GetIntoTeachingApiClient::Constants::EVENT_TYPES.values.index(type_id)
+    end
+
+    def group_by_type_id(events)
+      groups = events.group_by(&:type_id)
+
+      return groups unless @cap
+
+      groups.each.with_object({}) do |(type_id, events_in_group), capped_groups|
```
ethax-ross, line 31: I think this would be equivalent and a bit simpler to follow?

`groups.transform_values { |events| events.first(INDEX_PAGE_CAP) }`

ethax-ross (approved):
One take it or leave it comment, looks good 👏

app/presenters/events/group_presenter.rb
peteryates: I didn't know `#transform_values` existed, that's _way_ nicer. Thanks

ethax-ross (approved):

PR #447, opened 2020-10-12
Dependabot Step 2

jebw:
### JIRA ticket number

N/a

### Context

We rely on multiple sources of software, we should be checking them all for updates

1. Update the ruby checks to only run weekly
2. Check for NPM updates (using yarn)
3. Check for updates to the Docker base image

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,22 @@
 version: 2
 updates:
+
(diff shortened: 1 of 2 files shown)


peteryates (dismissed):

peteryates (approved):
Seems sensible

PR #448, opened 2020-10-12
Stop sending postcodes with extra spaces to API

peteryates:
### JIRA ticket number

N/A

### Context

It appears that postcodes with leading or trailing whitespace cause the API to return a ["'Postcode' is not in the correct format." ](https://sentry.io/organizations/dfe-bat/issues/1918399446/?project=5276949&referrer=slack) error.

`#strip` postcode values before sending to API

### Guidance to review

Is this the right approach? It would probably make sense for this to happen in Dynamics, but that might be a more substantial task :man_shrugging:


jebw (approved):
LGTM

PR #449, opened 2020-10-13
Disable analytics in PreProd environment

jebw:
### JIRA ticket number

N/a

### Context

The analytics are still enabled in the preprod environments for testing, but we also have a test suite which runs there and may skew analytics data.

1. Disabled the analytics in preprod


peteryates (approved):
Looks good. I guess commented out so toggling them on one-at-a-time is easier for debugging?

PR #450, opened 2020-10-13
Standardising the layouts

peteryates:
### JIRA ticket number

N/A

### Context

The structure of the pages had seemingly unnecessary differences caused by a `.container`. It's easier if they all follow the same format.

Remove the container for more consistency between the layouts. This should hopefully help avoid bugs like #445 

### Guidance to review

I've checked this side-by-side and can't spot any visual differences. There shouldn't be any but I'd be interested to hear if I've missed something.


jebw (approved):
LGTM

PR #451, opened 2020-10-13
Use fa_icon helper for the icons in the footer

peteryates:
Begin the move towards using the `#fa_icon` helper for font awesome icons. This was originally going to be the start of removing `<i>` tags from everywhere [where they're not semantically-correct](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i) but that's a _big job_ (the content repo is full of them!), so let's start small.

Also add some tests for the helper and allow [icon styles to be set](https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use).

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -55,9 +55,12 @@ module ApplicationHelper
     end
   end
 
-  def fa_icon(icon_name, *additional_classes)
-    classes = ["fas", "fa-#{icon_name}"] + additional_classes
-    content_tag :i, "", class: classes.join(" ")
+  # FA supports several styles:
+  # fas = solid, fab = brand, far = regular, fal = light, fad = duotone
+  # https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use
+  def fa_icon(icon_name, *additional_classes, style: "fas")
+    classes = [style, "fa-#{icon_name}"] + additional_classes
+    tag.span("", class: classes)
   end
   alias_method :fas, :fa_icon
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -62,7 +62,14 @@ module ApplicationHelper
     classes = [style, "fa-#{icon_name}"] + additional_classes
     tag.span("", class: classes)
   end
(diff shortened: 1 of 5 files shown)


app/helpers/application_helper.rb
```diff
@@ -55,9 +55,12 @@ def hero_image_path(imgpath)
...
-    classes = ["fas", "fa-#{icon_name}"] + additional_classes
-    content_tag :i, "", class: classes.join(" ")
+  # FA supports several styles:
+  # fas = solid, fab = brand, far = regular, fal = light, fad = duotone
+  # https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use
+  def fa_icon(icon_name, *additional_classes, style: "fas")
+    classes = [style, "fa-#{icon_name}"] + additional_classes
+    tag.span("", class: classes)
   end
   alias_method :fas, :fa_icon
```
jebw, line 65: Would it be better to replace this alias with

```ruby
  def fas(*args)
    fa_icon(*args, style: 'fas')
  end

  def fab(*args)
    fa_icon(*args, style: 'fab')
  end
```

app/views/sections/_footer.html.erb
```diff
@@ -3,26 +3,26 @@
   <div class="site-footer-top">
     <div class="site-footer-top__social">
       <a href="https://www.facebook.com/getintoteaching" class="site-footer-top__social__link" target="_blank" rel="noopener noreferrer">
-        <i class="fab fa-facebook-f"></i>
+        <%= fa_icon("facebook-f", "icon", style: "fab") %>
```
jebw, line 6: That would than simplify the calls the the helper

```erb
<%= fab 'facebook-f', 'icon' %>
```

jebw (approved):
Approving because it improves the warnings from Sonar but feels like a more specific helper would clean it up further

app/helpers/application_helper.rb
peteryates: Makes sense. I think I'd go with `#fab_icon` though - `fas`/`fab` aren't too obvious.

jebw (approved):
LGTM

PR #459, opened 2020-10-14
Add a wrapper that builds story using frontmatter

peteryates:
### [Trello Ticket](https://trello.com/c/UGfAWxXR/398-move-the-story-content-from-html-in-markdown-files-to-pure-markdown)

[Corresponding work on the content](https://github.com/DFE-Digital/get-into-teaching-content/pull/91).

### Context

Currently the stories are very HTML-heavy Markdown files. Much of the HTML is repeated across stories so this is an effort to move as much of the repeated code as possible to the templates and populate them with data stored in the frontmatter.

The `build_layout_from_frontmatter` flag allows this approach to be toggled on a per-story basis. It'll allow us to migrate slowly without having to have a big bang release.

Allow stories to be written purely in markdown and their surrounding data to be supplied from the frontmatter.

### Guidance to review

Does everything make sense? Better name suggestions for the toggle flag would be apprecieated :sweat_smile: 

Notes:

* The frontmatter is only used to build layouts that have a flag in the frontmatter called `build_layout_from_frontmatter` that is truthy. Stories that don't have that flag should be rendered as-before
* The new CSS applies to new stories only. They appear inside `article.story`, the old ones are rendered in a `div.stories` which is repeated in every story markdown file

Here's a WIP preview, markdown on the left and old HTML on the right.

Screenshot from 2020-10-14 13-35-12

## Final tasks

* [x] The back links are currently all hardcoded to the text for 'Career changers', this needs to be moved to the frontmatter so it can be set on a per-story-type basis
* [x] The links at the bottom of international stories (eg [this one](https://beta-getintoteaching.education.gov.uk/life-as-a-teacher/my-story-into-teaching/international-career-changers/returning-to-teaching-with-support-from-an-adviser)) aren't styled properly yet
* [x] Replace the fancy quotes with regular ones as they'll be added by Kramdown automatically

### DIFF
--- a/app/helpers/stories_helper.rb
+++ b/app/helpers/stories_helper.rb
@@ -1,13 +1,11 @@
 module StoriesHelper
   def story_heading(front_matter)
+    delimiter = ("," + tag.br + "\n").html_safe
+
     tag.h2 do
       safe_join(
-        [
-          front_matter.dig("story", "teacher"),
-          ",",
-          tag.br,
-          front_matter.dig("story", "position"),
-        ],
+        [front_matter.dig("story", "teacher"), front_matter.dig("story", "position")].compact,
+        delimiter,
       )
     end
   end
--- a/app/helpers/stories_helper.rb
+++ b/app/helpers/stories_helper.rb
@@ -12,6 +12,10 @@ module StoriesHelper
     end
   end
 
+  def story_image_alt(name)
+    "A photograph of #{name}"
+  end
+
   def youtube(url)
     tag.iframe(
       "",
--- a/app/helpers/stories_helper.rb
+++ b/app/helpers/stories_helper.rb
@@ -24,4 +24,11 @@ module StoriesHelper
       class: "story__video",
     )
   end
+
+  def more_stories_thumbnail(path)
+    tag.div(
+      class: "more-stories__thumbs__thumb__img",
+      style: %(background-image:url('#{path}')),
+    )
+  end
 end
--- a/app/helpers/stories_helper.rb
+++ b/app/helpers/stories_helper.rb
@@ -1,0 +1,27 @@
+module StoriesHelper
+  def story_heading(front_matter)
+    tag.h2 do
+      safe_join(
+        [
+          front_matter.dig("story", "teacher"),
+          ",",
+          tag.br,
+          front_matter.dig("story", "position"),
+        ],
+      )
+    end
+  end
+
+  def youtube(url)
+    tag.iframe(
+      "",
+      width: 560,
+      height: 315,
+      src: url,
+      frameborder: 0,
+      allow: "accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture",
+      allowfullscreen: true,
+      class: "story__video",
+    )
+  end
+end
--- a/app/views/layouts/stories.html.erb
+++ b/app/views/layouts/stories.html.erb
@@ -14,7 +14,7 @@
                 <%= tag.h1(@front_matter["title"]) %>
 
                 <header class="story__header">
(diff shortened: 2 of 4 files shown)


peteryates:
Update, now with 'More stories'

Screenshot from 2020-10-14 15-14-37

peteryates:
Added support for blockquotes generated by markdown quote syntax (`>`). Quote icons are positioned entirely using CSS.

Screenshot from 2020-10-14 16-38-11

peteryates:
The last change, 6ea28de, change the mobile styles of block quotes. When the screen is narrower than 500px the block quote now expands to the full width. This prevents text from being squished up against the right side of the page and flow being compromised.

| New (Markdown) | Old (HTML) |
| ------------- | ----------- |
|  Screenshot from 2020-10-15 10-26-39 | Screenshot from 2020-10-15 10-27-04 |

peteryates:
Another tweak to the `blockquote` element style to make them match the old design more-closely. Here's how they look now, new on the left old on the right.

Screenshot from 2020-10-19 12-43-02

app/views/layouts/stories.html.erb
```diff
@@ -2,13 +2,69 @@
...
-            <%= yield %>
-        </main>
-        <%= render "sections/footer" %>
-        <%= render "components/videoplayer" %>
-        <%= render "sections/cookie-acceptance" %>
+      <%= render "sections/header" %>
+      <main class="content container-1000" role="main" id="main-content">
+        <% if @front_matter.dig("build_layout_from_frontmatter") %>
+
+          <div class="content-wrapper">
```
jebw, line 9: Is it worth splitting this into a partial just to reduce the nesting complexity?

app/views/layouts/stories.html.erb
```diff
@@ -2,13 +2,69 @@
...
+                <%= yield %>
+
+                <% if @front_matter["more_information"] %>
+                  <%= link_to(@front_matter.dig("more_information", "text"), @front_matter.dig("more_information", "link"), class: "git-link") %> <%= fas_icon "chevron-right" %>
+                <% end %>
+              </article>
+            </div>
+          </div>
+
+          <% if @front_matter.dig("more_stories").present? %>
```
jebw, line 36: Likewise this could become `<%= render 'more_stories' if @front_matter.dig("more_stories).present? %>`

jebw (dismissed):
I've approved but I've made some suggestions on a couple of tweaks that could improve clarity.

Another point would be from the screen grabs, should the yellow blockquotes have the same margin below as they have to the side? At the moment the paragraph text below looks like its hard up against the lower border of the block quote?

app/views/layouts/stories.html.erb
peteryates: This is laying the groundwork for the next PR which will move it to a view component and add a load of tests.

app/helpers/stories_helper.rb
```diff
@@ -0,0 +1,38 @@
+module StoriesHelper
+  def story_heading(front_matter)
+    tag.h2 do
+      safe_join(
```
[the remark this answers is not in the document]
peteryates: Ah, yeah - that's definitely nicer.

jebw (approved):

peteryates:
> I've approved but I've made some suggestions on a couple of tweaks that could improve clarity.
> 
> Another point would be from the screen grabs, should the yellow blockquotes have the same margin below as they have to the side? At the moment the paragraph text below looks like its hard up against the lower border of the block quote?

Counter-intuitively it's the new one on the left of the screengrab and the old on the right. I improved margins on the `blockquote` styling and simplified the CSS. As a side-effect, the text flow is generally better now, text isn't squished up when making the page smaller.

PR #460, opened 2020-10-15
Update cve.yml

sfawcett123:
Allow the workflow to be run manually.
Changed to version 2 of the CVE action

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,22 @@
 version: 2
 updates:
+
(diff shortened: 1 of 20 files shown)


sfawcett123:
There is a HIGH CVE error being raised with the yargs-parser causing the CVE to fail. This seems to be highlighted in V2, where it wasn't displaying in v1.
There is a dependabot alert for the same problem.  
So although related it is not part of this issue and a bit beyond my scope ( capabilities to resolve )

PR #461, opened 2020-10-15
Discover image

jebw:
### JIRA ticket number

N/a

### Context

Currently only changes to the content repo will result in a release of a new image through to production. Changes in this repo will be ignored at the point of deploying to TEST and PROD environments.

This is caused by TEST and PROD only having access to the Git SHA for the Content repo. This SHA is insufficient to correctly identify the appropriate docker image so a docker image with only the Content SHA is used. When the build occurs a new image is built but with the same docker image tag (because it uses only the Content SHA and that has not changed).

When terraform comes to deploy, because the docker image tag has not changed, it will succeed but not actually pull the new image. This is ineffect the same issue as when pipelines try to deploy `:latest`

1. Rather than dispatching an event to the Content repo, instead the FROM line in Dockerfile is changed and a new commit created. This will trigger the pipeline the same, but deploying the single sha version is now succesfully because the Sha in the Content repo has also changed as a result of the commit.

2. Removed using curl to trigger an event dispatch in the content repo

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -189,9 +189,18 @@ jobs:
           cd content
           git show HEAD
 
-      - name: Push the commit
-        env:
-          API_TOKEN_GITHUB: ${{ secrets.BUILD_CONTENT_TOKEN }}
+      # - name: Push the commit
+      #   env:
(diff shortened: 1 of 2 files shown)


jebw:
Implementing over in the content repo instead so closing here

PR #486, opened 2020-10-20
Add acronyms support

jebw:
### Trello card

https://trello.com/c/0J164eou

### Context

We should be providing expansion for accronyms. This means marking acronyms with the `<abbr>` and expanding them where appropriate.

1. Added a helper class, `Abbreviations` to parse HTML and replace abbreviations in text nodes, but not in attributes.
2. Update the Markdown render to filter content through the Abbreviations filter
3. Added support for a global frontmatter file, to which the per page front matter file is deep merged - this allows combining acronyms defined in both locations.

### Guidance to review

There will be a follow on PR to explain this together with some default acronyms in the content repo after this is merged

### DIFF
--- a/lib/abbreviations.rb
+++ b/lib/abbreviations.rb
@@ -1,0 +1,41 @@
+class Abbreviations
+  ABBR_REGEXP = %r{\b([A-Z][A-Z]+)\b}.freeze
+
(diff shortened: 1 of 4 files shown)


ethax-ross (approved):
LGTM

PR #493, opened 2020-10-21
Take first token of violated-directive

ethax-ross:
The `violated-directive` of the CSP report sometimes comes through with more information that is useful for aggregating on:

``` violated-directive: img-src *.google.com *.gov.uk ```

Instead, we only care about the root directive/first token `img-src`. This will let us get a better breakdown of which CSP  directives are being violated most frequently.


peteryates (approved):
Looks good

PR #494, opened 2020-10-21
Add HotJar cookies to website usage section

ethax-ross:
Add the HotJar cookies that will be set for users that are visiting the website to the cookies page.


peteryates:
The copy sounds like it's aimed at _us_ rather than our users.

peteryates (approved):
Accept as it's ok but I think the copy should be tweaked sooner rather than later

peteryates (approved):
Looks good

PR #497, opened 2020-10-22
Feature/420 small feedback banner

peteryates:
### JIRA ticket number

[420](https://trello.com/c/xUrGnxcX/420-feedback-design-create-smaller-pop-up)

### Context

Change from a feedback popup to a small bar that appears at the bottom of the screen when the user has visited more than two pages.

Screenshot from 2020-10-22 12-07-46

* remove the feedback popup
* add a feedback bar that's hidden by default
* add a counter that increments on every page view until the feedback bar is dismissed
* the feedback bar is shown **if** the view counter is greater than or equal to 3 and the feedback bar has not been dismissed already
* dismissing happens either when the 'hide' link or 'give us feedback' links are clicked

### Guidance to review

Is this sane? It's a bit light on tests because I don't appear to be able to get at `window.localStorage` from the javascript specs.


app/webpacker/controllers/feedback_controller.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  // we'll probably want to record the fact that
+  // someone's followed the feedback link
+  giveFeedback() {
+    this.dismiss();
+  }
+
+  incrementFeedbackPageCounter() {
+    const currentValue = this.pageViewCount;
```
ethax-ross, line 37: I think it would be clearer returning `0` from `this.pageViewCount` here instead of dealing with null, then you can drop the conditional and the `pageViewCount` getter becomes:

```
let count = window.localStorage.getItem(countKey) || 0;
return parseInt(count);
```

Don't feel that strongly about it though.

ethax-ross (commented):
Looks good, one take it or leave it suggestion. I think we can also drop the `is-touch-device` dependency now as well.

A few more tests on the Stimulus controller and it'll be good to go 👍

spec/javascript/controllers/feedback_controller_spec.js
```diff
@@ -0,0 +1,139 @@
...
+
+  describe("clicking 'Give us feedback on this website'", () => {
+    beforeEach(() => {
+      const application = Application.start();
+      application.register('feedback', FeedbackController);
+    })
+
+    it('hides the feedback bar', () => {
+      const feedbackBar = document.getElementById('feedback-bar');
+      const hideLink = document.getElementById('hide-feedback-bar');
```
ethax-ross, line 57: Should this not target the `/some/page` anchor?

peteryates: :man_facepalming:

ethax-ross (approved):
LGTM

PR #498, opened 2020-10-22
Rate limit pin code generation

ethax-ross:
### JIRA ticket number

[Trello-352](https://trello.com/c/MjWfjLi7/352-set-rate-limiting-on-api)

### Context

The API will begin rate limiting requests that issue a verification code. In order to do that we need to send an `X-Client-IP` header with the `create_access_token` requests. We should also display a suitable message to the user if they issue too many requests.

- Retain client IP address in Wizard

We want to be able to access the end user IP address from the wizard step that makes the request to the `create_candidate_access_token` endpoint. If we pass this as the `X-Client-IP` header value it will cause the API to rate limit the endpoint for the individual user, instead of for the app as a whole (based it's API key).

- Update get_into_teaching_api_client_faraday

Add support for passing `x_client_ip` option to API requests.

- Send x_client_ip with access token requests

We want the API to be able to rate limit candidate access token requests by the end user that is using the website; it determines the end user based on the `x_client_ip` header, if provided.

- Display message when verification code rate limit exceeded

When the user attempts to request a verification code more than 3 times/minute the API will respond with `429` - too many requests.

We catch this response and display a suitable message to the user, asking them to wait a minute before attempting to request another verification code.

### Guidance to review

**NOTE: the API is not yet rate limiting requests; we are going to ship this ahead of enforcing the rate limits, but this will be user tested as part of that work**

Screenshot 2020-10-22 at 15 18 56

Screenshot 2020-10-26 at 16 06 13

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,11 +1,11 @@
 GIT
   remote: https://github.com/DFE-Digital/get-into-teaching-api-ruby-client.git
-  revision: b790c3adf3f7bf8aae0335c8c4b76d99b0326265
+  revision: de4e6b9d429a3615f88b988d3d766b626e229fe4
   specs:
-    get_into_teaching_api_client (1.1.3)
+    get_into_teaching_api_client (1.1.4)
       json (~> 2.1, >= 2.1.0)
       typhoeus (~> 1.0, >= 1.0.1)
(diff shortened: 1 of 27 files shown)


ethax-ross:
Closing as we're taking a different appraoch.

PR #499, opened 2020-10-23
Extract path from document_uri

ethax-ross:
Using the path as the label (removing the hostname, which is always the same for this attribute) makes the visualisation nicer in Grafana.


peteryates (approved):
Looks good

PR #501, opened 2020-10-23
Tweak to not squish wide images in story headers

peteryates:
Remove the copied CSS and replace with just `object-fit: cover`, the image sizing's already taken care of and anything that overflows will automatically be cropped.


ethax-ross (approved):
I don't think IE will support this, do we need to account for that?

peteryates:
We have <0.5% IE users. As this will only have a minor effect on a couple of pages where wider story pictures might be served in the wrong aspect ratio I'll merge. All the pages will still be usable.

PR #507, opened 2020-10-26
Render 404 on Wizard::UnknownStep

ethax-ross:
We currently fail with a 500 error if the wizard is initialised with an unknown step; this happens if the user enters a bad URL, so it would make more sense to return a 404 in this scenario.


peteryates (approved):
:+1:

PR #508, opened 2020-10-26
Remove 'See all events' link from category pages

peteryates:
The link links to the current page, it can be toggled via
show_see_all_events and is off by default.

### JIRA ticket number

https://trello.com/c/rIzaetoX/478-implement-the-new-functionality-in-schools-and-university-see-all-events-page-search-pagination-reduced-info-being-shown

### Context

Now the partial is being shared the display of the button needs to be controlled.

Add a `show_see_all_events` argument to the `event_group` partial and default it to false. Set it to true only on the events index page.

### Guidance to review

Does it look sensible? The next step is probably to wrap this in a component and pull the logic out of the template.

### DIFF
--- a/app/views/events/_event_category.html.erb
+++ b/app/views/events/_event_category.html.erb
@@ -4,6 +4,6 @@
   </div>
 
   <% category_groups.each do |type_id, events| %>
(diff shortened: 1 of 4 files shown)


ethax-ross (approved):
Looks sensible 👍

PR #512, opened 2020-10-27
Convert story layout to a component and add specs

peteryates:
Move the story part of the template to its own component and add some specs.


ethax-ross (approved):
Looks good 👍 

One for later, but we're using a mixture of `front_matter` and `frontmatter` - we should probably settle on a convention at some point and make it all consistent.

PR #513, opened 2020-10-27
Add and configure a stories landing page layout

peteryates:
This page will be built entirely from the frontmatter provided in the content repo. Much of the CSS is now shared with the other pages, the unused rules can now be safely culled. I've also added a little more spacing between the sections to make it clearer which links belong to which.

Related (and depends on) to DFE-Digital/get-into-teaching-content/pull/118

## Examples

Screenshot from 2020-10-27 14-28-53

Screenshot from 2020-10-27 14-31-08


ethax-ross (commented):
Worth a quick request spec to test the new route/action?

ethax-ross (approved):

PR #514, opened 2020-10-27
Allow HTML in event descriptions

ethax-ross:
Reverts DFE-Digital/get-into-teaching-app#438

We want to allow basic HTML tags in event descriptions; this commit runs the description through safe_html_format instead of safe_format (which only converts line breaks to paragraphs).

We're going to ship this to dev again; testing it again it seems to work now so Myles is going to give it a thorough test on dev.

Screenshot 2020-10-27 at 16 04 48


peteryates (approved):
Makes sense.

PR #515, opened 2020-10-28
Support both HTML and plain text event descriptions

ethax-ross:
The current event descriptions are plain text (with line breaks) - the CRM team are  migrating to a WYSIWYG for inputting event descriptions that will return HTML. We need to be able to support both while they perform the migration.

Checks to see if the event description contains any HTML tags, if it does then it runs it through `sanitize`, otherwise continue to use `simple_format`.

If we run through both `simple_format` and `sanitize` instead we end up getting extra spacing between lines.

### DIFF
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -37,6 +37,14 @@ module EventsHelper
     video_url&.sub("watch?v=", "embed/")
   end
 
+  def formatted_event_description(description)
+    if strip_tags(description) != description
+      safe_html_format(description)
+    else
+      safe_format(description)
+    end
+  end
+
   def event_has_provider_info?(event)
     event.provider_website_url ||
       event.provider_target_audience ||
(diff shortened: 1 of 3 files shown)


app/helpers/events_helper.rb
```diff
@@ -37,6 +37,13 @@ def embed_event_video_url(video_url)
     video_url&.sub("watch?v=", "embed/")
   end
 
+  def formatted_event_description(description)
+    contains_html = strip_tags(description) != description
+    return safe_html_format description if contains_html
+
+    safe_format(description)
```
peteryates, line 44: Could this be expressed more clearly with a regular `if` statement?
```
    if strip_tags(description) != description
      safe_html_format(description)
    else
      safe_format(description)
    end
```

peteryates (dismissed):
Looks good. One bit that might warrant a second look.

peteryates (approved):

PR #516, opened 2020-10-28
Display upcoming teaching events irrespective of month

ethax-ross:
### JIRA ticket number

[Trello-427](https://trello.com/c/6YY80hFQ/427-events-for-virtual-ttt-events-show-next-months-events-if-none-exist-for-that-month-on-home-page)

### Context

We want to display the next 9 upcoming events by type when landing on the events index page, regardless of month.

We also want to update the search results to only show a maximum of 9 events per type.

The event category page should continue to display all results.

There are two new endpoints in the API for retrieving events indexed by type and limiting the results per type. The current `search_teaching_events` endpoint is being deprecated.

- Bump API client version

Update includes new `*_indexed_by_type` API endpoints.

- Transition events#show_category to new search method

We are transitioning away from the `search_teaching_events` method to the new `search_teaching_events_indexed_by_type` as events are categorised in the majority of places.

Updates `events#show_category` to use new categorised events method. The amount is set somewhat arbitrarily to 1,000 for now, we are expecting to change this in the future to be paginated.

- Migrate index/show to search_teaching_events_indexed_by_type

Updates the `Events::Search` model to use the new search endpoint. This now returns events grouped by type and also supports limiting results, which allows us to simplify the the `Events::GroupPresenter`.

We have also decided to limit both the events index page and search results to at most 9 events per group.

- Switch index page to use upcoming_search_events_indexed_by_type

We want to display the upcoming events regardless of month when landing on the events index page, up to a maximum of the next 9 by type.

The event search is also limited to 9 per type but the events will not cross month boundaries.

### Guidance to review

Best reviewed by commit.

The event category results have a somewhat arbitrary limit of 1,000 - this is just because the API expects a value, so it should be high enough to ensure all events are returned. We expect to migrate away from this to a paginated results set in the near future.

More can probably be done to simplify the `EventsController`, but I'd rather this was picked up in a follow up PR.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,11 +1,11 @@
 GIT
   remote: https://github.com/DFE-Digital/get-into-teaching-api-ruby-client.git
-  revision: b790c3adf3f7bf8aae0335c8c4b76d99b0326265
+  revision: b54d486736e47a88d9d86e7080891f564c4f9dc3
   specs:
-    get_into_teaching_api_client (1.1.3)
+    get_into_teaching_api_client (1.1.5)
       json (~> 2.1, >= 2.1.0)
       typhoeus (~> 1.0, >= 1.0.1)
-    get_into_teaching_api_client_faraday (0.1.4)
+    get_into_teaching_api_client_faraday (0.1.6)
       activesupport
       faraday
       faraday-encoding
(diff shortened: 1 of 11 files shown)


peteryates (approved):
Looks good to me.

PR #518, opened 2020-10-29
Rate limit transactions

ethax-ross:
### JIRA ticket number

[Trello-352](https://trello.com/c/MjWfjLi7/352-set-rate-limiting-on-api)

### Context

The API is rate limited to 30 requests/client API key/minute for transactional/vulnerable endpoints (both apps currently share the same API key, but this will change in the future).

We want to apply per-user/IP rate limiting in the front end applications as well, to add more granular control over submissions. The rate limits initially will be 5 requests/IP/minute - this should be plenty for now, but may need tweaking as we roll out to more users (or to take into account advertising 'spikes' in traffic).

When a user encounters a rate limit we want to display a 429 error page similar to the existing 404/500 error pages.

If the user encounters an API rate limit we want to display the same page.

- Rate limit transaction endpoints

The transaction endpoints are rate limited to 30 requests per minute by the API. We want to enforce the rate limiting client-side per user/ip as well to a maximum of 5 requests/IP/minute.

If a user encounters a rate limit, instead of returning a standard 429 and text response we instead render a 'Too many requests' error page, similar to the existing 404 and 500 error pages.

- Rescue ApiError and render 429 error page

Whilst the app itself is rate limited the API also has rate limiting. In the unlikely event that the client rate limits don't trigger and instead the API's rate limiting is enforced we want to display the same 429 error page to the user.

Raise `ApiError` up to the `ApplicationController` when the `code` is 429. The `ApplicationController` will then rescue the error and render the error page.

### Guidance to review

Screenshot 2020-10-29 at 14 06 59

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,6 +2,7 @@ class ApplicationController < ActionController::Base
   include UtmCodes
 
   rescue_from ActionController::RoutingError, with: :render_not_found
+  rescue_from GetIntoTeachingApiClient::ApiError, with: :handle_api_error
 
   before_action :http_basic_authenticate
   before_action :record_utm_codes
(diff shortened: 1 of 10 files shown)


peteryates (approved):
Nice.

PR #519, opened 2020-10-29
Improve events category mobile view

peteryates:
### JIRA ticket number

https://trello.com/c/rgYheuRh/477-design-develop-new-ttt-online-schools-and-university-events-page-with-pagination-search-functionality-and-removal-of-categories

### Context

Users on mobile can only see one or two events at once and when on the events category page (which shows all the events) it can be overwhelming.

#### Make the event card smaller by hiding the description on mobile.

Screenshot from 2020-10-29 14-36-09

#### Filtering the category page

* Filtering is similar to searching except
  * there's an option to not filter by month, this is the default
  * type isn't available to filter by, the user is already on a 'type' page and are filtering within that type.

Screenshot from 2020-10-29 14-32-37

Screenshot from 2020-10-29 14-38-20

### Guidance to review

Is this the right approach? Pagination will follow in a separate PR.


app/components/events/search_component.html.erb
```diff
@@ -0,0 +1,47 @@
...
+
+      <%= tag.div(class: input_field_classes(:month)) do %>
+          <%= f.label :month, "Month", class: "search-for-events__content__input__label" %>
+          <%= f.select :month, f.object.available_months, **month_args %>
+      <% end %>
+    </div>
+
+    <% if f.object.errors.any? %>
+      <div class="search-for-events__content__errors">
+        <% f.object.errors.messages.values.flatten.each do |error_message| %>
```
ethax-ross, line 37: I think it would be cleaner to move some of these `f.object.x` calls up into the view component, where they could also be tested - for example this one could be:

```
def error_messages
  @search.errors.messages.values.flatten
end
```

app/models/events/search.rb
```diff
@@ -51,27 +51,35 @@ def available_months
     end
 
     def query_events
```
ethax-ross, line 53: Having the two search methods feels a little confusing here. I wonder if we really need to enforce the `presence` on month in the model, given it's only picking up on the scenario when someone faffs with the HTML intentionally. If we dropped it then we could revert to a single method:

```
def query_events(limit = RESULTS_PER_TYPE)
  valid ? query_events_api(limit) : {}
end
```

Or if we keep the conditional validation I think we need to make the naming of the use cases clearer as `query_events` and `filter_events` are a bit misleading IMO.

ethax-ross (commented):
Looking good overall, just a couple of suggestions.

app/components/events/search_component.html.erb
peteryates: Yeh, that's definitely nicer.

app/models/events/search.rb
peteryates: Sounds like a good move. I wasn't sure how important the validation on time was so left it.

ethax-ross (approved):
Looks good to me 👍

PR #520, opened 2020-10-30
Sanitize metrics for better aggregation

ethax-ross:
Aggregating the metrics on path doesn't provide a good insight at present due to the query parameters being included in the path. By removing the query parameters we can aggregate on the underlying path/action.

Similarly, the view metrics contain the absolute path to the views directory which makes displaying them in a table awkward. Removing the `/path/to/app/views/` prefix leaves us with the template identifier.


peteryates (approved):
Looks good. As it's aware of individual partials does it also need to be aware of components?

ethax-ross:
I don't think it will report on view components as they're not Rails - they'll likely just be bundled up in which ever view/partial they are rendered from

PR #527, opened 2020-11-02
Restore the missing APP SHA in the /etc file

jebw:
### Trello card

https://trello.com/c/9lMe3WsU

### Context

Currently the App sha is not being reported in the App

1. Corrected the variable name

Wondering if this would be better as part of the Docker build file (copy file relevant ref out of `.git`) but this is easier for now


sfawcett123 (approved):
LGTM

PR #528, opened 2020-11-02
Add config-based redirection route building

peteryates:
### JIRA ticket number

N/A

### Context

Recently we had an issue where paths changed and the old locations were referenced in emails; recipients following these links saw a 404.

Allow us to maintain a list of moved files and their new locations in the content repo (see DFE-Digital/get-into-teaching-content/pull/127) which, as part of the docker build will replace the dummy one in this repo (in `config/redirects.yml`).

The dummy one is present for testing, it means we don't need to inject the routes we want to test while the app is running.

### Guidance to review

Is this a sensible approach?

### DIFF
--- a/config/redirects.yml
+++ b/config/redirects.yml
@@ -1,0 +1,7 @@
+# this file will be replaced with config/redirects.yml from the content repo
+# during the docker build. The values below are used by the redirection tests
+redirects:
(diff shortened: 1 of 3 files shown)


jebw (approved):
LGTM

PR #529, opened 2020-11-02
Added a save button to the Cookie Preferences page

jebw:
### Trello card

https://trello.com/c/oyXVIACa

### Context

Currently the user does not realise their changes are being saved as they update their cookie preferences

1. Added a save button
2. Only save when save button is clicked
3. Show a message to the user notifying them that their preferences are being saved
4. Hide the message if the user subsequently updates their preferences

### Guidance to review

Probably worth downloading the PR locally and having a play with it

### DIFF
--- a/app/views/cookie_preferences/show.html.erb
+++ b/app/views/cookie_preferences/show.html.erb
@@ -176,6 +176,9 @@
             <span data-target="cookie-preferences.flash" class="save-with-confirmation__message">
               Your cookie preferences have been saved
             </span>
+
+            <br />
+            <%= link_to "Back to page", :back, class: "secondary-button" %>
           </p>
 
           <p>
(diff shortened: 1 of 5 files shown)


peteryates (commented):
This looks good to me, everything works as expected. My only feedback is really that the path onwards from clicking 'Save' isn't obvious. I need to scroll back up and click 'Back'

Should the save button be called 'Save and return' and do both? Perhaps there should be a link/button alongside it that takes the user back to the homepage?

jebw:
Yeh - good call - I'll talk to Ben in the morning.

I think it will need to be a separate link, this is all client side so there's no 'flash' - hence navigating away removes the ability to give feedback.

peteryates (approved):
Nice!

PR #532, opened 2020-11-03
Try to get SonarCloud working again

peteryates:
This PR fixes the sending of simplecov and rubocop's output to Sonarcloud. Simplecov changed their format at version 0.18, so it's been downgraded to 0.17.1 for the time being.


peteryates:
Screenshot from 2020-11-03 15-52-54

ethax-ross (approved):
Looks good to me. I wonder if we can get the JS coverage reporting in at some point?

PR #534, opened 2020-11-03
Replaced usage of deprecated bundler option

jebw:
### JIRA ticket number

https://trello.com/c/O9T3Qco5

### Context

We are using the `--without` bundler option but this has now been deprecated

1. Switch to using the env var replacement
2. Removed the unused sassc compillation option


peteryates (approved):
Makes sense.

PR #536, opened 2020-11-04
Blur semitransparent backgrounds

peteryates:
### JIRA ticket number

N/A - discussed in the design session

### Context

Some text over busy background images is tricky to read

Blur the background slightly to prevent 'hard edges' from the image showing through. This is applied via a `rotated-block` utility function and can be adjusted in one place and overridden where necessary.

Also remove some duplication from the colours file.

### Guidance to review

Does it look good?

| Before | After |
| ------ | ------- |
| Screenshot from 2020-11-04 09-48-01 | Screenshot from 2020-11-04 09-48-17 |


ethax-ross (approved):
I missed the discussion but assume this was approved design/content-wise. Have you tested how this behaves in older browsers given support for `backdrop-filter` is flakey?

peteryates:
In older browsers it looks exactly as it does now, this just enhances for those that support it.

PR #537, opened 2020-11-04
Revert "Revert "Added a save button to the Cookie Preferences page""

jebw:
Reverts DFE-Digital/get-into-teaching-app#535

And adds a fix for :back not re-running js in cdn environments

In non-cdn environments, `history.back()` will result in the earlier page being
reloaded.

In our production environment, `history.back()` restores the earlier page but
does not reload the content, or re-run the JS which means the cookie acceptance
overlay continues to be shown, even though the user has accepted cookies by this
point.


ethax-ross (approved):
Looks good, and TIL!

PR #542, opened 2020-11-06
No results update

ethax-ross:
### Trello card

### Context

Currently, if there are no results we display a single 'no results' message to the user. We want to change this behaviour such that;

- If a user searches across all types and there are no results for _some_ of the types, we still want to display the type sections but have a 'no results' message in each section that has no events.
- If a user searches across all types and there are no results in _any_ of the types, the existing behaviour remains (single 'no results' message).
- If the user searches by a specific type, the existing behaviour remains (single 'no results' message).
- If there are no results on the landing page for a specific type, we want to hide the section completely (retaining the existing behaviour).

We also want to remove the Application Workshop events from the application, as there are none in the CRM at the moment and aren't likely to be any in the near future.

- Refactor Event::Search to class methods
- Display no results message per category
- Remove Application Workshop events from the app

### Guidance to review

Screenshot 2020-11-06 at 10 40 21

### DIFF
--- a/app/components/events/search_component.html.erb
+++ b/app/components/events/search_component.html.erb
@@ -11,13 +11,13 @@
       <% if include_type %>
         <%= tag.div(class: input_field_classes(:type)) do %>
             <%= f.label :type, "Event type", class: "search-for-events__content__input__label" %>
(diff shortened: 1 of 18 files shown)


ethax-ross:
Closing as I've split this out into 3 individual PRs.

PR #543, opened 2020-11-06
Redirect to first step on resend_verification bad request

ethax-ross:
If a candidate's session expires (so they sit on the authenticate step overnight, for instance) then they hit the link to resend their verification code the app attempts to make an invalid request to the API (email, first and last name are nil). This currently results in a 500 error for the user. Instead, we rescue this condition and redirect to the first step so that they can re-authenticate.

Fixes https://sentry.io/organizations/dfe-bat/issues/2010050592/?project=5276960&referrer=slack


peteryates (approved):
Again, looks good :+1:

PR #544, opened 2020-11-06
Refactor Event::Search to class methods

ethax-ross:
### Trello card

### Context

A number of the methods on `Event::Search` can be class methods; this will come in handy when we need to make assertions in some of the specs later on (in particular referencing the `available_event_types`).

- Refactor Event::Search to class methods

Also contains minor improvement for test coverage in `Events::Search`.

### DIFF
--- a/app/models/events/search.rb
+++ b/app/models/events/search.rb
@@ -8,6 +8,8 @@ module Events
     DISTANCES = [30, 50, 100].freeze
     MONTH_FORMAT = %r{\A20[234]\d-(0[1-9]|1[012])\z}.freeze
 
+    delegate :available_event_type_ids, :available_distance_keys, to: :class
+
     attribute :type, :integer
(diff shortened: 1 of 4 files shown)


app/models/events/search.rb
```diff
@@ -21,32 +23,34 @@ class Search
...
-      available_distances.map(&:last)
-    end
+      def available_distances
+        [["Nationwide", nil]] + DISTANCES.map { |d| ["Within #{d} miles", d] }
+      end
 
-    def available_months
-      (0..5).map do |i|
-        month = i.months.from_now.to_date
+      def available_months
```
peteryates, line 45: I guess this could be shortened a bit to something like `(0..5).map { |m| m.months.from_now.strftime("%Y-%m") }`, but no strong feelings

peteryates (approved):

PR #545, opened 2020-11-06
Remove Application Workshop events from the app

ethax-ross:
### Trello card

### Context

As there are no Application Workshop events in the CRM at the moment, we want to hide it as an option in the search type drop down and from any search results.

We are also removing the Application Workshops category page and associated content.

- Remove Application Workshop events from the app


peteryates (approved):
:+1:

PR #546, opened 2020-11-06
Convert remaining content to markdown

peteryates:
This is the companion PR for DFE-Digital/get-into-teaching-content/pull/134. The changes here allow more rendering responsibility to be taken on by the app by expanding the frontmatter.

* [x] Tests
* [x] Add styling for external links
* [x] Add `target="_blank"` behaviour to the external links. This will have to happen via JS I think. More investigation needed. This should also be taken as an opportunity to add the "link opens in new window" assistive text

### DIFF
--- a/app/views/layouts/steps_to_become_a_teacher.html.erb
+++ b/app/views/layouts/steps_to_become_a_teacher.html.erb
@@ -18,19 +18,6 @@
                 :mailinglist => @front_matter["mailinglist"]
             %>
             <section class="content content--<%= @front_matter["direction"] %> <%= @front_matter["fullwidth"] ? "" : "container-1000" %>">
-                <% if @front_matter["on_this_page"].present? %>
-                  <div class="content__right">
-                    <div class="link-block link-block--jump">
-                      <h2 class="link-block__header">On this page:</h2>
-                      <ul class="link-block__list">
-                        <% @front_matter["on_this_page"].each do |item, anchor| %>
-                          <%= link_to(item, anchor) %>
-                        <% end %>
-                      </ul>
-                    </div>
-                  </div>
-                <% end %>
-
                 <div class="content__left">
                   <% if @front_matter["alert"].present? %>
                     <%= tag.p(@front_matter["alert"], class: "content-alert") %>
--- a/app/views/layouts/steps_to_become_a_teacher.html.erb
+++ b/app/views/layouts/steps_to_become_a_teacher.html.erb
@@ -37,13 +37,15 @@
                   <% end %>
 
                   <div class="accordions" data-controller="accordion">
-                    <% @front_matter["steps"].each.with_index(1) do |(title, partial), i| %>
-                      <%= tag.button(class: "steps-header", id: "step-#{i}", data: { action: "click->accordion#toggle", target: "accordion.header" }, aria: { controls: "collapsable-content-#{i}", expanded: (i == 1) }) do %>
-                        <%= tag.h2(%(#{i}. #{title})) %>
-                      <% end %>
+                    <% if @front_matter["steps"].present? %>
+                      <% @front_matter["steps"].each.with_index(1) do |(title, partial), i| %>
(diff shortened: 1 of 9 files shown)


app/views/layouts/steps_to_become_a_teacher.html.erb
```diff
@@ -0,0 +1,64 @@
...
+            <%= render "sections/hero",
+                :header => @front_matter["title"],
+                :image => @front_matter["image"],
+                :mobileimage => @front_matter["mobileimage"],
+                :deep => @front_matter["deepheader"],
+                :subtitle =>  @front_matter["subtitle"],
+                :mailinglist => @front_matter["mailinglist"]
+            %>
+            <section class="content content--<%= @front_matter["direction"] %> <%= @front_matter["fullwidth"] ? "" : "container-1000" %>">
+                <% if @front_matter["on_this_page"].present? %>
```
ethax-ross, line 21: Not sure what `on_this_page` relates to? I can't find it in the content repo at all?

app/webpacker/styles/content.scss
```diff
@@ -97,9 +97,18 @@
                 color: $black;
                 text-decoration: none;
             }
-        }
-        
 
+            // any content hyperlink that contains two slashes is
+            // deemed 'external' and will get a icon
+            &[href*="//"] {
```
ethax-ross, line 103: That's a neat way of targeting them!

spec/features/content_pages_spec.rb
```diff
@@ -11,7 +11,7 @@
 
   class << self
     def files
-      Dir["app/views/content/**/*.md"]
+      Dir["app/views/content/**/[^_]*.md"]
```
ethax-ross, line 14: What's the rationale here? If we are migrating to markdown there'll be no HTML pages left in the content repo and therefore nothing to test?

ethax-ross (commented):
Couple of questions/bits I'm not following here

spec/features/content_pages_spec.rb
peteryates: Ah, I introduced [markdown partials](https://github.com/DFE-Digital/get-into-teaching-content/tree/convert-remaining-content-to-markdown/content/steps-to-become-a-teacher/accordion) for the accordion content and they have no front matter so the tests were failing.

ethax-ross: Oh I see, I missed that it was checking for the underscore/partial. Makes sense now 👍

app/views/layouts/steps_to_become_a_teacher.html.erb
peteryates: It's in the linked branch only, like on [this page](https://github.com/DFE-Digital/get-into-teaching-content/blob/convert-remaining-content-to-markdown/content/returning-to-teaching.md). It's used to generate these boxes for the right column

Screenshot from 2020-11-09 14-14-52

ethax-ross: Ah yeah I see them now, I wonder if calling it something else would be clearer, maybe `jump_links`? also, it may not be worth the effort just now but if we're doing this on a lot of pages I wonder if we can infer them from the headings and just have a flag in the markdown like `render_jump_links: true` or similar?

peteryates: They're currently on 3 pages. It's definitely not the best or most-scalable approach. Kramdown does support tables of contents, but rendering the `toc` outside of the body of the document isn't straightforward.

As there are only three at the moment let's go with the manual approach and it can be something to revisit if we add more.

ethax-ross (approved):

app/views/layouts/steps_to_become_a_teacher.html.erb
peteryates: I did rename the entry to `jump_links` though, it's definitely less cryptic.

ethax-ross (dismissed):

ethax-ross (approved):

PR #547, opened 2020-11-06
Display no results message per category

ethax-ross:
### Trello card

### Context

Currently, if there are no results we display a single 'no results' message to the user. We want to change this behaviour such that;

- If a user searches across all types and there are no results for some of the types, we still want to display the type sections but have a 'no results' message in each section that has no events.
- If a user searches across all types and there are no results in any of the types, the existing behaviour remains (single 'no results' message).
- If the user searches by a specific type, the existing behaviour remains (single 'no results' message).
- If there are no results on the landing page for a specific type, we want to hide the section completely (retaining the existing behaviour).

- Display no results message per category

### Guidance to review

Screenshot 2020-11-06 at 15 48 15

### DIFF
--- a/app/views/events/index.html.erb
+++ b/app/views/events/index.html.erb
@@ -30,18 +30,26 @@
     </section>
 
     <% if @events_by_type.any? %>
-      <% if @group_presenter.get_into_teaching_events.any? %>
-        <%= render partial: "event_category", locals: { organised_by: t("event_groups.get_into_teaching"), category_groups: @group_presenter.get_into_teaching_events } %>
+      <% if @group_presenter.get_into_teaching_events.values.flatten.any? || @display_empty_types %>
+        <%= render partial: "event_category", locals: { 
+          organised_by: t("event_groups.get_into_teaching"), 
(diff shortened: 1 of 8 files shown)


app/views/events/index.html.erb
```diff
@@ -30,18 +30,26 @@
     </section>
 
     <% if @events_by_type.any? %>
-      <% if @group_presenter.get_into_teaching_events.any? %>
-        <%= render partial: "event_category", locals: { organised_by: t("event_groups.get_into_teaching"), category_groups: @group_presenter.get_into_teaching_events } %>
+      <% if @group_presenter.get_into_teaching_events.values.flatten.any? || @display_empty_types %>
```
peteryates, line 33: Perhaps this could be a method on the presenter?

ethax-ross: Good suggestion 👍 I'll do that now

peteryates (approved):
Looks great.

peteryates (approved):
LGTM.

PR #553, opened 2020-11-09
Feature/content cta styles

sigmaben:
Add styles for new Content CTA


app/webpacker/styles/components/content-cta.scss
```diff
@@ -0,0 +1,12 @@
+.content__left {
```
ethax-ross, line 1: Does this need to be in the `content__left` block? I think we're trying to move more towards a traditional BEM naming convention where possible

ethax-ross (approved):

PR #554, opened 2020-11-09
Extract window.focus to method and mock

ethax-ross:
We aren't able to mock the iFrame `contentWindow.focus` from Jest; instead, the call gets extracted to a method which can then be mocked from the specs.

Quietens the 'method not implemented' error Jest was throwing in the test suite.


jebw (approved):
Nice!

PR #558, opened 2020-11-11
Align postcodes validation with api

jebw:
### Trello card

https://trello.com/c/32JmO988

### Context

The API implements different validation rules to the App. The app also allows through postcodes which can be auto-corrected (eg `TE57 ING` -> `TE57 1NG`) but doesn't actually auto-correct it - which means the API still receives the un-autocorrected version.

1. Replace `UKPostcode` gem with Regex from the API
2. Added validation for the error messages
3. Upcase postcodes as part of the normalisation

### DIFF
--- a/spec/models/events/search_spec.rb
+++ b/spec/models/events/search_spec.rb
@@ -63,6 +63,8 @@ describe Events::Search do
       end
 
       context "with assigned distance" do
+        let(:msg) { "Enter a valid postcode" }
+
         subject do
           described_class.new distance: described_class::DISTANCES.first
         end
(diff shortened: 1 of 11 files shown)


spec/models/events/search_spec.rb
```diff
@@ -113,11 +116,11 @@
       end
 
       context "when there's whitespace around a provided postcode" do
-        subject { build(:events_search, postcode: " TE57 1NG  ") }
+        subject { build(:events_search, postcode: " te57 1ng  ").tap(&:validate) }
```
ethax-ross, line 119: It looks like this is an invalid postcode so it would never query the API, does it not need to be a valid postcode here?

jebw: That is a validate postcode?

`\A([A-Z][A-HJ-Y]?\d[A-Z\d]? ?\d[A-Z]{2}|GIR ?0A{2})\Z`

jebw: [A-Z] = T

ethax-ross: Ah sorry I misread this line - I thought this was testing that the postcode isn't valid, but it's an `I` not a `1` 🙄 

https://github.com/DFE-Digital/get-into-teaching-app/pull/558/files#diff-89ee7cb8cc45135f0e45924fb716574b01e79ee5ef9118992fe6f5b2d613848bR76

ethax-ross (approved):

PR #560, opened 2020-11-11
Fix for multiple acronyms in a single node

jebw:
The replacement logic uses a gsub loop to find the replacements for each
acronym. When there was not a replacement it exited the loop early, but the
output from the loop was being used to update the string.

The early exit from the loop was returning nil, so replacing the unknown acronym
with a nil instead of leaving it as is.


peteryates (approved):
Looks good. My approach was to make `acronym_for` return the supplied acronym if nothing matched. Both work 👍🏽

PR #569, opened 2020-11-16
Rescue ApiError 404 in ApplicationController

ethax-ross:
If the API ever returns a 404 error we would most likely want to display the 404 page to the user.

Rescue the `ApiError` 404 at the `ApplicationController` level.

The dynamic render highlighted by Brakeman is a false positive; the event ID is whitelisted by the API and I think the use of ViewComponents is confusing it.


jebw (approved):
LGTM

PR #571, opened 2020-11-16
Use parallel_tests gem in CI

ethax-ross:


ethax-ross:
Closing, has issues I'll pick it up again at some point.

PR #572, opened 2020-11-16
Added an exception for the false positive from Brakeman

jebw:
Brakeman is treating the rendering of a ViewComponent as a Dynamic Render Path.

As far as I can tell this is a false positive.

This appears to be a combination of both not recognising the API call and not
being aware of ViewComponents.

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,0 +1,37 @@
+{
+  "ignored_warnings": [
+    {
+      "warning_type": "Dynamic Render Path",
+      "warning_code": 15,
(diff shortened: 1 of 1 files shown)


peteryates (approved):
Looks good. Could we avoid this by just doing a `params.permit` in the controller?

jebw:
I'll give that a go, but I think the issue is that it see's render as being passed something user supplied and assumes its a template because it doesn't know about ViewComponents - but also doesn't recognise that the user supplied data has been filtered, because it goes through an API components - comments I've found imply it has an exception the the param is handed to an ActiveModel

jebw:
Just tried the permit option, still the same issue

PR #574, opened 2020-11-17
Remove telephone field from Mailing List wizard

ethax-ross:
> :warning: **Waiting to hear back from Shivani to ensure the step rename won't cause issues with analytics**

### Trello card

[Trello-557](https://trello.com/c/4MqJYbvB/557-development-remove-optional-phone-number-from-mailing-list-flow)

### Context

Based on the data we've seen so far, it appears that users at a higher commitment will provide their phone number more often than lower commitment people.

There is currently no value behind this offering, so we're going to remove it from the flow, and instead, ask users to self serve once sign up is complete (another ticket to follow on the content for the confirmation page).

Work required:

1. remove the optional phone number box
2. work with content colleagues to change title of the page
3. work with content colleagues to draft new help text (if required)

- Remove telephone field from Contact step

- Rename MailingList::Steps::Contact to PrivacyPolicy

As we have removed the telephone field from the step it no longer makes sense to call it `Contact`; instead, it only asks the user to accept the privacy policy, so renaming it to `PrivacyPolicy`. 

- Update PrivacyPolicy step copy

This step was recently changed to remove the telephone field; this updates the copy to better reflect the state of the step now.

### Guidance to review

Screenshot 2020-11-17 at 13 47 59

### DIFF
--- a/spec/factories/mailing_list/contact_factory.rb
+++ b/spec/factories/mailing_list/contact_factory.rb
@@ -1,5 +1,0 @@
-FactoryBot.define do
-  factory :mailing_list_contact, class: MailingList::Steps::Contact do
-    accept_privacy_policy { "1" }
-  end
-end
--- a/spec/factories/mailing_list/contact_factory.rb
+++ b/spec/factories/mailing_list/contact_factory.rb
@@ -1,6 +1,5 @@
 FactoryBot.define do
   factory :mailing_list_contact, class: MailingList::Steps::Contact do
-    telephone { "01234567890" }
     accept_privacy_policy { "1" }
   end
 end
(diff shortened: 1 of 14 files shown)


spec/factories/mailing_list/contact_factory.rb
```diff
@@ -1,6 +1,5 @@
 FactoryBot.define do
-  factory :mailing_list_contact, class: MailingList::Steps::Contact do
```
jebw, line 2: This should have the factory file renamed as well

jebw (dismissed):
LGTM

spec/factories/mailing_list/contact_factory.rb
ethax-ross: Ah good spot

jebw (approved):

PR #575, opened 2020-11-17
Use wizard gem

jebw:
### Trello card

https://trello.com/c/dzHEOYnI

### Context

We should be sharing the the wizard code, at least between the apps and making it available for other teams.

1. Wizard code split out into `DFE-Digital/dfe_wizard` - this still needs some tests and a Readme.md but allows for code sharing for now
2. Wizard classes change to depend upon versions from gem
3. Old wizard classes removed

### Guidance to review

This will clash with Ross' unmerged PR so I'll rebase once thats on `master`


peteryates:
Closing this for now, we can re-open if/when we revisit.

PR #577, opened 2020-11-18
Move degree status field to its own step

ethax-ross:
### Trello card

[Trello-562](https://trello.com/c/kVyEAwaw/562-remove-what-stage-are-you-at-with-your-degree-from-page-1-of-mailing-list-flow-reword-question-and-use-radio-buttons)

### Context

The question "What stage are you at with your degree?" is presumptive and might be contributing towards drop outs. We want to remove the question from page 1 and then rephrase the question/answer and add to a new page 2.

We also want to change the select box to a radio input and reword the answers.

- Move degree status field to its own step

### Guidance to review

Screenshot 2020-11-23 at 10 31 40

### DIFF
--- a/app/models/mailing_list/steps/degree_status.rb
+++ b/app/models/mailing_list/steps/degree_status.rb
@@ -1,0 +1,25 @@
+module MailingList
+  module Steps
+    class DegreeStatus < ::Wizard::Step
+      attribute :degree_status_id, :integer
+
+      validates :degree_status_id,
+                presence: true,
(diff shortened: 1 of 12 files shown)


peteryates (approved):
Looks good.

My only concern is that if new degree statuses are added by the CRM and we're not aware will we get a `"translation missing: something"` in the select?

I don't know how frequently they're likely to change though.

peteryates (approved):

PR #587, opened 2020-11-23
Work around an IE bug on word-wrapping links

peteryates:
[IE 11 suffers from a bug](https://www.webdevelopmentreference.com/ie-11-word-wrap-word-break/) where long links aren't properly-wrapped using `word-wrap`. Rather than go down the rabbithole of hacks to rectify this we're now just always displaying the list vertically for IE users without affecting the main audience.

| Before | After |
| ------ | ------ |
| Screenshot from 2020-11-23 12-22-31 | Screenshot from 2020-11-23 12-21-46 |

### DIFF
--- a/app/webpacker/styles/home.scss
+++ b/app/webpacker/styles/home.scss
@@ -192,6 +192,16 @@
         @media (max-width: $steps-row-limit) {
             flex-direction: column;
         }
+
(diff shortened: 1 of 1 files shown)


jebw (approved):
LGTM

PR #592, opened 2020-11-24
Abstracted card component

jebw:
### Trello card

https://trello.com/c/xmUzKlT8

### Context

The Trello story requires that we show different types of cards. Depending upon settings. The should all render the same way but with different headers, links, etc.

Plan is to use template inheritance to use a common 'card' template, but different ViewComponent classes inheriting from it.

1. Moved as much of `Stories::CardComponent` into a generalised `CardComponent` as possible
2. Added a `Stories::CardComponent` to inherit from it


peteryates (dismissed):
Looks good

peteryates (approved):
Looks good

PR #597, opened 2020-11-25
Navbar current page highlight

jkempster34:
- Add highlight to current page title in navbar
- Update navbar underline to match design

### DIFF
--- a/spec/components/navbar/navbar_component_spec.rb
+++ b/spec/components/navbar/navbar_component_spec.rb
@@ -1,0 +1,38 @@
+require "rails_helper"
+
+describe Navbar::NavbarComponent, type: "component" do
+  let(:all_frontmatter) do
+    { "/mock-path" => { title: "Mock title", navigation: 20 } }
+  end
+
+  before do
+    allow(Pages::Frontmatter).to receive(:list).and_return all_frontmatter
+  end
+
+  subject! { render_inline(Navbar::NavbarComponent.new(all_frontmatter)) }
+
+  describe "Desktop navbar component" do
+    context "when rendered" do
+      it "shows background colour on link that matches the current page" do
+        desktop_navbar = page.find(".navbar__desktop")
+        home_link = desktop_navbar.find("a", text: "Home")
+        parent_li = home_link.find(:xpath, "..")
+
+        expect(parent_li["class"]).to match("active")
+      end
+
+      it "shows no background colour on links that do not match the current page" do
+        desktop_navbar = page.find(".navbar__desktop")
+        home_links = desktop_navbar.all("a").reject do |anchor|
+          anchor.native.children.text == "Home"
+        end
+        parent_lis = []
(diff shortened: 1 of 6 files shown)


spec/components/navbar/navbar_component_spec.rb
```diff
@@ -0,0 +1,38 @@
+require "rails_helper"
+
+describe Navbar::NavbarComponent, type: "component" do
+  let(:sitemap) do {
+    markdown_content:
+    [
```
jebw, line 6: This will require #598 to be merged (once the fixes to the release have gone out)

But after that - This can be 

```ruby
let(:all_frontmatter) do
  {
    "/accessibility", { title: "Accessibility", navigation: 20 },
    "/accessibility2", { title: "Second accessibility page", navigation: 30 },
  }
end

before do
  allow(Pages::Frontmatter).to receive(:list).and_return all_frontmatter
end
```

One thing to note is the index of markdown pages can only have one hash of frontmatter per path, because in reality there can only be one markdown file per path

spec/components/navbar/navbar_component_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      it "shows no background colour on links that do not match the current page" do
+        desktop_navbar = page.find(".navbar__desktop")
+        home_links = desktop_navbar.all("a").reject do |anchor|
+          anchor.native.children.text == "Home"
+        end
+        parent_lis = []
+        home_links.each { |anchor| parent_lis.push(anchor.find(:xpath, "..")) }
+
+        parent_lis.each do |li|
+          expect(li["class"]).to_not match("active")
```
peteryates, line 33: I looks like it's working fine but I think you can make it a bit more concise by asserting that the one 'active' link is the only active one present:

```ruby
page.all(".navbar__desktop li.active", count: 1)
```

spec/components/navbar/navbar_component_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  subject! { render_inline(Navbar::NavbarComponent.new(all_frontmatter)) }
+
+  describe "Desktop navbar component" do
+    context "when rendered" do
+      it "shows background colour on link that matches the current page" do
+        desktop_navbar = page.find(".navbar__desktop")
+        home_link = desktop_navbar.find("a", text: "Home")
+        parent_li = home_link.find(:xpath, "..")
+
+        expect(parent_li["class"]).to match("active")
```
peteryates, line 21: This is fine but we can make use of a block on the expectation to test what's within, rather than starting on the inside and working outwards:

```ruby
expect(page).to have_css("li.active") do |active_li|
  expect(active_li).to have_link("Home", href: "/")
end
```

peteryates (changes requested):
Good stuff, everything looks fine to me. I left a few pointers on the specs, I think they can be made a touch less verbose.

peteryates (approved):
Looks good to me.

PR #598, opened 2020-11-25
Added Page objects

jebw:
### Trello card

https://trello.com/c/xmUzKlT8

### Context

Currently we can't access either data or frontmatter from the controller, we also don't have any way of accessing page related data from either the templates or view components.

This is primarily a refactor aimed at addressing the above issue

1. Add a `Pages::Frontmatter` class to parse out all the front matter - this does dynamic loading (to avoid needing server restarts) in `development` and `test` environments and caches the frontmatter in memory in `production`-like environments.

2. Add a `Pages::Page` class to represent a page - at the moment this is minimal and just provides access to frontmatter but in a follow on PR it will allows loading data into pages

3. Change the ERB template handler to load the sitemap using `Pages::Frontmatter` - this could probably be dropped at somepoint and use the @page variable.

### Guidance to review

1. Visual checking - both Markdown and ERB pages should load identically
2. Navigation menus should show the same

*Note:* this does add a semantic change - keys on frontmatter are now symbolized _but_ nested keys are not. Its done to optimize memory use (symbols are reused). Whether nested keys/no keys should be symbolized is open for debate, I'm looking for feedback on that - easy to change if they should be?

### DIFF
--- a/app/models/pages/frontmatter.rb
+++ b/app/models/pages/frontmatter.rb
@@ -4,8 +4,12 @@ module Pages
     attr_reader :content_dir
 
     class << self
-      def find(template, content_dir)
+      def find(template, content_dir = nil)
         instance(content_dir).find template
+      end
+
+      def list(content_dir = nil)
+        instance(content_dir).list
       end
 
       def instance(content_dir)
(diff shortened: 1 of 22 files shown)


app/models/pages/frontmatter.rb
```diff
@@ -0,0 +1,106 @@
...
+      self
+    end
+
+    def preloaded?
+      !@frontmatter.nil?
+    end
+
+    class NotMarkdownTemplate < RuntimeError
+      def initialize(tmpl)
+        super "Cannot find Markdown Page #{tmpl}.md"
```
peteryates, line 56: Minor inconsistency in variable naming here

peteryates (commented):
Looks great, miles neater, just one minor inconsistency I commented on

peteryates (dismissed):
Looks good to me

peteryates (approved):
Nice

PR #600, opened 2020-11-26
Accordion calls to action

peteryates:
### Trello card

[536](https://trello.com/c/aF449leD/536-development-code-up-new-popper-bens-word-for-steps-inc-new-international-and-ways-to-train-pages-for-sign-posting)

### Context

To support the upcoming content changes and wider use of accordions this, we need some 'richer' content than text. These calls to action (aka 'poppers') will sit under the rendered markdown and contain either a link to another page or in the case of the first component, a chat widget.

* Make the accordion background white
* [Change the format of steps in frontmatter so they can have a `call_to_action` key](https://github.com/DFE-Digital/get-into-teaching-content/pull/155)
* Create an accordion component that accepts slots for each step
* Add a `ChatOnlineComponent` that's _registered_ with the accordion, whenever `chat_online` is passed in the `ChatOnlineComponent` will be rendered
* Make sure it errors if an unrecognised cta is supplied and that the error message is obvious

### Guidance to review

This approach was taken so we have flexibility with the front matter arguments, it will be easy to replace the `chat_online` argument with an object if we need more functionality in the subcomponents.

Screenshot from 2020-11-26 09-40-02


app/components/content/accordion_component.rb
```diff
@@ -0,0 +1,35 @@
+module Content
+  class AccordionComponent < ViewComponent::Base
+    include ViewComponent::Slotable
```
ethax-ross, line 3: Nice, I've never seen this used before. Looks like its experimental still so one to keep an eye on when we upgrade the gem.

spec/support/views/content/_dummy.html.erb
```diff
@@ -0,0 +1 @@
+<p>Dummy content</p>
```
ethax-ross, line 1: I can't see this file referenced anywhere?

peteryates: Ah yeah, good spot. Remnant of a failed attempt 😅

ethax-ross (approved):

PR #602, opened 2020-11-26
Add LID tracking pixel

ethax-ross:
> :warning: **Associated content changes to be merged at the same time: https://github.com/DFE-Digital/get-into-teaching-content/pull/158**

### Trello card

[Trello-314](https://trello.com/c/8PmnT4DD/314-development-add-lid-pixels-to-pages-for-audio-advertising)

### Context

We want to include an audio ad tracking pixel (which is actually 2x image type pixels) to various dynamic and static web pages.

- Add LID tracking pixel controller

Add a StimulusJS controller for loading the LID tracking pixels. We appear to load the same pixel on multiple domains according to the spreadsheet:

```
https://pixelg.adswizz.com&?client=GovernmentDFE&action=cs&j=0&event=eventName
https://tracking.audio.thisisdax.com&?client=GovernmentDFE&action=cs&j=0&event=eventName
```

The event name is passed as state to the controller, for example:

```
<div data-controller="lid" data-lid-action="track" data-lid-event="eventName"></div>
```

Introduces `ImagePixelBaseController` to DRY up the implementation when using image tracking pixels.

- Enable LID tracking pixels in production

- Allow LID pixel in CSP

- Fire LID tracking pixel on dynamic pages

Loads the LID tracking pixel for dynamic pages. The static pages need another solution as we can't directly access the content/they don't have unique layouts.

- Render analytics partial for LID pixels in content

We want to be able to display LID pixels on certain content pages. To achieve this we render a `components/analytics` partial that checks the frontmatter for `lid_pixel_event` - if present, it will render the HTML that hooks in the `LidController` Stimulus controller.


app/views/layouts/steps_to_become_a_teacher.html.erb
```diff
@@ -42,6 +48,7 @@
         <%= render "components/videoplayer" %>
         <%= render "sections/cookie-acceptance" %>
         <%= render "sections/feedback-bar" %>
+        <%= render "components/analytics" %>
```
peteryates, line 51: Would rendering this at the top allow us to avoid passing the global frontmatter around?

peteryates (commented):
Looks good, I think just rendering the tracking pixels at the top is preferable over holding on to the global frontmatter and reassigning though

app/views/layouts/steps_to_become_a_teacher.html.erb
ethax-ross: I almost took that route, but there are other things that depend on the front matter further down, such as:

https://github.com/DFE-Digital/get-into-teaching-app/blob/73b6992f8726590012646ade57a0abb8c65bfde8/app/views/layouts/steps_to_become_a_teacher.html.erb#L43

peteryates: Ok. I think we're almost in a position to remove `hide_page_helpful_question` and render it based on the template, perhaps we can adjust this at a later date 👍🏽

peteryates (approved):

PR #605, opened 2020-11-27
Attempt to simplify sitemap access

peteryates:
Currently the sitemap is only used for the navbar which is present in the header and footer, in prod it'll only change when a new version of the site is deployed so building at boot feels like it makes sense. This is an attempt to simplify its population while still leaving it accessible from the views and controllers (where we'll need it in case we do end up building [an actual sitemap](https://en.wikipedia.org/wiki/Sitemaps#File_format).


peteryates:
This was made unnecessary by #620

PR #606, opened 2020-11-27
Tweak EventBox icon sizes

ethax-ross:
The `EventBox` icon sizes weren't adhering to what was in the CSS due to the display type being `table-cell`. Instead, the icon is wrapped in a `div` with the display type and the icons then take their correct dimensions.

Tweaks the icon sizes so that they look more uniform (the sizes are different but it looks more uniform to the naked eye).

|   Before   |       After       | 
|----------|-------------|
|Screenshot 2020-11-27 at 13 46 56|Screenshot 2020-11-27 at 13 46 53|


peteryates (approved):
Looks good 👍🏽

PR #616, opened 2020-11-30
Add ways to train page

peteryates:
### Trello card

https://trello.com/c/lS52QiJi/533-development-ways-to-train-accordion-style-content-page

### Context

The ways to train content is currently on a guidance page, which jars users when they're taken away from GiT.

Create an accordion style page that contains the main points from the ways to train guidance page. It uses exactly the same accordion as the steps to become a teacher page but some additions have been made to the accordion.

The accordion component has been extended to allow calls to action to be specified in the frontmatter and they will be rendered after the content in the accordion segment. Additionally, content that's rendered before or after the accordion can be specified using the `content_before_accordion` and `content_after_accordion` keys. They accept calls to action or partials, which will be rendered automatically. In the screenshot below, the content above the accordion is rendered from a partial and the 'Next Steps' CTA is rendered by passing in `cta: next_steps` in the front matter.

ways-to-train

#### Mobile

Screenshot from 2020-12-01 12-46-36

Screenshot from 2020-12-01 12-51-26

### Guidance to review

The 'next steps' call to action has stuff we'd usually consider as 'content' but it's in the app repo. It's too complicated to sensibly add to frontmatter as it contains lists and lists of actions. I think expecting editors to add/remove these without a live preview is unrealistic so I've left it in the app repo on purpose. Embedding markdown in frontmatter seemed like a really bad idea.

### Final tweaks

@croftla and I have had a little tweak/adjustment of the final CTA - mainly so the buttons have an 'action' associated with them. Here's the WIP

Screenshot from 2020-12-02 09-55-22


CroftLA:
Make 2 columns with buttons underneath - need to edit the last bit of text, shorten to training provider and funding options.

config/routes.rb
```diff
@@ -59,6 +59,7 @@
 
   get "/guidance/*page", to: "guidance#show"
   get "/steps-to-become-a-teacher", to: "steps_to_become_a_teacher#show"
+  get "/ways-to-train", to: "ways_to_train#show"
```
ethax-ross, line 62: Worth having a request spec to hit up this route?

app/views/layouts/accordion.html.erb
```diff
@@ -17,14 +16,31 @@
                     <%= tag.div(tag.p(@front_matter["alert"]), class: "content-alert content-alert--fullwidth") %>
                   <% end %>
 
-                  <%= render Content::AccordionComponent.new do |accordion| %>
-                    <% @front_matter["steps"]&.each do |title, contents| %>
-                      <%= accordion.slot(:step, title: title, call_to_action: contents["cta"]) do %>
-                        <%= render(partial: contents["partial"]) %>
+                  <%# FIXME: we're tapping the frontmatter here because
```
ethax-ross, line 19: We've got two different work arounds we're using for this in the source code now, I don't think we should block this going out because of it but having FIXME's going in doesn't sit well -- shall we have a think about a longer-term solution we can live with to follow up this PR?

ethax-ross (commented):
Looks good, there are a few gaps in test coverage that SonarCloud highlighted I think it would be worth addressing though?

One for later, but we may want to look at creating some previews for view components, both to aid in reviewing and also it would just be handy now that our collection is growing to better keep track/document them:

https://viewcomponent.org/#previewing-components

app/views/layouts/accordion.html.erb
peteryates: I switched the `@parent_front_matter` approach to this one so it's just in one place. Agree we could do with a more robust fix though

peteryates: I changed this back to how it was before

ethax-ross (approved):

PR #618, opened 2020-11-30
Labels added to PRs

sfawcett123:
When A PR is created the labeler will attempt to assign it a label based upon what code has changed.
This is useful in the DevOps environment as we are monitoring repositories as a team, identifying work the team could do if an individual is not available


jebw (approved):
LGTM

PR #619, opened 2020-11-30
Update validator to allow outward only postcodes in event search

jkempster34:
### Trello card
https://trello.com/c/qGLRMZNe

### Context
Allow half-postcodes in the events search

### DIFF
--- a/app/models/events/search.rb
+++ b/app/models/events/search.rb
@@ -17,7 +17,7 @@ module Events
 
     validates :type, presence: false, inclusion: { in: :available_event_type_ids, allow_nil: true }
     validates :distance, inclusion: { in: :available_distance_keys }, allow_nil: true
-    validates :postcode, presence: true, postcode: { allow_blank: true }, if: :distance
+    validates :postcode, presence: true, postcode: { allow_blank: true, outward_only_postcode: true }, if: :distance
     validates :month, presence: false, format: { with: MONTH_FORMAT, allow_blank: true }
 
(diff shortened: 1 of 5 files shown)


app/models/events/search.rb
```diff
@@ -17,7 +17,7 @@ class Search
 
     validates :type, presence: false, inclusion: { in: :available_event_type_ids, allow_nil: true }
     validates :distance, inclusion: { in: :available_distance_keys }, allow_nil: true
-    validates :postcode, presence: true, postcode: { allow_blank: true }, if: :distance
+    validates :postcode, presence: true, postcode: { allow_blank: true, outward_only_postcode: true }, if: :distance
```
ethax-ross, line 20: The naming of this made me do a double take initially, as I thought it meant it would _only_ allow an outward postcode but in fact it allows an outward or full postcode. I would maybe rename it to something like `accept_partial_postcode`, but I don't feel strongly about it (I don't think that suggestion is actually all that much better, but there's maybe something clearer we can use).

ethax-ross (commented):
Looks good to me - it may be worth checking with Laura to see if we want to change the error message (or any of the other copy, such as adding hint text to the postcode field) now we accept a partial postcode as well:

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/config/locales/en.yml#L108

ethax-ross (approved):

ethax-ross (approved):
LGTM

PR #620, opened 2020-12-01
Stop setting the sitemap variable in template handler

peteryates:
### Trello card

N/A

### Context

Setting the sitemap using the template handler had some unpredictable
side effects that meant the instance variable's scope crept further than
it should've.

Now Pages::Frontmatter is in place we can use it to simplify the process
and format an output compatible with the navbar directly from
`navigation_resources`.

Also switch one of the tests to use `match_array` rather than an equality check as some specs failed locally. I think this is due to the unpredictable nature of [`glob`](https://ruby-doc.org/core-2.6.3/Dir.html#method-c-glob).

### Guidance to review

Does this make sense? Should `#navbar` be a method on `Pages::Frontmatter` instead of being in a helper? I know accessing a model from a helper isn't _ideal_, but it's not a _typical_ model. I guess that could be set to a variable using a `before_action` and we can remove the `SitemapHelper` altogether. 🤷🏽‍♂️


jebw (approved):
LGTM - loads cleaner!

Technically its a data fetch in the helper but since this is global state I think thats a fair tradeoff for the increased simplicity.

PR #621, opened 2020-12-01
Allow <div> tags in safe_html_format

ethax-ross:
The CRM appears to use `<div>` tags instead `<p>` tags; we currently strip these out so some of the text that should be on a new paragraph is squashed onto the end of other text. Allowing the `<div>` tag correctly puts the text on a new line.

|   Before   |       After       | 
|----------|-------------|
|Screenshot 2020-12-01 at 11 35 28|Screenshot 2020-12-01 at 11 35 07|

### DIFF
--- a/app/helpers/text_formatting_helper.rb
+++ b/app/helpers/text_formatting_helper.rb
@@ -7,6 +7,6 @@ module TextFormattingHelper
   end
 
   def safe_html_format(html)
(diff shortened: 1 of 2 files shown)


jebw (approved):

peteryates (commented):
Looks good.

PR #622, opened 2020-12-01
Fix Rack::Attack exception

ethax-ross:
According to the Rack::Attack docs we should be returning an array of strings here.


jebw (approved):
LGTM

PR #623, opened 2020-12-01
Override GOV.UK Transport with Sans-serif

peteryates:
This will at least keep things uniform until we get the 'official' fonts in place. Currently GDS Transport can be seen in lists, non-paragraph divs and a few other places 😞

### DIFF
--- a/app/webpacker/styles/application.scss
+++ b/app/webpacker/styles/application.scss
@@ -1,14 +1,15 @@
 @charset 'UTF-8';
 
-$govuk-font-family: Sans-serif;
-
 @import './git.scss';
+
+$govuk-font-family: $git-font-family;
+
 @import "~govuk-frontend/govuk/all";
 @import './govuk-reset.scss';
 
 @import './components/map';
 
-$fa-font-path: '~@fortawesome/fontawesome-free/webfonts'; 
+$fa-font-path: '~@fortawesome/fontawesome-free/webfonts';
 @import '~@fortawesome/fontawesome-free/scss/fontawesome';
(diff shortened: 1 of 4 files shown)


app/webpacker/styles/application.scss
```diff
@@ -1,5 +1,7 @@
 @charset 'UTF-8';
 
+$govuk-font-family: Sans-serif;
```
jebw, line 3: The rest of the site is using the 'system font stack' - I think this should be doing the same for consistency.

Since that may change in the future, I think the font family should be set in a variable before this govuk one, and the govuk one derives from the GiT one? Likewise anything else in the SCSS which assigns the system font stack should instead use the GiT font variable?

jebw (changes requested):
I think we should be using the same font family as the rest of the site? And it should all be controlled from a single variable?

app/webpacker/styles/application.scss
peteryates: Where are we defining a font stack? I couldn't find any reference to one.

```
$ ag font-family

app/webpacker/styles/layout.scss
2:    font-family: sans-serif;

app/webpacker/styles/application.scss
3: $govuk-font-family: Sans-serif;

app/webpacker/styles/text.scss
2:    font-family: sans-serif;
```

peteryates: Ok, I added `$git-font-family` variable that's defined in `text.scss` and used that to override the hard-coded `sans-serif` rules.

jebw (approved):
:+1:

PR #624, opened 2020-12-01
Explore get into teaching

jebw:
### Trello card

https://trello.com/c/OaNPI7oG

### Context

The "More Stories" section on a Story page should be replaced with a "Explore Get into teaching" section

Changing from 

Screenshot from 2020-11-18 15-19-36

to

Screenshot from 2020-11-18 15-18-38

This work has several components -;

1. Styling changes
2. Providing headers for story cards derived from the stories they link to
3. Querying Event Data in an Event Component
4. Adding an Chat online component
5. Allowing the Content designer to control the components used from Frontmatter

This PR deals with (1) and (2) from the list above

1. Refactored the basic Card template to support showing without borders, showing the blue headers and showing inline titles (this isn't needed right now but will allow this to potentially act as a base for the homepage in the future)
2. Added a mechanism for querying data for use in components
3. Changed the StoryComponent to pass (2) through
4. Added a Component for the new style Cards, which fetches the page title automatically from the Sitemap, for use in the components header.
5. Some minor test refactoring to avoid double rendering components when there are nested contexts 

### Guidance to review

This can be tested by renaming the `more_stories` frontmatter key to `explore` on a given story - eg `content/my-story-into-teaching/career-changers/swapping-senior-management-for-students.md`

For now without the corresponding content change, there should be no visible difference to users.

Subsequent PRs will deal with (3), (4) and (5) from the Todo list above

### DIFF
--- a/app/components/card_component.html.erb
+++ b/app/components/card_component.html.erb
@@ -1,4 +1,4 @@
-<div class="card<%= " card--no-border" unless border %>">
+<div class="card <%= modifier_classes %>">
   <% if header %>
   <header class="card__header">
     <%= truncated_header %>
--- a/app/components/card_component.html.erb
+++ b/app/components/card_component.html.erb
@@ -7,6 +7,12 @@
 
   <%= media_link %>
 
+  <% if title %>
+  <h3>
+    <%= title %>
+  </h3>
+  <% end %>
+
   <div class="card__snippet">
     <p><%= snippet %><p>
   </div>
--- a/app/components/card_component.html.erb
+++ b/app/components/card_component.html.erb
@@ -1,8 +1,16 @@
-<div class="card">
+<div class="card<%= " card--no-border" unless border %>">
+  <% if header %>
+  <header class="card__header">
+    <%= header %>
+  </header>
+  <% end %>
+
   <%= media_link %>
+
   <div class="card__snippet">
     <p><%= snippet %><p>
   </div>
+
   <footer class="card__footer">
     <%= link_to(link, class: "git-link") do %>
       <%= link_text %> <%= helpers.fas_icon("chevron-right") %>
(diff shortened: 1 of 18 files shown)


app/components/card_component.html.erb
```diff
@@ -1,8 +1,22 @@
-<div class="card">
+<div class="card<%= " card--no-border" unless border %>">
```
peteryates, line 1: I think this can be simplified a bit by introducing a `card_classes` method on the component that builds an array

```erb
<%= tag.div("card", class: card_classes) do %>
...
```

spec/components/card_component_spec.rb
```diff
@@ -10,58 +10,102 @@
...
+    specify "border is removed" do
+      is_expected.to have_css(".card.card--no-border")
+    end
+  end
+
+  context "with header set" do
+    let(:card) { base.merge(header: "Further info") }
+
+    specify "header is shown" do
+      is_expected.to have_css(".card header")
```
peteryates, line 62: I think it makes sense to check the text matches here too

spec/components/card_component_spec.rb
```diff
@@ -10,58 +10,102 @@
...
+      end
+    end
+  end
+
+  context "with title set" do
+    let(:title) { "a title" }
+    let(:card) { base.merge title: title }
+
+    specify "title is shown" do
+      is_expected.to have_css(".card > h3")
```
peteryates, line 81: Also here (ensuring the right text is set)

peteryates (changes requested):
Looks great overall. Few minor comments

app/components/card_component.html.erb
jebw: Sure, I'll take a look

spec/components/card_component_spec.rb
jebw: Good call

spec/components/card_component_spec.rb
jebw: :+1:

peteryates (approved):
Looks great.

PR #625, opened 2020-12-01
Add DAX cookie policy details

ethax-ross:
This is for the LID tracking pixel.


jebw (approved):

peteryates:
Is it worth us linking to or explaining to users who/what DAX (and BAM) are? Or linking to them or something.

ethax-ross:
Yeah I think that would be a nice enhancement, one to bring up with Myles maybe? I'll merge this down for now as it can be picked up separately

PR #626, opened 2020-12-02
Make redirects not cause test failures

peteryates:
Testing for 200 (:success) is a little strict here, we just want to ensure that the request isn't returning an error.


jebw (approved):
LGTM as an immediate fix but I think we need to walk the redirect chain (with a protection from a loop) - otherwise there's a risk of a dangling redirects

peteryates:
Holding off on this, it's worth just fixing our content to prevent internal links from needing redirects.

PR #629, opened 2020-12-03
Fix margin for Train to Teach results box

ethax-ross:
When the train to teach results appear in a search the margin at the top is too small and it cuts off the top of the logo. This adds a bit of extra margin for the TTT results box.

Before:
Screenshot 2020-12-03 at 08 11 24

After:
Screenshot 2020-12-03 at 08 11 13


peteryates (approved):
Looks good

PR #631, opened 2020-12-03
Move calls to action components to more logical place

peteryates:
This is just a rearrangement of CTAs as now they'll be used in places other than the accordion.


spec/components/calls_to_action/next_steps_component_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "rails_helper"
+
+RSpec.describe CallsToAction::StoryComponent, type: :component do
```
ethax-ross, line 3: Typo in `describe` class name

ethax-ross (approved):

PR #640, opened 2020-12-07
Enabled randomisation of specs

jebw:
### Context

Tests which rely on the ordering of tests will prevent future parallelisation of the test suite

1. Turn on randomisation
2. Turn on recording the seed in case we need to replicate any flaky test failures


peteryates:
Closing this for the time being too, again we can reopen if we revisit.

PR #643, opened 2020-12-07
Improved page load speed by deferring VWO code loading

jebw:
### Trello card

https://trello.com/c/hL0FH0mA

### Context

Our page loads are slowed down by blocking on VWO loading.

Changed the VWO script to load using the defer strategy, this should result in
page rendering before it attempts to load VWO.

Changed the application.js script to load async instead of defer - this should
load as soon as available without blocking the page, and is necessary to avoid
blocking behind the defer'd VWO script since that gets injected higher up the
HEAD before our own application.js script tag.

Reference: https://www.digitalocean.com/community/tutorials/html-defer-async


ethax-ross (approved):
Looks good, I came across a JS issue when I was testing this but its unrelated so I'll raise it in another ticket.

PR #644, opened 2020-12-08
Simplify TalkToUs Stimulus controller

ethax-ross:
This controller previously had two targets, `chat` and `tta`. The `tta` target was unused so could be removed and the `chat` target was used to show the chat CTA if Javascript is enabled.

There was a bug introduced when another chat partial used the `TalkToUsController` but didn't specify a `chat` target. Instead, this commit refactors the controller to use the built in `element` target as the chat element. It now also accepts a display type as DOM state, since the other use case required a `flex` display type (and the initial use case used `inline-block`).

Updates the test cases to simplify the stub HTML and cover the `startChat` method.

### DIFF
--- a/app/components/calls_to_action/chat_online_component.html.erb
+++ b/app/components/calls_to_action/chat_online_component.html.erb
@@ -1,4 +1,4 @@
-<%= tag.div(class: %w[call-to-action call-to-action--chat-online], data: { controller: "talk-to-us" }) do %>
+<%= tag.div(class: %w[call-to-action call-to-action--chat-online], data: { controller: "talk-to-us", talk_to_us_display: "flex" }) do %>
   <%= icon %>
 
   <div class="call-to-action__content">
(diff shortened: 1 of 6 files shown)


jebw (changes requested):
I think this should be working a bit different - I think there should just be a 'state' css class getting added when its shown. And the CSS for that class should be responsible for deciding presentation (ie, `inline-block`)

There's actually a new feature in stimulus 2 to handle that scenario? Though maybe not worth tying this PR to stimulus 2.

ethax-ross:
Yeah that's fair I can switch it to use a class name instead. I was looking at the new Stimulus 2 features and it does have better state management APIs; I started an upgrade yesterday but parked it as it turned into a fair amount of changes - they have changed how you specify targets/actions and the find/replace is a bit involved as a bunch of ours are specified as Ruby hashes and others are inline, plus it will span the content and app repos still I _think_.

jebw:
Yeh, best not to tie the two together than.

jebw (approved):
I think the previous behaviour of setting a default if the `display-class` isn't set seems like a good idea? Other than that this seems ok.

I am wondering about a different approach to this might be cleaner though - really all the stimulus controller needs to know about is 'make this visible' - everything else is just presentation so can be handled in CSS? 

something like the following

```
.call-to-action--chat-online.visible { display: flex }
.call-to-action--chat-online.visible.show-inline { display: inline-block }
.call-to-action--chat-online { display: none }
```

So the html is declared (potentially with the presentation css class on it), and the stimulus controller just needs to add the 'visible' class without getting into what that will do?

ethax-ross:
I've updated to switch a `visible` class instead; I almost went thing route originally but decided against it as it spreads out the 'visible/invisible' logic in the CSS; I think its the right approach but I think this also needs a follow up PR to decouple the 'TalkToUsController' from its usage in the footer and make it a more generic 'ChatCallToActionController'

jebw (approved):

jebw (approved):

PR #647, opened 2020-12-08
Stop overwriting the frontmatter for partials

peteryates:
Now we're using markdown partials more and more we're hitting a problem where the template handler tries to set the front matter for each partial despite them having none.

This change will only amend the front matter if there is some present in the template we're currently rendering. This will pave the way for us breaking up pages into section partials and rendering CTAs inbetween.

If for some reason there _is_ front matter inside a partial the old behaviour will continue.

### DIFF
--- a/lib/template_handlers/markdown.rb
+++ b/lib/template_handlers/markdown.rb
@@ -32,6 +32,8 @@ module TemplateHandlers
 
     def call
       # inspect objects to Ruby strings which can be eval'd
+      return %(#{render.inspect}.html_safe) if front_matter.empty?
+
       %(@front_matter = #{front_matter.inspect}; #{render.inspect}.html_safe)
     end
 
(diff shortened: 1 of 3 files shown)


lib/template_handlers/markdown.rb
```diff
@@ -32,6 +32,8 @@ def initialize(template, source = nil, **options)
 
     def call
       # inspect objects to Ruby strings which can be eval'd
+      return %(#{render.inspect}.html_safe) if front_matter.empty?
```
ethax-ross, line 35: Can we stick in a quick unit test for this? for documenting why we're doing it as much as anything

peteryates: 👍🏽

ethax-ross (approved):

PR #650, opened 2020-12-08
update image for events module on homepage

sigmaben:
Add new images as high res version of old one not available.


peteryates (approved):
Looks good. I like the in-brand yellow polygon on the board!

PR #651, opened 2020-12-08
Make homepage steps into a column

peteryates:
### Trello card

N/A

### Context

This is an experiment to see whether just showing a regular list is clearer when the items are longer. They can look squished:

| Before | After |
| ----- | ----- |
| Screenshot from 2020-12-08 13-43-00 | Screenshot from 2020-12-08 13-52-24 | 

Make the list always vertical (rather than just when the screen is <800px wide)

### Guidance to review

There should probably be a discussion around this first.  cc @MylesJarvis @sigmaben

### DIFF
--- a/app/webpacker/styles/home.scss
+++ b/app/webpacker/styles/home.scss
@@ -4,15 +4,15 @@
     box-sizing: border-box;
     display: table;
     width: 100%;
-    
+
     &__left {
-        
+
         display: table-cell;
         vertical-align: top;
         width: 370px;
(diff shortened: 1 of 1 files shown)


MylesJarvis (approved):
Personally, I'm happy with this. I'd welcome Ben's view on whether there's anything else we can do to keep it horizontal. If not, we should go with this.

peteryates:
Closing for the moment, might revisit if the longer step names don't look great.

PR #653, opened 2020-12-08
Revert "Revert "Update validator to allow outward only postcodes in event search""

jkempster34:
Reverts DFE-Digital/get-into-teaching-app#628


jebw (commented):
Merging this should be held back until its confirmed the API changes are in Prod.

Otherwise there's a risk this gets deployed out to Prod ahead of the API changes

jebw (approved):
LGTM

PR #658, opened 2020-12-09
Upgrade to stimulus 2.0.0

peteryates:
### Trello card

N/A

### Context

[Stimulus 2.0.0 has been released](https://github.com/stimulusjs/stimulus/releases/tag/v2.0.0) and while there weren't any breaking changes that affect us the new version was generating lots of warnings. The main change seems to have been the switch of `data-target='controller.action'` to `data-controller-target='action'`.

Change the templates, components and specs that reference `data-target` to `data-controller-target`.

### Guidance to review

Everything _looks_ ok for me locally, the search form, cookies, accordion, video player etc all seem to work without generating warnings.

If any of them were missed we'll get warnings rather than breakages until `data-target` is officially deprecated.

Fixes #636, #637, #638

Also see related DFE-Digital/get-into-teaching-content/pull/203


jebw (approved):
Its still using the old `.data.get` syntax but I can't find if thats been deprecated in favour of Values or not.

I think that can get swapped at a future point

PR #659, opened 2020-12-10
Change styling of content headings

ethax-ross:
### Trello card

[Trello-596](https://trello.com/c/L9vaAu1m/596-development-implement-design-for-full-content-pages-where-the-heading-blue-underlines-are-removed)

### Context

Instead of having black text with a blue underline we want to have white text on a blue background.

- Change styling of content headings

### Guidance to review

Since we set this as the default styling on a `.content__left h2` there were a few instances where we override `h2` in other content components; I think I've updated them all but it would be good for someone else to do a quick run through.

| Before | After |
| ----------- | ----------- |
|    Screenshot 2020-12-10 at 10 47 40   | Screenshot 2020-12-10 at 10 47 30       |

### DIFF
--- a/app/webpacker/styles/accordion.scss
+++ b/app/webpacker/styles/accordion.scss
@@ -16,12 +16,12 @@
 
             .step-header__text {
                 text-align: left;
-                border: 0;
                 margin: 0;
                 font-size: 140%;
                 flex-grow: 2;
                 white-space: initial;
-                padding-right: 2em;
+                padding: 0 2em 0 0;
+                background: transparent;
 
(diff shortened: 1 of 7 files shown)


peteryates (approved):
Overall I think it looks good. Not a massive fan of `-20px` border, I think in the longer run we should be starting from the outside and working in, but that's a much bigger change which I'll get round to at some point 🤦🏽‍♂️ 

I think that we might need to look at our content on the back of this. Some pages are a bit overbearing:

Screenshot from 2020-12-10 11-27-07

PR #663, opened 2020-12-11
Update chat to new theme

ethax-ross:
We have a new 'theme' available for the chat that updates the logo and questions asked; this updates the `klick2contact` url so we are using it.

Screenshot 2020-12-11 at 08 33 35


peteryates (approved):
Looks good

PR #665, opened 2020-12-11
Revert "Update chat to new theme"

ethax-ross:
Reverts DFE-Digital/get-into-teaching-app#663 - Eckoh had to revert their change for the new theme this morning due to an issue in production, so we can't ship this just yet.


peteryates (approved):
👍🏽

PR #671, opened 2020-12-14
Add calls to action to right column

peteryates:
### Trello card

https://trello.com/c/RmC3uEST/537-content-check-journeys-and-ctas-from-returners-funding-salaries-internationals-taac-as-key-entry-points-and-align-sense-check-on

### Context

Calls to action are wanted on the international candidates page but the intention is to make use of the right column rather than placing them among the content.

Add a 'Narrow' CTA component and allow it to be initialised and automatically added to pages when the relevant frontmatter is present.

Screenshot from 2020-12-14 12-12-25

Screenshot from 2020-12-14 12-37-26

#### Final page

ways-to-train-with-right-cta

### Guidance to review

Also see https://github.com/DFE-Digital/get-into-teaching-content/pull/213 which is related

### DIFF
--- a/app/components/calls_to_action/narrow_component.html.erb
+++ b/app/components/calls_to_action/narrow_component.html.erb
@@ -1,0 +1,22 @@
+<%= tag.div(class: %w[call-to-action call-to-action--narrow]) do %>
+  <div class="call-to-action__content">
+    <header>
+      <%= icon %>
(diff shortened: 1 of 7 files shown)


MylesJarvis (approved):

ethax-ross (approved):
Looks good

PR #672, opened 2020-12-14
Fixed link to Cookies page from privacy policy

jebw:
### Trello card

https://trello.com/c/Bop3uAte/650-update-link-to-cookie-policy-in-the-privacy-policy-page

### Context

The HTML content for the cookies page in the CRM has a link to cookies which points to the BAU site.

This is held in the CRM so we need to fixup the html returned from the CRM.

1. Add a view helper to swap out the link

### Guidance to review

1. Highly specific but I don't think this is a scenario we'll encounter elsewhere, it can be generalised at that point if it is

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -94,4 +94,10 @@ module ApplicationHelper
 
     referer.gsub(root_url, root_path)
   end
+
+  def replace_bau_cookies_link(html_content)
+    html_content&.gsub \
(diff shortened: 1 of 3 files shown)


ethax-ross (approved):
Could we not just ask the CRM team to change the URL? as far as I'm aware this privacy policy only applies to our website

jebw:
I'll pick this up with the CRM team but this will resolve it for now

PR #673, opened 2020-12-14
Extracted no 'latest event' behaviour into separate component

jebw:
### Trello card

https://trello.com/c/xmUzKlT8

### Context

There's now a need to render a generic 'find an event card'.

1. Extract the 'fallback' behaviour out of the LatestEvent into a new FindEventsComponent
2. LatestEvent now chains to FindEvents if it can't find the event


ethax-ross (approved):
Nice

PR #674, opened 2020-12-15
Set active accordion step from data attr

peteryates:
This will allow accordions to have a step open automatically based on what's specified in the `data-accordion-active-step-value` attribute. It's so we can open the first 'Steps to become a teacher' step by default and leave the 'Ways to train' one fully closed.

Also see https://github.com/DFE-Digital/get-into-teaching-content/pull/215

### DIFF
--- a/spec/javascript/controllers/accordion_controller_spec.js
+++ b/spec/javascript/controllers/accordion_controller_spec.js
@@ -2,9 +2,8 @@ import { Application } from 'stimulus' ;
 import AccordionController from 'accordion_controller.js' ;
 
 describe('AccordionController', () => {
-
     document.body.innerHTML = `
-    <div data-controller="accordion">
+    <div data-controller="accordion" data-accordion-active-step-value="2">
         <section id="step-1" class="step" data-id="1">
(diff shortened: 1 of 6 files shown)


spec/javascript/controllers/accordion_controller_spec.js
```diff
@@ -32,38 +31,37 @@ describe('AccordionController', () => {
...
         </section>
     </div>
-    ` ;
+    `;
 
     const application = Application.start() ;
     application.register('accordion', AccordionController) ;
 
-    describe("when first loaded", () => {
+    describe("when first loaded with 'active' step set to '2'", () => {
```
ethax-ross, line 39: It may be because `location.href` can be a pain to mock in Jest, but it would be nice to have some coverage of `stepFromURI`.

ethax-ross (approved):

spec/javascript/controllers/accordion_controller_spec.js
peteryates: Agree, it was because it's a pain in Jest. Perhaps this is something that would be better handle by Cypress?

PR #675, opened 2020-12-15
Replace TypeEntity with PickListItem/LookupItem

ethax-ross:
### Trello card

[Trello-358](https://trello.com/c/K6lPR3ev/358-return-types-with-integer-ids-as-integer-rather-than-string)

### Context

The /types endpoints are being deprecated and have been split up into two new endpoints:

/lookup_items (types with a string id)
/pick_list_items (types with an integer id)

We need to replace all calls to the `TypesApi` with the replacement APIs.

- Replace TypeEntity with PickListItem/LookupItem

### Guidance to review

I've given this a decent amount of manual testing and I've also ran the integration tests against it, all coming back green.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -22,7 +22,7 @@ class EventsController < ApplicationController
   end
 
   def show_category
-    @type = GetIntoTeachingApiClient::TypesApi.new.get_teaching_event_types.find do |type|
+    @type = GetIntoTeachingApiClient::PickListItemsApi.new.get_teaching_event_types.find do |type|
       I18n.t("event_types.#{type.id}.name.plural").parameterize == params[:category]
     end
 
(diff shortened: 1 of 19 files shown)


app/controllers/events_controller.rb
```diff
@@ -22,15 +22,15 @@ def show
...
+    @type = GetIntoTeachingApiClient::PickListItemsApi.new.get_teaching_event_types.find do |type|
       I18n.t("event_types.#{type.id}.name.plural").parameterize == params[:category]
     end
 
     render(template: "errors/not_found", status: :not_found) && return if @type.nil?
 
     @event_search = Events::Search.new(event_filter_params.merge(type: @type.id))
     all_results = @event_search.query_events(MAXIMUM_EVENTS_IN_CATEGORY)
-    @events = paginate(all_results[@type.id.to_sym])
+    @events = paginate(all_results[@type.id.to_s.to_sym])
```
peteryates, line 33: Is the `.to_s` necessary here?

peteryates (approved):
Looks great. One minor question but it's inconsequential

app/controllers/events_controller.rb
ethax-ross: Yeah annoyingly you can't do `int.to_sym` you have to go to string first 🤷

PR #678, opened 2020-12-15
Move guidance routes below redirects

peteryates:
The *page wildcard param for `/guidance/*page` pages is making the route match before we've checked for redirects.

As per the more-general `*page` wildcard, this needs to be matched after we've done the redirect check.


jebw (changes requested):
For redirects to work with any 'wildcard' routes, they'll need to have priority above them

I think its the YAML block that needs to move much further up (but below healthcheck)

ethax-ross (approved):

jebw (approved):
LGTM

PR #681, opened 2020-12-16
Add a content spec that ensures no localhost links

peteryates:
This is on the back of a release that included one; it was missed by the other tests as they omit any link that starts with http or https.

Screenshot from 2020-12-16 09-51-34


jebw (approved):
LGTM

PR #682, opened 2020-12-16
Add no results message to event category results

ethax-ross:
### Trello card

[Trello-663](https://trello.com/c/N45f2PjE/663-add-no-results-message-to-event-category-page)

### Context

Currently if a user clicks the 'explore' button on an event category that has no results we are displaying an empty events container, which looks a bit broken. Instead, we want to display a no results message and direct the user back to the main search page so that they can try searching with a different type.

- Add no results message to event category
- Move no results HTML in to ViewComponent

### Guidance to review

Screenshot 2020-12-16 at 11 22 33

### DIFF
--- a/app/components/events/no_results_component.html.erb
+++ b/app/components/events/no_results_component.html.erb
@@ -1,0 +1,3 @@
+ <div class="no-results">
+  <%= content %>
+</div>
--- a/app/components/events/no_results_component.rb
+++ b/app/components/events/no_results_component.rb
@@ -1,0 +1,4 @@
+module Events
+  class NoResultsComponent < ViewComponent::Base
+  end
+end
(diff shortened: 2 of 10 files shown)


peteryates (approved):
Looks good to me.

jebw (commented):
Whats the advantage of swapping this to a component? AFAICT its just setting the CSS class? 

The alternative would be to make the component smarter - hand it the events list, and let it decide whether to show the message or not.

But I trend towards just a partial here since the only logic is an `if xxx.any?`

ethax-ross:
I thought about doing it as a partial but I figured we were favouring view components in general? I like the idea of introducing the view component previews as well for these, but I don't feel that strongly about it - happy to swap it out for a partial instead?

peteryates:
I like that the ViewComponent approach allows blocks to be passed in.

jebw:
No strong feelings from me.

In general I'm very in favour of ViewComponents - it didn't seem to bring much to this but if its more about achieving consistency of style that makes sense :+1:

PR #684, opened 2020-12-16
Featured story card

jebw:
### Trello card

https://trello.com/c/xmUzKlT8

### Context

Plan has changed somewhat and the idea is now to mark a story as featured, and any page with this card on will show details from that page.

Where possible it tries to mirror the behaviour of the existing Stories::CardComponent whilst just delegating whats shown to the story being linked to - eg links, snippet text, images

1. Add a simple query mechanism for the pages list in Pages::Frontmatter
2. Allow Pages::Page to look up the featured page using (1)
3. Added a FeaturedStory component to render the featured page as a card

### Guidance to review

1. Add `featured: yes` to the metadata for a story
2. Add the following snippet to another stories metadata

```yaml
explore:
  card_type: featured story
```

### DIFF
--- a/app/components/cards/featured_story_component.rb
+++ b/app/components/cards/featured_story_component.rb
@@ -10,7 +10,7 @@ module Cards
     end
 
     def link_text
-      "Read more"
+      @link_text || "Read #{story_name}'s story"
     end
 
     def border
@@ -18,22 +18,22 @@ module Cards
     end
 
     def image
-      @image ||= story.frontmatter.featured[:image] || story.image
+      @image ||= featured_metadata["image"] || story.image
     end
 
     def header
-      story.frontmatter.featured[:title] || story.title
+      @header ||= featured_metadata["title"] || "#{story_name}'s story"
     end
 
     def snippet
-      story.frontmatter.featured[:snippet] || raise(MissingSnippet, story.path)
+      @snippet ||= story.frontmatter.title || raise(MissingTitleOnFeatured, story.path)
     end
 
     class NoFeaturedStory < RuntimeError; end
 
-    class MissingSnippet < RuntimeError
+    class MissingTitleOnFeatured < RuntimeError
       def initialize(story_path)
-        super "Featured story is missing the 'snippet' - #{story_path}"
+        super "Featured story is missing a 'title' - #{story_path}"
       end
     end
 
@@ -41,6 +41,17 @@ module Cards
 
     def story
       @story ||= @page_data.featured_page || raise(NoFeaturedStory)
+    end
+
+    def story_name
+      story.frontmatter.story["name"]
+    end
+
+    def featured_metadata
+      case story.frontmatter.featured
+      when Hash then story.frontmatter.featured
+      else {}
+      end
     end
   end
 end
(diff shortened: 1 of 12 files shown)


app/components/cards/featured_story_component.rb
```diff
@@ -0,0 +1,57 @@
...
+      super(card: card)
+      @page_data = page_data
+    end
+
+    def link
+      story.path
+    end
+
+    def link_text
+      @link_text || "Read #{story_name}'s story"
```
ethax-ross, line 13: Better to use `pluralize` here? in case you have `Read Ross's story`

ethax-ross: Also I'm a bit confused at what this is doing, won't `@link_text` always be `nil`?

app/components/cards/featured_story_component.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+  private
+
+    def story
+      @story ||= @page_data.featured_page || raise(NoFeaturedStory)
+    end
+
+    def story_name
+      story.frontmatter.story["name"]
```
ethax-ross, line 47: Can be one for later, but I think it would be worth dropping the `story` key in the content frontmatter, not sure it needs it and it makes this read a bit funny

app/models/pages/frontmatter.rb
```diff
@@ -25,8 +31,15 @@ def perform_caching
       end
     end
 
-    def initialize(content_dir = nil)
-      @content_dir = content_dir ? Pathname.new(content_dir) : MARKDOWN_CONTENT_DIR
+    def initialize(content_dir_or_data = nil)
```
ethax-ross, line 34: This smells a bit to me - I may have misread but it looks like this has just been added for the test cases? I don't think we should be adding code to facilitate the test suite; could we create some mock markdown fixtures instead and use the existing initialiser?

ethax-ross (commented):
Looks good, just a couple of comments/questions. A general question - if the content is changed such that one of your custom errors is raised (`MissingTitleOnFeatured` and `MultipleFeatured`) will it result in a test failure or a 500 error in prod?

app/components/cards/featured_story_component.rb
jebw: No, with inheritance - it could be overridden in the card definition, that being said this should be `@link_text ||=`

app/components/cards/featured_story_component.rb
jebw: This is already there and used elsewhere - I'll change the `story` method

app/models/pages/frontmatter.rb
jebw: I was finding it difficult to test with and took the view if its hard to test than its probably assuming too much internally - hence allow for dependency injection

I can change this to a context instead though

jebw:
@ethax-ross Realised I didn't address this but a duplicate featured story will in an exception which will be caught on the branch in `-content` by the content specs failing 

Its an incorrect page structure and I think its better to 'fail early' before it hits master rather than leave the user wondering why prods showing the wrong 'featured story'

jebw:
I've also removed the ability to use the injected frontmatter data

ethax-ross (approved):

PR #685, opened 2020-12-16
Add page title to event/mailing list completion

ethax-ross:
### Trello card

[Trello-550](https://trello.com/c/9vh9fdOf/550-mailing-list-event-sign-up-completion-pages-dont-have-a-custom-page-title)

### Context

The event/mailing list sign up completion pages do not have titles set.

The event sign up wizard titles are also not working, though the logic is in place the `before_action` had been removed.

- Add page title to event/mailing list completion

Also adds tests to cover the page title behaviour of both wizards.

### DIFF
--- a/app/controllers/event_steps_controller.rb
+++ b/app/controllers/event_steps_controller.rb
@@ -1,9 +1,12 @@
 class EventStepsController < ApplicationController
   before_action :load_event
-  before_action :redirect_closed_events, only: %i[show update] # rubocop:disable Rails/LexicallyScopedActionFilter
 
   include WizardSteps
   self.wizard_class = Events::Wizard
+
+  before_action :redirect_closed_events, only: %i[show update]
+  before_action :set_step_page_title, only: [:show]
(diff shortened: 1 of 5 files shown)


app/controllers/event_steps_controller.rb
```diff
@@ -1,10 +1,15 @@
 class EventStepsController < ApplicationController
   before_action :load_event
-  before_action :redirect_closed_events, only: %i[show update] # rubocop:disable Rails/LexicallyScopedActionFilter
 
   include WizardSteps
   self.wizard_class = Events::Wizard
 
+  # rubocop:disable Rails/LexicallyScopedActionFilter
```
jebw, line 7: I think its probably better to exclude these files from that rule in rubocops config? Than there's just one place to add any future wizard controllers. Could potentially event be a wildcard on `app/controllers/**/*steps_controller.rb` ?

jebw (changes requested):

jebw (approved):
LGTM

PR #686, opened 2020-12-16
Reflect merge commit message through to `-content` repo

jebw:
### Trello card

N/a

### Context

Currently we just note the changed sha in the commits we create into the `-content` repo. If these are to be included in release notes than it would help to include further information about what was changed in the `-app` repo.

Copy the latest commit message in `-app` to the commit being created in `-content`.

### Guidance to review

This is on a branch at the moment - so I've disabled the push to `-content` and the check for `-master`.

Once reviewed I'll re-enable both of those, it'll require re-approval because its changed but than it can be merged

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -128,7 +128,7 @@ jobs:
   update_content_repo:
     name: Update the base image used in the Content repo
     runs-on: ubuntu-latest
-    # if: github.ref == 'refs/heads/master'
+    if: github.ref == 'refs/heads/master'
     needs: build
     steps:
       - name: Check out the repo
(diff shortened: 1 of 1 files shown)


sfawcett123 (commented):
do you needs to get the commit ?
    echo ::set-output name=message::$(git show -s --format=%B HEAD)
there should be a context..
     ${{ github.event.commits[0].message }}

jebw:
Ah thats better - tried dumping the github  object and it didn't show me anything - I'll update

.github/workflows/build.yml
```diff
@@ -128,9 +128,12 @@ jobs:
   update_content_repo:
     name: Update the base image used in the Content repo
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/master'
+    # if: github.ref == 'refs/heads/master'
```
sfawcett123, line 131: I am assuming this is for Debug

sfawcett123 (dismissed):
LGTM

.github/workflows/build.yml
jebw: YUp

sfawcett123 (approved):

PR #687, opened 2020-12-16
Fix JS error on guidance pages

ethax-ross:
### Trello card

[Trello-668](https://trello.com/c/FLC0rHxk/668-bug-going-back-from-ways-to-train-updates-the-address-bar-but-not-the-page-content)

### Context

We ran into a bug with Turbolinks imploding once you visit the a guidance page (so the browser back button failed to work). This was caused by a JS exception on the guidance pages, originating from some include JS. Removing the offending JS fixes the issues.

- Fix JS error on guidance pages

### Guidance to review

I gave the guidance page a good click about and nothing appears to be effected by removing this. We're also going to be removing/restyling these pages in the near future.


jebw (approved):
LGTM

PR #688, opened 2020-12-16
Update heading styles

ethax-ross:
### Trello card

[Trello-610](https://trello.com/c/2h6RhdYC/610-development-revise-event-page-headings-remove-bold-from-time-and-date-line)

### Context

There is too much bold text going on in event titles and the date/time; only the event title needs to be bold.

- Normal font weight for event date/time
- Remove blue background from event sign up sub-heading

This was introduced inadvertently when we moved all content `h2` tags to a new style that has a blue background and white text (dropping the previous `border`).

### Guidance to review

| Before      | After |
| ----------- | ----------- |
| Screenshot 2020-12-16 at 16 10 40    |  Screenshot 2020-12-16 at 16 10 19|
|   Screenshot 2020-12-16 at 16 10 43    | Screenshot 2020-12-16 at 16 10 28       |

### DIFF
--- a/app/webpacker/styles/event.scss
+++ b/app/webpacker/styles/event.scss
@@ -5,7 +5,6 @@
     }
 
     &__date {
-      font-weight: bold;
       font-size: 1.4em;
     }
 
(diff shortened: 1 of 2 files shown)


peteryates (approved):
Looks good.

PR #690, opened 2020-12-17
Revert "Add 15 minute cache headers to all static assets in /public"

jebw:
Reverts DFE-Digital/get-into-teaching-app#689


jebw:
Opened as a failsafe, not required

PR #691, opened 2020-12-17
Fix slanting hero from blocking top nav

peteryates:
Setting the overflow to `hidden` will make any part of the heading contents that falls outside of the container hidden. This is needed because if the window is made wide enough it'll eventually overlap with the top nav.

Screenshot from 2020-12-17 14-23-31


jebw (approved):
LGTM

PR #692, opened 2020-12-17
Add test to check internal images

ethax-ross:
### Trello card

[Trello-599](https://trello.com/c/MRt1mo8u/599-development-add-checking-for-missing-images-to-the-content-checks)

### Context

At the moment it is possible to incorrectly reference an internal image, resulting in a broken image being pushed live.

We want to add a test to verify all internal images can be successfully requested.

- Add test to check internal images

### Guidance to review

I've limited this to internal images; arguably we may want to also have a test case to catch any links to external images, I can't think of a reason we wouldn't vendor an image?


jebw (commented):
LGTM

jebw (approved):

PR #693, opened 2020-12-18
Change teacher training step to radio buttons

ethax-ross:
### Trello card

[Trello-682](https://trello.com/c/AVp3XQcs/682-development-change-we-need-some-more-details-from-drop-down-to-radio-buttons-on-mailing-list)

### Context

Switch the teacher training step answers from a drop down to radio buttons.

- Change teacher training step to radio buttons

Also updates `WizardSteps` to handle posting no data (which happens if no radio values are checked). Rails form builder has supported this natively with a hidden field for a while, but the GOV.UK form builder doesn't seem to.

### Guidance to review

Screenshot 2020-12-18 at 09 02 23

### DIFF
--- a/app/controllers/concerns/wizard_steps.rb
+++ b/app/controllers/concerns/wizard_steps.rb
@@ -75,7 +75,7 @@ private
   end
 
   def step_params
-    params.require(step_param_key).permit @current_step.attributes.keys
+    params.fetch(step_param_key, {}).permit @current_step.attributes.keys
   end
 
   def step_param_key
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/wizard_steps.rb
```diff
@@ -75,6 +75,8 @@ def completed_step_path
   end
 
   def step_params
+    return {} unless params.key?(step_param_key)
```
jebw, line 78: You could switch to using `fetch` with a default param instead?

`params.fetch(step_param_key, {}).permit ....`

jebw (dismissed):
Looks fine but the strong params line could be simplified a bit

jebw (approved):
LGTM

PR #700, opened 2020-12-18
Check the HTML pages as well as markdown since the homepage is html

jebw:
### Trello card

Found working on https://trello.com/c/Z1iYnBtC

### Context

We don't currently do page checks on the home page and should be

1. Add html pages to the list of pages being checked
2. Added IPv6 localhost address
3. Fixed the error messages on broken links to interpolate

### Guidance to review

1. Requires https://github.com/DFE-Digital/get-into-teaching-content/pull/226 merging first because this'll break the test suite as `-content` currently stands


peteryates (approved):
Makes sense.

PR #701, opened 2020-12-18
Added check that mobile nav has the same links as desktop nav

jebw:
### Trello card

https://trello.com/c/Z1iYnBtC

### Context

We nearly made a release which had incorrect navigation on the mobile pages

1. Check that desktop nav includes links for all pages marked as `navigation:`
2. Check that mobile nav includes links for all pages marked as `navigation:`
3. Check that all the links on the desktop nav appear in the mobile nav

### Guidance to review

1. Do the content specs still pass

### DIFF
--- a/spec/features/content_pages_spec.rb
+++ b/spec/features/content_pages_spec.rb
@@ -97,7 +97,7 @@ RSpec.feature "content pages check", type: :feature, content: true do
 
       let(:navigation_pages) { Pages::Frontmatter.select(:navigation) }
 
-      scenario "navigatable pages appear in desktop navbar" do
+      scenario "navigable pages appear in desktop navbar" do
         navigation_pages.each do |url, frontmatter|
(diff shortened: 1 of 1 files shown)


spec/features/content_pages_spec.rb
```diff
@@ -87,4 +88,38 @@ def content_urls
       end
     end
   end
+
+  content_urls.slice(0, 1).each do |first_url|
```
peteryates, line 92: `content_urls.first.tap do |first_url|` could be easier to follow here?

peteryates (commented):
Looks good, just one comment about the outer loop. Also 'navigable' :)

spec/features/content_pages_spec.rb
jebw: :+1:

peteryates (approved):
Looks good

PR #710, opened 2020-12-22
Automatic authentication using long-lived access tokens

ethax-ross:
### Trello card

[Trello-661](https://trello.com/c/ebrz8S9V/661-development-work-with-crm-team-on-developing-approach-to-migrating-past-registrants-on-to-mailing-list-subscription)

### Context

We are adding API endpoints that can generate 'long lived' access tokens against candidate records in the CRM. The idea being the CRM will call this endpoint to generate a token for a candidate then send that candidate an email containing the mailing list sign up 'magic link' link, including their candidate id and token:

```
/mailinglist/signup?candidate_id=f4488e35-8fd5-4884-af9f-ca98f76e0432&token=3097d0cb-5b9d-450f-8b4a-1c5f6dbeb6a2
```

When a candidate opens their 'magic link' the application can use the `candidate_id` and `token` to authenticate the candidate and retrieve their details from the API, pre-filling the mailing list form and taking them to the first step _after_ the `Authenticate` step (so skipping entering their name/email and verification code).

- [Temp] Point app at local API instance (for testing)
- Authenticate using long-lived access tokens

### Guidance to review

I think this needs more restructuring; we currently 'authenticate' and 'prefill' within the `Wizard` class and the `Authenticate` step; I think this could/should probably be consolidated so the logic is in the `Wizard` alone.

I need to gracefully handle when an authentication fails still; this is TBC.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,2 +1,4 @@
 # Add public visible environment variables here for the Developemnt environment
 TTA_SERVICE_URL="https://get-teacher-training-adviser-service-dev.london.cloudapps.digital/"
+GIT_API_TOKEN="abc123"
+GIT_API_ENDPOINT="https://localhost:5001/api"
--- a/Gemfile
+++ b/Gemfile
@@ -47,7 +47,7 @@ gem "prometheus-client"
 
 gem "sentry-raven"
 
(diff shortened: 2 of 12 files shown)


ethax-ross:
Closing this spike - raising the actual implementation as a new PR.

PR #723, opened 2021-01-04
Rubocop Rspec Step 1 updates

jebw:
### Trello card

https://trello.com/c/D6sNREFU

### Context

We are using RSpec-rubocop to enforce a consistent set of rules for specs, currently with a lot of checks disabled.

1. Enabled various rules which can be enabled without invasive changes to specs.

### Guidance to review

I don't think any of the rules being enabled should be controversial.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -58,10 +58,6 @@ RSpec/DescribeClass:
   Exclude:
     - spec/**/*.rb
 
-RSpec/DescribedClass:
-  Exclude:
-    - spec/**/*.rb
-
 RSpec/EmptyLineAfterFinalLet:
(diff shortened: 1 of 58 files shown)


peteryates (dismissed):
Looks good, although I prefer naming the class explicitly (rather than via `described_class`) unless it's in a shared example.

jebw:
I think it comes back to the DRY principle - if you're duplicating the class name its another thing to change if a class gets renamed, it does also makes it easier to extract to shared_examples/shared_context if suitable as well.

ethax-ross (approved):

ethax-ross (approved):

PR #727, opened 2021-01-05
Update Event::Search to explicitly search future events only

ethax-ross:
### Trello card

[Trello-672](https://trello.com/c/Mx2EzMU4/672-development-create-an-archive-repository-for-online-events-which-are-3-4-months-old)

### Context

Currently the `Event::Search` class relies on the API only returning future-dated events (as we only cache upcoming events from the CRM).

Going forward that will change as we're displaying archived/past events in the website. Before this change happens in the API we need to explicitly search future events _only_ in the application.

This commit ensures that when no month is selected we are only going to search for events after the start of the current day.

Going forward it may be worth deprecating the `upcoming_events_indexed_by_type` and just relying on the search endpoint, which can do this anyway when given today as the `start_after` parameter.

- Update Event::Search to explicitly search future events only

### Guidance to review

Upcoming API PR: https://github.com/DFE-Digital/get-into-teaching-api/pull/434

### DIFF
--- a/app/models/events/search.rb
+++ b/app/models/events/search.rb
@@ -76,7 +76,7 @@ module Events
     end
 
     def start_of_month
-      return nil if month.blank?
+      return DateTime.now.utc.beginning_of_day if month.blank?
 
       DateTime.parse("#{month}-01 00:00:00").beginning_of_month
     end
--- a/spec/models/events/search_spec.rb
+++ b/spec/models/events/search_spec.rb
@@ -123,6 +123,17 @@ describe Events::Search do
             receive(:search_teaching_events_indexed_by_type).with(**expected_attributes.merge(postcode: "TE57 1NG"))
         end
       end
+
+      context "when the month is nil" do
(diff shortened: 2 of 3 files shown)


app/models/events/search.rb
```diff
@@ -76,7 +76,7 @@ def query_events_api(limit)
     end
 
     def start_of_month
-      return nil if month.blank?
+      return DateTime.now.beginning_of_day if month.blank?
```
jebw, line 79: Does handle timezone? Should this be `Time.zone.now.beginnging_of_day`

Or should it be the start of day for the users timezone?

jebw (commented):
I'm wondering if it requires handling the timezone

app/models/events/search.rb
ethax-ross: Good catch, the API expects UTC - I'll update that now.

jebw: :+1:

jebw (approved):

PR #711, opened 2020-12-22
CSS refactor

peteryates:
### Trello card

N/A

### Context

Making style changes is sometimes a bit of a chore because we are relying on [float](https://developer.mozilla.org/en-US/docs/Web/CSS/float) for laying out pages and using lots of very-specific `px` measurements.

### Changes proposed in this pull request

Try to remove unnecessary CSS and markup, favouring flexbox, semantic elements and moving HTML to components where necessary.

* [x] Update the story layouts and styles
* [x] Update the content layout and styles
* [x] Update the accordion layout and styles
* [ ] Update the homepage layout and styles

It's going to be quite a big one.

PR #728, opened 2021-01-05
CSS layout refactoring

peteryates:
### Trello card

[444](https://trello.com/c/IinPwbxD/444-consistency-of-naming-in-css-ideally-removing-naming-and-relying-on-element-names-where-possible)

### Context

This is a tentative first step towards reorganising our layouts. The previous attempt, #711, was closed as it became a can of worms. Here we'll introduce the changes tested in #711 bit by bit.

Once the transition is complete the old styles can be removed and the remaining ones tidied up.

This is the first part of a migration away from the `content` and `content-1000` float-based layouts and towards a simpler flexbox model.

Here we introduce the `semantic` class; this is a stopgap to target only the pages we're transitioning while the migration is in flow. Once the old layouts are removed we can drop the class and simplify further.

The intent is to structure the contents of `main` based on the direct child elements. Currently there are two; `article` with `aside` and a 70/30 split and `.feature` - intended for things like index pages and lists of events - which is full-width.

##### With aside content:
Screenshot from 2021-01-05 10-56-55

##### Without aside content:
Screenshot from 2021-01-05 10-57-30

###### Mobile

Screenshot from 2021-01-05 11-07-36

###### Resizing the page

https://user-images.githubusercontent.com/128088/103881953-0c52b300-50d3-11eb-83cd-20bfd54a0e5d.mp4

### Guidance to review

Does it look ok?

### DIFF
--- a/app/components/stories/story_component.html.erb
+++ b/app/components/stories/story_component.html.erb
@@ -1,47 +1,46 @@
-<div class="content-wrapper">
-  <div class="content__left">
-    <article class="story">
-      <%= helpers.back_link(backlink, text: tag.span(backlink_text)) %>
+<article class="story markdown">
+  <%= helpers.back_link(backlink, text: tag.span(backlink_text)) %>
 
-      <%= tag.h1(title) %>
+  <%= tag.h1(title) %>
 
-      <header class="story__header">
-        <% if image.present? %>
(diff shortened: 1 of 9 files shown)


ethax-ross (approved):
Makes sense 👍

PR #730, opened 2021-01-05
Update Event::Search to explicitly search future events only (take #2)

ethax-ross:
Re-instates https://github.com/DFE-Digital/get-into-teaching-app/pull/727 (reverted due to a bug resulting in a 500 error).

The fix is pulled in as part of the updated ApiClient lib: https://github.com/DFE-Digital/get-into-teaching-api-ruby-client/pull/28

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,11 +1,11 @@
 GIT
   remote: https://github.com/DFE-Digital/get-into-teaching-api-ruby-client.git
-  revision: fc55a73e7192f95199f4c3d2e9c04c3e7eb6ada3
+  revision: f04da90eb269eb537a614248596536926f20f97b
   specs:
     get_into_teaching_api_client (1.1.7)
       json (~> 2.1, >= 2.1.0)
       typhoeus (~> 1.0, >= 1.0.1)
-    get_into_teaching_api_client_faraday (0.1.9)
+    get_into_teaching_api_client_faraday (0.1.10)
       activesupport
       faraday
       faraday-encoding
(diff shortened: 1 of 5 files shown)


config/initializers/date_formats.rb
```diff
@@ -3,5 +3,7 @@
 Date::DATE_FORMATS[:yearmonth] = "%Y-%m"
 Date::DATE_FORMATS[:humanmonthyear] = "%B %Y"
 Date::DATE_FORMATS[:long] = "%-d %B %Y"
+# Important: the GetIntoTeachingApiClient will use this format - do not change it!
+Time::DATE_FORMATS[:default] = "%Y-%m-%d %H:%M:%S %z"
```
jebw, line 7: I don't think we want this as the default format within the app - I think this should probably be getting fixed in the API client, but failing that the code should specify the serialization format for interactions with the API

jebw (changes requested):
We should hold off changing this until we've got the JS test suite updated to test those pages in the QA process

Also I think we need to consider how 'dates with zones' are handled - this should not require changing the default way dates are shown across the site.

If it does need to be handled at app level, it'll need tests to go with it.

config/initializers/date_formats.rb
ethax-ross: We don't appear to use the default date format anywhere in the app at the moment so I figured it was safe to change - you could argue that we shouldn't rely on Ruby's default date format in the app anyway and it would be better to override it regardless, as it could change in a future version of Ruby.

I get it feels like an awkward place to be fixing it though. I'll look at if the API could be updated to support this format, or I could look at what monkey patching the ApiClient would look like but that feels iffy to me as well, it would probably require a call like:

```
when 'DateTime'
  date.change(offset: "+00:00").to_s if date.utc?
```

ethax-ross:
Closing in favour of https://github.com/DFE-Digital/get-into-teaching-api-ruby-client/pull/28

config/initializers/date_formats.rb
ethax-ross: I'll drop the second commit here and update the gem to pull in the fix:

https://github.com/DFE-Digital/get-into-teaching-api-ruby-client/pull/28

jebw (approved):
LGTM

PR #734, opened 2021-01-06
Added support for optional preview views folder

jebw:
### Trello card

https://trello.com/c/pdn3qdOK

### Context

The docker images built from this repo should be verifiable on a basic level without adding in the `-content` repo. This is already possible from the `Rails.env.test?` because it adds a custom view path with additional templates to use.

We should add one with page templates for visual testing

1. Add an additional `app/preview_pages` folder for templates - this can be enabled with the environment variable `PREVIEW_PAGES=1`. This is set by default in the rails environments `development`, `test`
2. This can be disabled locally if using symlinks to `-content` by setting `PREVIEW_PAGES=0` in `.env.local`
3. Add a test home page - the templates can be expanded later as needed
4. `Pages::Frontmatter` has been taught to handle multiple content directories, and by default will use all view_paths originating within the `Rails.root` rather than just the first one
5. The "prepend fake views" rspec support code now always includes the folder of templates from the rspec folder - previously with was being repeatedly added causing a performance regression on tests when `Frontmatter` started using all paths

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,2 +1,3 @@
 # Add public visible environment variables here for the Developemnt environment
 TTA_SERVICE_URL="https://get-teacher-training-adviser-service-dev.london.cloudapps.digital/"
+PREVIEW_PAGES=1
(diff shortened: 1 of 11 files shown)


ethax-ross (dismissed):

peteryates (approved):
Looks great.

PR #735, opened 2021-01-06
Introduce event type archives and enable for Online Events

ethax-ross:
### Trello card

[Trello-672](https://trello.com/c/Mx2EzMU4/672-development-create-an-archive-repository-for-online-events-which-are-3-4-months-old)

### Context

We want to be able to view an archive of online events (only) going back a period of 4 months. The archive page should be identical to the event category page, but with "Past" suffixed to the category name copy and instead of showing the next 5 months of events it should display the previous 4 months of events. This should display up to and including the day before today, but no events on the current date (even if the event time has passed as they stay on the main event category page until the following day).

In addition, we only want to support archives for the online event type initially and we want to add a callout at the bottom of the online event category page that highlights the archive and links users through to view it.

- Support viewing archived online events

This commit adds a new route `/category/:category/archive` that points to the existing event category controller action and reuses the existing view.

The action is updated to determine if the event type supports an archive and to configure the event search `period` to be either `past` or `future`.

The `Event::Search` module contains the majority of the updates, which stem from a new `period` parameter that supports either `:future` (current behaviour) or `:past` (for archived events). We use this parameter to determine the `available_months` to generate along with capping the earliest/latest dates to query.

### Guidance to review

Does this feel like the right approach? is the `Event::Search` module or the `show_category` action doing too much? I think it's worth splitting the event category logic out into its own `EventCategoriesController` in a follow up PR, and probably splitting it into two distinct actions that use the same view which should simplify things futher.

screencapture-0-0-0-0-3000-events-category-online-events-archive-2021-01-07-10_21_24

Screenshot 2021-01-06 at 14 50 50

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -26,9 +26,16 @@ class EventsController < ApplicationController
       I18n.t("event_types.#{type.id}.name.plural").parameterize == params[:category]
     end
 
-    render(template: "errors/not_found", status: :not_found) && return if @type.nil?
+    raise_not_found && return if @type.nil?
 
-    @event_search = Events::Search.new(event_filter_params.merge(type: @type.id))
+    @is_archive = params[:archive]
+    has_archive = has_archive?(@type)
+
+    raise_not_found && return if @is_archive && !has_archive
+
+    period = @is_archive ? :past : :future
+    @event_search = Events::Search.new(event_filter_params.merge(type: @type.id, period: period))
+
     all_results = @event_search.query_events(MAXIMUM_EVENTS_IN_CATEGORY)
     @events = paginate(all_results[@type.id.to_s.to_sym])
   end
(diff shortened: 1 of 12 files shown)


app/controllers/events_controller.rb
```diff
@@ -26,11 +26,21 @@ def show_category
...
+    raise_not_found && return if @is_archive && !has_archive
+
+    period = @is_archive ? :past : :future
+    @show_archive_link = !@is_archive && has_archive
+
+    @event_search = Events::Search.new(event_filter_params.merge(type: @type.id, period: period))
+    events_by_type = @event_search.query_events(MAXIMUM_EVENTS_IN_CATEGORY)
+    group_presenter = Events::GroupPresenter.new(events_by_type, false, @event_search.future?)
+    events_of_type = group_presenter.sorted_events_by_type.first { |v| v.first == @type.id }&.last
+    @events = paginate(events_of_type)
```
peteryates, line 43: This feels a little heavy/intense for a single controller method

peteryates (approved):
Approved, I think the approach is good overall

I do think that the `#show_category` method could perhaps be simplified or separated out a bit, there's a lot going on.

ethax-ross:
Agreed 👍 I'm planning on following this up with a PR to split out that action into a new `EventCategoriesController`

PR #738, opened 2021-01-06
Added initial trello workflow

jebw:
### Trello card

https://trello.com/c/2CUilw5w

### Context

When a PR is opened with a link to the associate Trello card, automatically add a comment to the Trello card with the associate Pull Request

1. Create a comment in the trello card with a link to back to the PR

### DIFF
--- a/.github/workflows/trello.yml
+++ b/.github/workflows/trello.yml
@@ -1,0 +1,42 @@
+name: Link Trello card
+
+on:
+  pull_request:
+    types: [ opened ]
+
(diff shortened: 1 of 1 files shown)


sfawcett123 (approved):
Looks fun

I would suggest that code like this where you have to write specific actions is put in a common repository like https://github.com/DFE-Digital/github-actions so that it can be reused in the future.

jebw:
Sure I'll send them a pull request

PR #742, opened 2021-01-08
Transition to search_teaching_events_grouped_by_type

ethax-ross:
### Trello card

[Trello-649](https://trello.com/c/yBEK1KXr/649-dev-refactor-search-indexed-by-endpoints-to-have-a-better-response)

### Context

We currently use the `search_teaching_events_indexed_by_type` and `upcoming_teaching_events_indexed_by_type` queries, which return in the format:

```
{
  "<type_id>": [... events ...],
  ...
}
```

Instead, the API now has a new `search_teaching_events_grouped_by_type` endpoint that returns results in a more rest-ful format:

```
[
  { "typeId": <type_id>, "teachingEvents": [... events ...] },
   ...
]
```

Note the upcoming endpoint has not been udpated, and instead we are removing it completely as it can be achieved by passing a `startAfter` parameter to the search method.

All occurances of calls to `..._indexed_by_..` have been updated.

- Transition to search_teaching_events_grouped_by_type

- Bump ApiClient version

Pulls in the new `search_teaching_events_grouped_by_type` method.

### Guidance to review

I've given this a decent test locally to ensure all the event searching etc works as expected.

### DIFF
--- a/spec/support/helpers/events.rb
+++ b/spec/support/helpers/events.rb
@@ -1,0 +1,9 @@
+module Helpers
+  module Events
+    def group_events_by_type(events)
+      events.group_by(&:type_id).map do |type_id, events_of_type|
+        GetIntoTeachingApiClient::TeachingEventsByType.new(typeId: type_id, teachingEvents: events_of_type)
+      end
+    end
+  end
+end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,11 +1,6 @@
 GIT
-  remote: https://github.com/waiting-for-dev/front_matter_parser.git
-  revision: 71d1d61c4feeb2c73de7d986a7f97e9ab931917e
-  specs:
-    front_matter_parser (0.2.1)
-
-PATH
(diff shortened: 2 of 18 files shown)


spec/support/helpers/events.rb
```diff
@@ -0,0 +1,9 @@
+module Helpers
```
jebw, line 1: This is already a top level module in rails, and _I think_ will include automatically include all modules below it in Controllers and Views - potentially causing issues for specs in the future.

I'd suggest renaming this to something like `SpecHelpers` ?

jebw (changes requested):
LGTM apart from the use of `Helpers` module in the `spec/support` space :+1:

spec/support/helpers/events.rb
ethax-ross: Oh ok, I was going off the RSpec docs but happy to rename it:

https://relishapp.com/rspec/rspec-core/docs/helper-methods/define-helper-methods-in-a-module

jebw (approved):
LGTM

PR #744, opened 2021-01-08
Move event category actions into EventCategoriesController

ethax-ross:
### Trello card

[Trello-706](https://trello.com/c/qwTLpPgS/706-split-category-action-out-into-eventcategoriescontroller-and-tidy-up)

### Context

Currently we have one, reasonably complex action in the `EventsController` that deals with showing an event category and its associated archive.

This commit pulls that action out and splits it into two new actions; `show` and `show_archive` in a new `EventCategoriesController`.

The routes have also been updated to better match what you would expect from a Rails app, so:

```
/event_categories/:id
/event_categories/:id/archive
```

Redirects are in place to ensure any old/bookmarked traffic gets redirected, though we can probably remove the redirects in a few days if we observe no traffic going to the old endpoints.

- Move event category actions into EventCategoriesController

### DIFF
--- a/app/components/events/type_description_component.rb
+++ b/app/components/events/type_description_component.rb
@@ -20,7 +20,7 @@ module Events
     end
 
     def read_more_href
-      event_category_events_path(type_name_plural.parameterize)
+      event_category_path(type_name_plural.parameterize)
     end
 
   private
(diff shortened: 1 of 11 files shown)


peteryates (approved):
Much nicer 👍🏽

PR #748, opened 2021-01-11
Clean up online/offline/virtual logic

ethax-ross:
### Trello card

[Trello-695](https://trello.com/c/ON3HS3MN/695-clean-up-online-moved-online-virtual-naming-and-references)

### Context

An event can have three different 'states':

- Offline (in person)
- Online (fully online with no building)
- Virtual (online, but has an associated building - aka moved online)

The API client has been updated with an `is_virtual` attribute that enables the app to determine these states more succinctly.

- Clean up event online/offline/virtual checks

Also removes `condensed` as an option from `EventBoxComponent` as this is no longer used anywhere.

### DIFF
--- a/app/components/events/event_box_component.html.erb
+++ b/app/components/events/event_box_component.html.erb
@@ -1,13 +1,11 @@
 <div class="event-box">
     <header class="event-box__header">
-      <h4><%= heading %></h4>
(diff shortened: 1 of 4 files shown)


peteryates (approved):
This makes sense and feels much cleaner 👍🏽

PR #750, opened 2021-01-11
Added a varient of Call to action button which excludes the chevron

jebw:
### Trello card

https://trello.com/c/cmJ1aEdI

### Context

We require a page providing access to the slides from presentations. That page is in `-content` but requires a tweak to the CSS style

1. Add varient of call-to-action button which doesn't include the chevron


peteryates (approved):
Makes sense.

PR #752, opened 2021-01-12
Bump API client version

ethax-ross:
Removes the `x_teaching_events_indexed_by_type` methods, which are no longer used in the app


sfawcett123 (approved):
LGTM

PR #756, opened 2021-01-14
Add FeatureTableComponent for Ways to Train redesign

ethax-ross:
### Trello card

[Trello-708](https://trello.com/c/1lP8Kcde/708-development-ways-to-train-table-style-layout)

### Context

We are implementing a redesign of the Ways to Train layout, which replaces the current accordion design with a series of 'feature tables'.

This PR introduces the `FeatureTableViewComponent` that will be used in the redesigned content.

- Exclude /rails from default content route

Currently the main content route will try and map `/rails` to a content page. This excludes it so that we can access the in-built
Rails routes in development (for ViewComponent previews). 

- Add FeatureTableComponent

This is a table component that will be used in the new Ways to Train page. It consists of a strapline heading and two column table in desktop, with the first column being the heading and the second column the associated value. In mobile this changes to a single column layout, alternating headings and values.

- Enable ViewComponent previews

Configures `ViewComponent` previews that can be accessed at:

```
/rails/view_components
```

Adds a basic view component layout and the `FeatureTableComponentPreview`.

- Switch table to definition list

Uses a definition list instead of a table, which simplifies the HTML. Currently the first column of the list in desktop mode is fixed-width as trying to implement a variable width first column is not going to be straight forward -- this will be picked up in a follow up PR if/when we have another use case for this table.

### Guidance to review

The CSS isn't too pretty due to the overlap of the strap-line and the need to be flush with the page width on mobile; open to any suggestions if it can be improved. 

| Desktop      | Mobile |
| ----------- | ----------- |
| Screenshot 2021-01-14 at 10 49 11      | Screenshot 2021-01-14 at 11 02 23       |

### DIFF
--- a/app/components/content/feature_table_component.html.erb
+++ b/app/components/content/feature_table_component.html.erb
@@ -1,0 +1,11 @@
+<%= wrapper do %>
+  <%= heading %>
+  <%= table do %>
+    <tbody class="desktop">
+      <%= desktop_rows %>
+    </tbody>
+    <tbody class="mobile">
+      <%= mobile_rows %>
+    </tbody>
+  <% end %>
+<% end %>
--- a/app/components/content/feature_table_component.html.erb
+++ b/app/components/content/feature_table_component.html.erb
@@ -1,11 +1,25 @@
-<%= wrapper do %>
-  <%= heading %>
-  <%= table do %>
+<div class="<%= wrapper_class %>">
+  <% if title? %>
+    <h2 class="strapline strapline--blue"><%= title %></h2>
+  <% end %>
+  <table aria-label="<%= description %>">
     <tbody class="desktop">
-      <%= desktop_rows %>
+      <% data.map do |key, value| %>
+        <tr>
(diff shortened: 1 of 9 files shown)


jebw (commented):
I'm a bit unsure on this for a couple of reasons

1. All the HTML generation is being done dynamically in the component file - I suggest either implement `def render` and generate the output entirely in the component - or move the html to the html file

2. Whilst this is rendered using an html table - its actually a definition list - and if a definition list is used instead than you can use a single html structure with CSS break points control whether to show vertically or horizontally?

ethax-ross:
Happy to move the HTML into the ERB if you think that makes more sense; I just find it easier to derive the structure of the HTML if the 'dynamic' bits are abstracted into the component itself.

For the definition list I'd argue that the contents better fit the semantics of a table -  a definition list should:

> consist of two parts: a term and a description

I don't think the values we represent here fit that, hence went for a table.

peteryates (approved):
Looks good and works well for me.

PR #757, opened 2021-01-14
Add COVID-19 link to footer

peteryates:
### Trello card

https://trello.com/c/yt7r00D8/716-add-a-covid-19-link-to-the-footer

### Context

Once the COVID banner has been dismissed there's no way for users to find the link to it.

Add a COVID-19 link to the end of the nav options in the footer.

Also remove unused indexing on the links loop.

### Guidance to review

Screenshot from 2021-01-14 11-35-12

### DIFF
--- a/app/views/sections/_footer.html.erb
+++ b/app/views/sections/_footer.html.erb
@@ -26,7 +26,7 @@
       </div>
       <div class="site-footer-top__links-container">
         <%= link_to("Home", page_path(page: :home)) %>
-        <% navigation_resources.each_with_index do |resource, index| %>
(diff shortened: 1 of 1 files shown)


jebw (dismissed):

jebw:
I've approved this but is it worth instead having a way of marking content pages (ie frontmatter) to show only in the footer - rather than explicitly linking each one out the app (or is the covid-19 page in `-app` rather than `-content`?)

peteryates:
Yeah, agree totally on moving this to the content repo. Will do that separately.

jebw (approved):

PR #758, opened 2021-01-14
Added additional File types to labeller

sfawcett123:
Added additional file types to labeller
ERB and SCCS

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,8 +1,9 @@
 ---
 
 Style:
-  - app/**/*.sccs
-
+  - app/**/*.scss
+  - app/**/*.css
+  
 Ruby:
   - app/**/*.rb
   - app/**/*.erb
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -1,7 +1,11 @@
 ---
 
+Style:
+  - app/**/*.sccs
```
jebw, line 4: I'm guessing you're not wanting to label old internal repo files from the Source Code Control System as Style?

Maybe label `*css` and `*.scss?

Would a label of 'Design' make more sense as well? In which case I'd than move the ERB files into 'Design' also?

jebw (changes requested):
I think the file types need to be scss

.github/labeler.yml
sfawcett123: Dead right sccs is an old Unix "Source Code Control System" file,

PR #759, opened 2021-01-14
Add full stops to titles with a colour bg via CSS

sigmaben:
To match branding, adding full stops to titles in colour bgs.


peteryates (approved):
Looks good.

PR #736, opened 2021-01-06
Wide display styling

jkempster34:
### Trello card
https://trello.com/c/bBRBK28x

### Context
Update the styling to scale better on large displays

### Changes proposed in this pull request
- Limit width of main content and update hero to match
- Limit width of footer contents
- Limit width of header contents
- Move hero into `<main>` element in "My story into teaching" and "Find an event near you"

### 6000 width
image

### 6000 width
image

### 3000 width
image

### header
image

PR #761, opened 2021-01-14
Update events section to use main semantic element

jkempster34:
### Trello card
N/A

### Context
The main element is used across the rest of the site, this updates the events section to match. 

Main is also used for styling, so styles applied in [this PR](https://github.com/DFE-Digital/get-into-teaching-app/pull/736) were not applied to the events section.

- Update events section to use main semantic element


peteryates (approved):
Makes sense to me.

PR #762, opened 2021-01-15
Tweak CSS overhang and apply to feature table

ethax-ross:
- Tweak overhang and apply to feature table

Until now we were only applying an overhang to the left margin, which meant the right of the 'overhanging' box was inline with the other content. This adds the same margin to the right side so that it overhangs equally.

Applies the overhang to the `feature-table` class and also tweaks the top-margin so the feature table sits neater within the surrounding content.

Screenshot 2021-01-15 at 11 49 31


peteryates (approved):
Looks good to me

PR #764, opened 2021-01-15
Swap to loading story layouts from frontmatter

ethax-ross:
### Trello card

[Trello-728](https://trello.com/c/EolVAvlQ/728-update-stories-to-define-the-layout-in-the-content)

### Context

Remove the `StoriesController` so that story pages go through the normal `PagesController`, which applies the correct layout as defined in the story/landing/list frontmatter.

White-lists the story layouts so that they are available to the content.

- Swap to loading story layouts from frontmatter

### Guidance to review

Content PR: https://github.com/DFE-Digital/get-into-teaching-content/pull/264


app/controllers/pages_controller.rb
```diff
@@ -25,6 +29,8 @@ def cookies
   end
 
   def show
+    render_not_found && return if directly_requesting_index_page?
+
```
peteryates, line 33: As we only have `index.md` pages in these story directories, would we get the same effect by moving `returners/index.md` up a level and renaming to `returners.md`?

peteryates (approved):
Looks good overall. Left one comment, maybe a quick reorg of story content would mean we need less logic in the app. No strong feelings either way (and that might have its own cost!)

app/controllers/pages_controller.rb
ethax-ross: I'm blaming 3pm Friday for not seeing this 😬 In order to rename those files we would either need the changes to redirect `/index` renders or extra routes to account for it though, so I'll merge this down as-is, then update the content repo and raise another PR to pull out these changes.

PR #771, opened 2021-01-18
Remove temp code to render /index story lists

ethax-ross:
This code was added so that we could render the story list content at the `/section` path when it exists at both `/section.md` and `/section/index.md`. The story list content has now all been moved to be in the `/section.md` format, so this code is no longer required.


peteryates (approved):
Way nicer!

PR #773, opened 2021-01-18
Make h2/strapline font size consistent

ethax-ross:
### Trello card

[Trello-729](https://trello.com/c/Oq1fyS9J/729-make-strapline-font-sizes-consistent)

### Context

We seem to have slight variations in the font size of straplines; a strapline is 36px but often overriden by the `h2` tag style to 34px/24pt. I think it looks better at 24pt so I've tweaked them all to be consistent.

Going forward it would be nice to have a set of variables for the different font sizes, but that's a different kettle of fish/left to a separate PR.

- Make h2/strapline font size consistent

### Guidance to review

I've had a decent click about and everything looks fine as far as I can tell. The change isn't big enough that it would case an earth shattering issue if something doesn't look quite right at the smaller font size.

### DIFF
--- a/app/webpacker/styles/components/cards.scss
+++ b/app/webpacker/styles/components/cards.scss
@@ -97,7 +97,7 @@
 
   h2 {
     padding: 0 20px 20px ;
-    font-size: 36px ;
+    font-size: 24pt;
   }
 
   .card {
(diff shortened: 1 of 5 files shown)


jebw (approved):
LGTM

We should standardise the font sizes in SCSS variables at some point (and standardise the units used)

PR #774, opened 2021-01-18
Support home page as markdown

ethax-ross:
### Trello card

[Trello-730](https://trello.com/c/rIrXR9Qa/730-define-home-page-layout-in-the-frontmatter)

### Context

Currently the home page is an ERB that relies on the `PagesController` applying the correct layout.

This moves the current home template to `home-old` (to be removed in an following PR) and adds a new `home` template that supports a markdown version of the home page, which will specify the content and layout as frontmatter.

- Support home page as markdown

### Guidance to review

Follow up PR: https://github.com/DFE-Digital/get-into-teaching-content/pull/272


peteryates (approved):
Looks good

PR #779, opened 2021-01-18
Fix new home layout partial rendering

ethax-ross:
I misread the original home layout thinking the set of partials rendered was the `content/home/_*` partials, but it wasn't. This commit re-instates missing partials and moves the home content partial rendering just below the `yield`, which will keep them in the same location as the original home layout.


peteryates (approved):
Looks good

PR #781, opened 2021-01-18
Added Grouped Cards component

jebw:
### Trello card

https://trello.com/c/xa8Eotqv

### Context

The website should provide an indexed list of assessment providers. For this we need a grouped cards listing of all assessment providers

1. Allow full width markdown pages
2. Add a GroupedCards::Listing component
3. Add a GroupedCards::Card component
4. Small CSS tweak to allow breaking long words (eg email addresses) in the cards

### Guidance to review

1. The event-box CSS should probably be split out into something more general but I'll do that in a follow up PR rather than overcomplicate this one

### Preview

Screenshot from 2021-01-18 16-32-21

### DIFF
--- a/app/components/grouped_cards/card_component.rb
+++ b/app/components/grouped_cards/card_component.rb
@@ -1,0 +1,21 @@
+module GroupedCards
+  class CardComponent < ViewComponent::Base
+    attr_reader :data
+
+    def initialize(data)
+      @data = data
+    end
+
+    def header
+      @data["header"]
+    end
+
+    def fields
+      @data.without("header")
+    end
+
+    def linked_value(value)
+      Rinku.auto_link(h(value)).html_safe
+    end
+  end
+end
--- a/app/components/grouped_cards/listing_component.rb
+++ b/app/components/grouped_cards/listing_component.rb
@@ -8,8 +8,8 @@ module GroupedCards
       @data.keys
     end
 
-    def region_link_anchor(region)
-      region.downcase.gsub %r{[^a-z0-9]+}, "-"
+    def group_link_anchor(group)
+      "group--" + group.parameterize
     end
 
     def providers(region)
--- a/app/components/grouped_cards/listing_component.rb
+++ b/app/components/grouped_cards/listing_component.rb
@@ -1,0 +1,19 @@
+module GroupedCards
+  class ListingComponent < ViewComponent::Base
+    def initialize(data)
+      @data = data
+    end
+
+    def regions
+      @data.keys
+    end
+
+    def region_link_anchor(region)
+      region.downcase.gsub %r{[^a-z0-9]+}, "-"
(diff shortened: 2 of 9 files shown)


app/components/grouped_cards/listing_component.rb
```diff
@@ -0,0 +1,19 @@
...
+    def initialize(data)
+      @data = data
+    end
+
+    def regions
+      @data.keys
+    end
+
+    def region_link_anchor(region)
+      region.downcase.gsub %r{[^a-z0-9]+}, "-"
```
ethax-ross, line 12: Would `parameterize` work here instead of a regex? you'd get a slightly different result in that it omits symbols, but I'm not sure if that would matter?

```
region.parameterize(separator: "-")
```

app/components/grouped_cards/card_component.rb
```diff
@@ -0,0 +1,21 @@
...
+    def header
+      @data["header"]
+    end
+
+    def fields
+      @data.without("header")
+    end
+
+    def linked_value(value)
+      Rinku.auto_link(h(value)).html_safe
```
ethax-ross, line 18: We have a helper you could use instead of `html_safe`, but since you're escaping anyway take it or leave it:

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/helpers/text_formatting_helper.rb#L10

ethax-ross (approved):
Looks good, just a couple of take it or leave it comments. Agree we would want to split the event-box CSS out/rename it for it to make more sense here.

app/components/grouped_cards/listing_component.rb
jebw: Hah - good call, I'd forgotten about that one - I'll see if it behaves as I want

app/components/grouped_cards/card_component.rb
jebw: Thanks, thats got different behaviour though so I'll leave it as is - that would still allow HTML through which I'm not wanting to - Rinku doesnt handle escaping well - hence the above approach (I really don't like using `html_safe` though)

ethax-ross (dismissed):

app/webpacker/styles/layout.scss
```diff
@@ -77,7 +77,7 @@ main.semantic {
     margin: 1em auto 4em;
   }
 
-  > article {
+  > article:not(.fullwidth) {
```
peteryates, line 80: I think explicitly stating how wide a full width article should be might be clearer here

```suggestion
  > article {
    &.fullwidth {
      flex: 1 0 100%;
    }
  }
```

peteryates (approved):
Looks good overall, just left a comment regarding the `fullwidth` selector.

peteryates (approved):
Looks good

PR #782, opened 2021-01-18
Hero CSS refactor (switching to Grid)

peteryates:
### Trello card

Not directly work on the [new campaign page](https://trello.com/c/TKmHrvze/696-development-mock-up-potential-modifications-to-the-campaign-landing-page-based-on-heat-maps-eg-upfront-info-and-ctas-for-mobile) but laying the groundwork.

### Context

The original hero layout was functional but brittle, making changes to the content would cause it to render incorrectly and [make some text illegible](https://user-images.githubusercontent.com/128088/104938740-6131db80-59a7-11eb-8793-43f75811e0c7.png).

This change is necessary because the upcoming new campaign page contains a list of navigation links beneath the main page heading that are taller than the subtitle and break the layout.

Switching to Grid should make this transition easier as the height of the hero can be governed by the content and it should automatically stretch to accommodate longer content.

Screenshot from 2021-01-18 16-18-47

Rewrite the functionality using [CSS grid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout), the only layout option that gives us the ability to explicitly have certain elements overlap others.

#### The plan

This is the rough sketch I started off with, the idea is to use [`grid-area`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-area) to place the image, title and subtitle on the page in pre-defined positions on a grid.

hero-grid

#### Hurdles

Grid is not supported by IE so we need a fallback. The fallback will make some compromises to ease development and keep the code simple:

* the header won't overlap the hero image
* no hero image will be displayed for mobile IE users (if we have any I'll write them a personal letter of condolence!)

### Guidance to review

| Browser | Wide | Medium | Narrow |
| -------- | -------- | -------- | ------- |
| **Firefox** | Screenshot from 2021-01-18 16-16-32 | Screenshot from 2021-01-18 16-16-37 | Screenshot from 2021-01-18 16-16-44 |
| **Chrome** | Screenshot from 2021-01-18 16-21-29 | Screenshot from 2021-01-18 16-21-36 | Screenshot from 2021-01-18 16-21-48 |

**IE and Safari screenshots to follow**.

https://user-images.githubusercontent.com/128088/104940547-cbe41680-59a9-11eb-9148-4343c89b24f4.mp4


app/components/sections/hero_component.html.erb
```diff
@@ -1,17 +1,22 @@
...
+    <%= responsive_image %>
+    <div class="hero__title">
+      <h1><%= tag.span(title) %></h1>
+    </div>
+      <% if subtitle.present? %>
+        <div class="hero__subtitle">
+          <%= tag.div(subtitle) %>
+        </div>
+      <% end %>
+      <% if false %>
```
ethax-ross, line 11: Did you mean to leave this in?

app/webpacker/styles/sections/hero.scss
```diff
@@ -1,177 +1,222 @@
...
-    $mobile-cutoff: 800px;
+// styles
+.hero {
+  background-color: $grey;
 
-    display: flex;
-    align-items: center;
-    background-color: $green;
-    margin-top: 50px;
+  // repeat( auto-fit, minmax(250px, 1fr) );
```
ethax-ross, line 111: Can we delete this?

ethax-ross (commented):
💯 Nice work! couple of queries around what I think may be temp/test code left in. The only thing I noticed when I was testing it locally was if I enabled content and a subtitle they overlapped. Not sure if that's a problem or not; it looks like the content part isn't all the way there yet as there's no way to specify it in the component interface (I just uncommented the bit of HTML in the component template) so I imagine that will just follow when we have a use case for it?

app/components/sections/hero_component.html.erb
peteryates: Ah no, was experimenting with sample content, will remove.

peteryates:
Thanks for having a look over it.

Currently there aren't any plans to have pages with both a subtitle and content, so they occupy exactly the same place on the grid.

```scss
  &__subtitle,
  &__content { grid-area: 4 / 1 / 5 / 4; }
```

If we were to go down that route it would be doable but we'd need some logic in the component that add classes (e.g., `.hero__content--middle`, `.hero__content--bottom`) to place them one above the other. I didn't implement that for the time being but it should be quite straightforward.

ethax-ross (approved):
Looks great 👍

PR #785, opened 2021-01-19
Standardise media queries throughout

ethax-ross:
### Trello card

[Trello-732](https://trello.com/c/VEwWzqcT/732-standardise-use-of-media-queries-thoughout)

### Context

We are redefining our media queries in lots of different places, hardcoding the breakpoints. To make it more maintainable and consistent we want to have variables for each of the standard/most common breakpoints and only hard code special cases.

This PR uses the Mq library for managing the media queries and breakpoints (which is what GOV.UK uses under the hood) and sets out the standard breakpoints of:

```
mobile: 500px
tablet: 800px
desktop: 1200px
wide: 2000px
```

With a little shoe-boxing (all of which I manually checked to make sure the breakpoint tweaks work well) this covers all of the current media queries apart from the navigation bar (900px).

- Add Mq lib for media queries
- Replace 800px media queries with tablet variable
- Replace 2000px media queries with wide variable
- Replace 500px media queries with mobile variable
- Remove unused media query
- Shoebox remaining media queries
- Remove sections/header.scss

### Guidance to review

It would be worth someone else having a good click about in the site. We may want to redefine the breakpoints in the future to be a bit more standardised/common, but for now this hits our use cases with minimal impact to how the site will behave/display.


peteryates:
So much cleaner! 🤩 

Will have a closer look shortly

peteryates (approved):
Works well for me, clicked around all the page types and everything looked fine.

PR #786, opened 2021-01-19
Adding alt attributes for accessibility

peteryates:
### Context

The recent accessibility report showed that we're missing alt attributes in a few places.

Add a generic alt tag to the video player and allow story card photo thumbnails to have their alt tags set via the new frontmatter key `image_description`.

### Guidance to review

Does this look ok and make sense? We could make this nicer in the future by getting rid of the list of card attrs in the index pages and using the global frontmatter and auto-generating the descriptions from their name/position. Some more work (ie to disable/unpublish without deleting) would need to be done to support that though.

### DIFF
--- a/app/components/card_component.rb
+++ b/app/components/card_component.rb
@@ -6,14 +6,15 @@ class CardComponent < ViewComponent::Base
   def initialize(card:)
     @card = card
 
-    @snippet   = card["snippet"]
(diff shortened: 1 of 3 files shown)


ethax-ross (dismissed):
Looks good to me

ethax-ross (approved):

PR #787, opened 2021-01-19
Clean up the page helpful partial

ethax-ross:
### Trello card

[Trello-739](https://trello.com/c/LiFR6iZk/739-clean-up-page-question-css)

### Context

The "Is this page helpful?" section that appears at the bottom of most pages is showing a bit misaligned. In addition the CSS could be simplified and the component itself renamed to be clearer.

The `page_question__answer` class needs to remain as-is given GTM targets this.

- Clean up page_helpful component
- Rename page_question to page_helpful

### Guidance to review

I've manually tested this in each of the different layouts its used in.

| Before      | After       |
| ----------- | ----------- |
| Screenshot 2021-01-19 at 15 28 59      |    Screenshot 2021-01-19 at 15 28 48   |

### DIFF
--- a/app/views/events/show.html.erb
+++ b/app/views/events/show.html.erb
@@ -93,5 +93,7 @@
         <div data-controller="scribble" data-scribble-id="<%= @event.scribble_id %>"></div>
       <% end %>
 
-      <%= render "sections/page_question" %>
+    <div class="content__left">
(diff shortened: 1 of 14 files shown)


jebw (approved):
LGTM

PR #788, opened 2021-01-20
Fix hero max-width

jkempster34:
### Context
Currently, the home page hero sits inside the `<main>` section, so is subject to it's max-width; the hero on the other pages sits above `<main>` so is not. 

### Guidance to review
Will this work as a quick fix until the home page is updated?


peteryates (approved):
👍🏽

PR #792, opened 2021-01-21
Clean up the application layout

ethax-ross:
### Trello card

[Trello-741](https://trello.com/c/oVfMQhYF/741-refactor-the-application-layout)

### Context

Transition the application layout to be more consistent with the content layout, which uses a cleaner structure/CSS.

Update all the views that use the application layout so that they are rendering correctly.

Add a number of missing page titles I noticed when I was auditing all the views.

- Clean up the application layout

### Guidance to review

Views changed:

- Cookie preferences
- Cookies
- Privacy policy
- Error pages
- Event show
- Event sign up
- Mailing list sign up

Still to do:

- Events index
- Event category + archive

Would it be better to define the `@fullwidth`/`@no_hero` etc in the controller as a `before_action: -> { @fullwidth = true }` or leave it in the views?

We have a couple of obscure heading sizes on the event and error templates I had to account for to keep consistent.


peteryates (dismissed):
Had a good click around and everything looks good to me. Nice.

peteryates (approved):
Looks good

PR #797, opened 2021-01-22
Make hero nav links encapsulate arrows

peteryates:
Minor CSS tweak to make the link include the down-pointing arrows.

This should make the link targets bigger and more intuitive for mobile users.

Screenshot from 2021-01-22 13-57-39


MylesJarvis (approved):
You've shown me a screen cap of this - performs as expected.

PR #804, opened 2021-01-25
Transition home page to semantic layout

ethax-ross:
Enables the home page to migrate to a semantic layout, matching the other layouts by replacing `home` (leaving `home_old` as the old layout which is still in use for now).

Adds a new CSS component `pagewidth-bg` to display a page-wide grey/yellow background that is used on both the home page and search for events page (the event search page is updated in this PR).

A few additional CSS classes have been added so that the new changes can co-exist with the existing home page structure while this is rolled out. We should be able to clean a lot of it up when the old layout is removed.

Content PR to follow this with: https://github.com/DFE-Digital/get-into-teaching-content/pull/286

| Home Page (old layout still)      | Evens Page (search component updated) |
| ----------- | ----------- |
| screencapture-0-0-0-0-3000-2021-01-25-10_08_39   | screencapture-0-0-0-0-3000-events-2021-01-25-10_08_18       |

### DIFF
--- a/app/components/events/search_component.html.erb
+++ b/app/components/events/search_component.html.erb
@@ -1,8 +1,8 @@
-<div class="pagewidth-bg pagewidth-bg--grey pagewidth-bg--with-title">
+<div class="colored-stripe colored-stripe--grey colored-stripe--with-title">
   <section class="container">
     <article class="markdown overhang fullwidth">
-      <h2 id="searchforevents" class="pagewidth-bg__title purple"><%= heading %></h2>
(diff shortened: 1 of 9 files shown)


peteryates (dismissed):
Looks great. My one reservation is the name `.pagewidth-bg` - trying to think of a more descriptive name. `.coloured-stripe` or something? 🤔

ethax-ross:
Yeah I flip flopped on the name a bit, `colored-stripe` sounds better I'll go with that.

peteryates (approved):

PR #805, opened 2021-01-25
Standardise font sizes

jkempster34:
### Trello card
https://trello.com/c/IGJu2Apb

### Context
Standardise font sizes. This also improves accessibility because users can change the font size in their browsers.

- Update font size to use rem constants
- Remove unused CSS

### Guidance to review
I have tried to reflect the current font sizes exactly. 
Removed some small bits of CSS that aren't used.
Also had a browse with font sizes set to extra-large and all seems fine. Image below shows after and before with browser font size set to extra large.
image


app/webpacker/styles/content.scss
```diff
@@ -18,15 +18,9 @@
         vertical-align: top;
         width: 66.6666%;
         float: left;
```
jkempster34, line 20: Removed CSS here

app/webpacker/styles/components/event-box.scss
```diff
@@ -1,16 +1,3 @@
-.event-link {
```
jkempster34, line 1: And here

app/webpacker/styles/components/event-box.scss
```diff
@@ -1,16 +1,3 @@
...
-    &:focus {
-        outline: none;
```
jkempster34, line 3: Removed CSS here

app/webpacker/styles/content.scss
```diff
@@ -18,15 +18,9 @@
...
-        .jumbotron {
-            font-size: 19px;
```
jkempster34, line 22: And here

app/webpacker/styles/components/event-box.scss
```diff
@@ -1,16 +1,3 @@
...
-        border: none;
-        background: transparent;
-        box-shadow: none;
-    }
-}
-
-.events-listing p {
-  font-size: 300px;
```
jkempster34, line 11: And here

peteryates (approved):
Looks good to me, had a look around the site and everything feels like it's working well.

We have an awful lot of font sizes, I think a good follow up might be to try to reduce the number, but this is a huge step in laying the groundwork for that 👍🏽

PR #811, opened 2021-01-26
Remove old/unused home page CSS

ethax-ross:
This was either unused or is now redundant with the new home page layout in use.


peteryates (approved):
👋🏽

PR #812, opened 2021-01-26
Replace occurrences of .container-1000 class

ethax-ross:
### Trello card

[Trello-757](https://trello.com/c/SbZOdKxq/757-remove-unused-css-content)

### Context

We still use the `.container-1000` class in a few places, which is associated with the old `.content` layout. We want to replace these with the new layout `section.container` so that we can kill off all the old `.content` related CSS.

The `section.container` class has been moved outside of the `main.markdown` class so that we can use it in the hero component and footer sections. We probably need to go one step further and decouple `.markdown` from `main` so that we can utilise the `.overhang` CSS in these areas as well, but I've not done that yet in an attempt to keep this PR manageable.

A few CSS selectors from the `.content` class had to be shifted into the event/mailing list component CSS as we seem to use custom margins and font sizes here for headings. We should look to clean this up further going forward. 

- Remove section.container outwith main.semantic
- Replace container-1000 in mailing list CTA
- Replace container-1000 in feedback-bar
- Replace container-1000 in talk-to-us section
- Replace container-1000 in mailing list sign up
- Replace container-1000 in event sign up

### Guidance to review

This PR will be followed up by another that removes the now unused CSS.

Pages/components effected:

- Mailing list CTA (home page)
- Feedback bar
- Talk to us section
- Mailing list sign up
- Event sign up

### DIFF
--- a/app/components/sections/hero_component.html.erb
+++ b/app/components/sections/hero_component.html.erb
@@ -17,7 +17,7 @@
 <% end %>
 
 <% if show_mailing_list %>
-  <div class="container-1000">
+  <section class="container">
     <div class="hero__mailing-strip">
         <div class="hero__mailing-strip__text">
             <p>Sign up to receive information via email that could help you make your decision to get into teaching.</p>
@@ -28,5 +28,5 @@
           <% end %>
         </div>
     </div>
-  </div>
+  </section>
 <% end %>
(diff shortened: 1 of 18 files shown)


peteryates (approved):
All looks good to me

PR #814, opened 2021-01-26
Switch cookies page to markdown

peteryates:
### Trello ticket

https://trello.com/c/H3fvGBTB/755-rework-the-tables-on-the-cookies-page

### Context and changes

Replace the cookies HTML page with a markdown document. To accommodate the change a 'disclaimers' layout was created that has some subtle styling differences to the regular content layout.

The main difference is that the inline content is artificially limited to 65% width on bigger screens but the tables are allowed to stretch to 100%, allowing the dense information to be displayed in an uncramped manner.

Screenshot from 2021-01-26 13-56-10

The table text has also been styled so that it resizes more elegantly on mobile (old on the left, new on the right):

Screenshot from 2021-01-26 13-39-26

Finally, some minor content tweaks were made to remove repetition from the tables, making them more digestible.

Thoughts @MylesJarvis?


ethax-ross (dismissed):
Looks good code-wise 👍

ethax-ross (approved):

PR #824, opened 2021-01-27
Update _name.html.erb

MylesJarvis:
### Context
Looking to optimise the flow for the Mailing List, which includes tightening the copy on the page.
Renaming the H1 header to reflect the CTA we use on the homepage.
Removing copy in the intro about updating details as we cover this in the confirmation page.

### DIFF
--- a/app/controllers/mailing_list/steps_controller.rb
+++ b/app/controllers/mailing_list/steps_controller.rb
@@ -22,7 +22,7 @@ module MailingList
     end
 
     def set_step_page_title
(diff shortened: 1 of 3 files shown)


peteryates (approved):
Looks good.

What about "advice _on_ getting into teaching"?

peteryates:
This has hit a similar problem to last time @MylesJarvis, some tests refer to the heading text. I'll finish this off 👍🏽

ethax-ross (approved):

PR #826, opened 2021-01-28
Accessibility fixes

peteryates:
## Context

Some issues were highlighted by an accessibility review that was conducted using the [SiteImprove](https://siteimprove.com/) browser extension.

#### 1- Text Alternatives: 
* [x] Image with no alt attribute

Caused by injected advertising/tracking.

Screenshot from 2021-01-28 13-00-46

#### Adaptable:
* [x] Input field has no description - caused by the SnapChat injection, see below
* [x] **Bold** tag used to format text - 030a22b
* [x] HTML is used to format content - 4d56968 

#### Distinguishable:

* [x] Color contrast is insufficient

Unsure why, but SiteImprove is reporting the text as white when it's clearly black.

| SiteImprove | Contrast checking |
| -------- | --------- |
| Screenshot from 2021-01-28 12-23-25 | Screenshot from 2021-01-28 12-22-27 |

#### Link Purpose (In Context): 

* [x] Link text used for multiple different destinations - 75974efd54a2

This was caused by the 'Home' link in the top nav bar linking to `/` and the one in the footer linking to `/home`. They've both been change to link to the `root_path`.

#### Predictable, Labels or Instructions, and Input Assistance

These issues all seem to be caused by SnapChat injecting a poorly marked-up form into our page.

* [x] Missing button in form
* [x] Input field has no description
* [x] Input field has no description
* [x] Frame is missing a title
* [x] Form is missing button

Screenshot from 2021-01-28 12-27-19

#### Adaptable:
* [x] Input field has no description - SnapChat (same as above)
* [x] Element ID is not unique - df638542d725
* [x] HTML is used to format content - 4d56968 (same as above)

#### Navigable:

* [x] Link text used for multiple different destinations - 75974efd54a2 (same as above)

#### Predictable:

* [x]  Missing button in form - SnapChat (same as above)

#### Input Assistance

* [x] Input field has no description - SnapChat (same as above)

#### Compatible

* [x] iFrame is missing a title - 9151c581ca

#### Adaptable:
* [x] **Bold** tag used to format text - 030a22b (as above)
* [x] HTML is used to format content - 4d56968 

#### Distinguishable:
* [x] Color contrast is insufficient - (as above, looks like a false positive)

#### Navigable:

* [x] Link text used for multiple different destinations - 75974efd54a2 (same as above)

#### Compatible:

* [x] iFrame is missing a title - 9151c581ca

## Other changes

* [x] Update the CTA headings so they're rendered in a `span` (with bold text) instead of a `h3`/`h4`

Screenshot from 2021-01-28 15-41-48

* [x] Make the mobile nav bar work like the desktop one, remove links to the current page and style the active page accordingly
Screenshot from 2021-01-28 15-33-53

### DIFF
--- a/app/components/navbar/navbar_component.rb
+++ b/app/components/navbar/navbar_component.rb
@@ -10,7 +10,7 @@ module Navbar
       class_name = "active" if first_uri_segment_matches_link?(link_path)
 
       content_tag(:li, class: class_name) do
-        link_to link_text, link_path
+        link_to_unless_current link_text, link_path
       end
     end
 
--- a/app/components/calls_to_action/chat_online_component.html.erb
+++ b/app/components/calls_to_action/chat_online_component.html.erb
@@ -2,9 +2,7 @@
   <%= icon %>
 
   <div class="call-to-action__content">
-    <h4 class="call-to-action__heading">
-      Chat to us
-    </h4>
+    <%= tag.span("Chat to us", class: "call-to-action__heading") %>
 
     <p class="call-to-action__text">
       If you're unsure whether your qualifications are equivalent, you can chat to us.
--- a/app/components/calls_to_action/multiple_buttons_component.html.erb
+++ b/app/components/calls_to_action/multiple_buttons_component.html.erb
@@ -2,9 +2,7 @@
   <%= icon %>
 
   <div class="call-to-action__content">
-    <h4 class="call-to-action__heading">
-      <%= title %>
-    </h4>
+    <%= tag.span(title, class: "call-to-action__heading") %>
 
     <div class="call-to-action__button-row">
       <% links.each do |link| %>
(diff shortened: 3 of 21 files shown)


app/components/navbar/navbar_component.rb
```diff
@@ -10,7 +10,7 @@ def nav_link(link_text, link_path)
       class_name = "active" if first_uri_segment_matches_link?(link_path)
 
       content_tag(:li, class: class_name) do
-        link_to link_text, link_path
+        link_to_unless_current link_text, link_path
```
ethax-ross, line 13: Do we need to replicate this in the mobile nav bar?

ethax-ross (dismissed):
Looks good

app/components/navbar/navbar_component.rb
peteryates: Ah didn't realise the mobile one wasn't using the `nav_link` method, will change it and make sure it looks ok

ethax-ross (approved):

PR #827, opened 2021-01-28
Search page - phase1 noJS

jebw:
### Trello card

https://trello.com/c/1VGYGEd3

### Context

Our users w wish to use a search function to find pages within our site. This story will be delivered in 2 PRs

The first adds the non-js version of the search, the second will add an inline autocompleting JS version building off the first version.

1. Reworked the NavBar to remove some of the fixed positioning and allow adding additional elements, icons - note this is still 2 nav bars for desktop/mobile but it should now be possible to merge them in the future.
2. Added Search page
3. Linked search page from Navbar

### Guidance to review

I've had to add a fixed height hack back in for IE for the NavBar because it cannot support the variable height (bugs in Flexbox implementation). This should be less of an issue on IE since its primarily a desktop browser so likely to be 1024px or wider

Screenshot from 2021-01-28 13-55-20

Screenshot from 2021-01-28 13-56-03


app/components/navbar/navbar_component.html.erb
```diff
@@ -12,14 +8,19 @@
             <% end %>
             <%= nav_link("Find an event near you", events_path) %>
         </ul>
+
+        <div class="navbar__desktop__search">
+          <%= link_to image_pack_tag("icon-site-search.svg"), search_path unless Rails.env.production? %>
+        </div>
```
peteryates, line 14: Should the `unless` statement wrap the whole `div` here? Not sure it will matter much.

jebw: Yeh - you're right, that was a last minute change - I'll update it

peteryates (approved):

PR #828, opened 2021-01-28
Consolidate main page header CSS

ethax-ross:
### Trello card

[Trello-784](https://trello.com/c/YDFJTUGD/784-kill-off-all-the-strapline-classes-in-favour-of-the-markdown-h2-styling)

### Context

We currently have two ways of styling h2 headings; by `.markdown > h2` and a `.strapline` class (legacy and for places where a wrapping div prevents the `.markdown > h2` selector applying).

This commit removes all uses of the `.strapline` class by bubbling the `h2` up a level where possible so that the `.markdown > h2` selector applies.

There are a couple of use cases outside the `.markdown` class, such as `.feature` and `footer`. To cover these the heading CSS has been abstracted into a mixin and mixed in to the particular classes.

Also adds a global right-margin on the headings when in mobile, as per the design (apart from on the feature tables where the right margin should always be present).

- Remove strapline class

### Guidance to review

I think this could be taken further in terms of formalising the different non-markdown content throughout the site, but this should be a step in the right direction.

### DIFF
--- a/app/components/content/feature_table_component.html.erb
+++ b/app/components/content/feature_table_component.html.erb
@@ -1,7 +1,8 @@
+<% if title? %>
+  <h2 class="always-indent"><%= title %></h2>
+<% end %>
+
 <div class="<%= wrapper_class %>">
-  <% if title? %>
-    <h2 class="strapline strapline--blue"><%= title %></h2>
-  <% end %>
   <dl>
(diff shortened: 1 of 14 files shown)


peteryates (approved):
Nice, a definite improvement.

PR #831, opened 2021-01-29
Replace 1.5rem/20px margins/padding with $indent-amount

ethax-ross:
- Replace 1.5rem margins/padding with $indent-amount

The `h2` have their own top margin now that the `.strapline` class was removed
in an earlier PR. We no longer need such a large bottom margin.
We have a standard `$indent-amount` already defined as 1.5rem, so it makes sense to reuse this where possible in case we want to tweak it in the future.

Also replaces 20px margin/padding values with `$indent-amount` as there is only 4px difference and it makes things align more nicely across the board.

- Tweak content-alert bottom margin

The `h2` have their own top margin now that the `.strapline` class was removed in an earlier PR. We no longer need such a large bottom margin.

### DIFF
--- a/app/webpacker/styles/accordion.scss
+++ b/app/webpacker/styles/accordion.scss
@@ -42,12 +42,12 @@
         }
 
         .call-to-action {
-          margin: 0 1.5rem;
+          margin: 0 $indent-amount;
         }
 
         .step-content__description {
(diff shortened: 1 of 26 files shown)


peteryates (approved):
Looks good.

PR #833, opened 2021-01-29
added css to reduce title size on cta in sidebar

sigmaben:
The title size it too large currently for the sidebar, this reduces it in the sidebar only.


peteryates:
In addition to the above description enhance the `CallsToAction::SimpleComponent` (and `CallsToAction::NarrowComponent`, which is a descendant of it) by adding arguments that control some extra classes; `hide-on-mobile`, `hide-on-tablet`, and `hide-on-desktop`.

These line up with the breakpoints we use elsewhere and the utility CSS classes use `mq` to toggle the visibility of the element.


https://user-images.githubusercontent.com/128088/106296775-fbb4d900-6249-11eb-9704-8d2a9b250e47.mp4

ethax-ross (approved):

PR #841, opened 2021-02-01
Fixed card images showing as too wide on mobile

jebw:
### Trello card

https://trello.com/c/isashLvM

### Context

Cards with headers are showing images too wide on mobile

1. Tweaked CSS so solve cards being too wide when on mobile

### Guidance to review

1. Is the change of box sizing on `.card` likely to cause an issue elsewhere.

`content-box` seems to be being used by default - most layout frameworks seem to have concluded `border-box` leads to better behaviour because 100% actually is 100% unlike with `content-box`

**Before:**

Screenshot from 2021-02-01 14-28-56

**After**

Screenshot from 2021-02-01 14-28-20

### DIFF
--- a/app/webpacker/styles/components/cards.scss
+++ b/app/webpacker/styles/components/cards.scss
@@ -9,6 +9,7 @@
         flex-direction: column;
         text-decoration: none;
         position: relative;
+        box-sizing: border-box;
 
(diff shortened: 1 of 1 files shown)


peteryates (approved):
Makes sense.

PR #847, opened 2021-02-02
Fix incorrect heading indentation

jkempster34:
### Trello card
https://trello.com/c/Csz3XTbz

### Context
- Heading indentation and padding are not correct in event sign up. 

- Fix event sign up heading indentaion/padding

### Guidance to review
image

### DIFF
--- a/app/views/event_steps/_form.html.erb
+++ b/app/views/event_steps/_form.html.erb
@@ -4,7 +4,7 @@
   </p>
 
   <h1>Sign up for this event:</h1>
-  <h2><%= event.name %></h2>
+  <h2 id="event-reg-main-title"><%= event.name %></h2>
 
   <%= govuk_form_for current_step, url: step_path do |f| %>
     <%= f.govuk_error_summary %>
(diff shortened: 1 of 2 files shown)


jebw (changes requested):
This feels like a very specific solution - why is this displaying offset for events sign up but not other for fields?

Does this really only affect those fields or should it be getting solved more generically?

jkempster34:
> This feels like a very specific solution - why is this displaying offset for events sign up but not other for fields?
> 
> Does this really only affect those fields or should it be getting solved more generically?

I can remove the `content-heading` mixin from the h2, which fixes the indentation without affecting the other targeted class `mailing-reg-main`. 

But, this [(link)](https://github.com/DFE-Digital/get-into-teaching-app/blob/012beed49f5f666229ead5f1639d21db9fb11a10/app/webpacker/styles/markdown.scss#L31-L36) for the top margin is used quite a lot, so the only solution I can think of is to override it?

peteryates:
The link you referenced above should only apply to the blue h2's that overhang the content; the rule targets only those that appear within `.markdown` elements (ours are all `article.markdown`).

I think we should remove all horizontal padding/margins elsewhere from headers, I can't think of anywhere else (apart from the hero) where we'd want it.

PR #854, opened 2021-02-04
Update API credentials

ethax-ross:
Update the API credentials to switch the GiT service from using the shared secret to its own API client key.


jebw (approved):
LGTM

PR #855, opened 2021-02-04
Change event_categories routes to event-categories

peteryates:
This is a cosmetic change to make event routes look a bit nicer.

Now, instead of `"/event_categories/some-event-category` the route will be `/event-categories/some-event-category`

The specs are requests specs rather than routing specs because routing specs haven't supported redirects for quite some time (link to 10 year old RSpec issue in commit message).

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -41,15 +41,17 @@ Rails.application.routes.draw do
     end
   end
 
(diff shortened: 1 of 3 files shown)


ethax-ross (approved):
Looks good

PR #864, opened 2021-02-05
Homepage refresh

peteryates:
### Trello card

https://trello.com/c/hkyt13A7/1116-pete-to-code-up-update-to-homepage-following-the-homepage-refinement-session

### Context and changes

The homepage was due a refresh and the new design is based on lots of research and user data.

Screenshot_2021-02-09 Get into teaching Inspire the next generation(1)

#### Steps to become a teacher (the purple box)

~~The original design had the SVG overlapping the right of the purple box. This was implemented using a grid but dropped in favour of a simpler approach for the time being. Using grid would give the desired result at the expense of a complicated layout. The newer, simpler version is based on Flexbox and is fully responsive.~~

The 'purple box' is a large CTA that appears at the top of the homepage. In the new design it contains a header, some text, a call to action button and a SVG graphic.

The complexity in this compenent lies in the fact that it overhangs the right of the purple box; this means we need to wrap the regular container in an outer container (which includes all content *plus* the overhanging part of the SVG).

As a result of this we can't use just go down the basic route of a pair of nested flexbox elements; the requirement for part of one element to overlap another requires either grid or hacks with margins.

This approach uses grid; the overhang amount is controlled vertically and horizontally via the variables `$px-either-side-of-content` and `$px-above-content` and the remainder of the grid should automatically fit the content. `$px-either-side-of-content` is also used to constrain the other homepage content accordingly, so it lines up with the purple box.

Fallback (no grid mode 😢 )| Desktop | Tablet | Mobile | Grid |
| ------- | ------- | ------- | ------- | ------ |
|  Screenshot from 2021-02-08 12-44-02 | Screenshot from 2021-02-08 12-44-19 |  Screenshot from 2021-02-08 12-44-30 | Screenshot from 2021-02-08 12-50-28 | Screenshot from 2021-02-05 13-35-41 |

#### Calls to action

| Desktop | Mobile |
| ------ | ------- |
| Screenshot from 2021-02-09 11-54-07 | Screenshot from 2021-02-09 11-27-56 |

#### Directory

The 'directory' is an alternative set of navigation links. Unlike the primary nav it's categorised and this should make it easer for people to find the exact content they're after. Awaiting icons so there's emoji in their place for the time being 👨🏽‍🏫 

| Desktop | Mobile |
| ------- | ------ |
| Screenshot from 2021-02-09 11-28-16 | Screenshot from 2021-02-09 11-28-24 |

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -4,5 +4,6 @@ module.exports = {
     'max-nesting-depth': null,
     'selector-no-qualifying-type': [true, { ignore: ['class'] }],
     'selector-max-id': 1,
+    'declaration-block-single-line-max-declarations': 1,
   },
 };
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -5,5 +5,6 @@ module.exports = {
     'selector-no-qualifying-type': [true, { ignore: ['class'] }],
     'selector-max-id': 1,
     'declaration-block-single-line-max-declarations': 1,
+    'declaration-no-important': false,
   },
 };
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -5,6 +5,5 @@ module.exports = {
     'selector-no-qualifying-type': [true, { ignore: ['class'] }],
     'selector-max-id': 1,
     'declaration-block-single-line-max-declarations': 1,
-    'declaration-no-important': false,
   },
 };
--- a/app/components/calls_to_action/homepage_component.html.erb
+++ b/app/components/calls_to_action/homepage_component.html.erb
@@ -1,0 +1,12 @@
+<%= tag.div(class: %w[call-to-action call-to-action--homepage]) do %>
+  <div class="call-to-action__content">
+    <%= tag.div(icon, class: "call-to-action__icon__container") %>
+
+    <%= tag.span(title, class: "call-to-action__heading") %>
+
+    <div class="call-to-action__action">
+      <%= link %>
+    </div>
+  </div>
+  <%= image %>
+<% end %>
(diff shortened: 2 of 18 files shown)


peteryates:
Looks decent on Safari, just had to make a little fix to account for the current version's lack of `gap` support.

Screenshot 2021-02-09 at 13 24 48

peteryates:
Mostly looks fine on IE11 too, a hint of weirdness with the button under the subtitle. Can't work out why it's not working :/

Screenshot from 2021-02-09 13-52-54

peteryates:
Now with icons!

Screenshot from 2021-02-09 16-07-54

sigmaben (changes requested):
Steps block changes:

- [x] Increase title size to equivalent of 36px with 45px line-height.
- [x] Make button not go full width (perhaps inline-block)
- [x] Make graphic fill the right-hand side more (background-size: contain)
- [x] Change grid layout slightly to allow more space for text content (grid-template-columns: 15px 1.2fr 1.2fr 0.8fr 0.8fr 15px; or similar)
- [x] On mobile match the design in the xd dev view (https://xd.adobe.com/view/c5325e7d-199f-4dbe-a8f4-52f979202f20-9061/specs/)

ethax-ross (dismissed):
Looks great, some grid wizardry 🧙

ethax-ross (approved):

sigmaben (changes requested):
The last couple of issues I've found - edge cases, both on wide mobile / narrow tableish views. 

1. CTA title is centre aligned, should be left-aligned. 
2. button arrow is hanging on its own, plus no padding at the bottom of the block.
3. On mobile / small screens reduce font size from 28 to 24 on CTA block titles. 
home-snag-1
home-snag-2

sigmaben (approved):

PR #872, opened 2021-02-08
Update Provider Event Link in Events Homepage

MylesJarvis:
Correcting link to provider form, which appears to have been updated.


peteryates (approved):
Looks good, link works. Didn't realise DfE had a forms service 😮

PR #874, opened 2021-02-08
Added javascript search

jebw:
### Trello card

https://trello.com/c/1VGYGEd3

### Context

We want to show the users a nice dropdown autocomplete search box when possible.

1. Added stimulus controller
2. Rejigged the Navbar some more for this.

### Guidance to review
I'm using the SJR approach to keep the look of the results consistent across JS and NoJS searches - should mean if a more complex search mechanism is added nothing needs to be done for the results.

Note: The results are not showing the highlighted result text colour in white - haven't got to the bottom of this one yet but its off in prod anyway.

### DIFF
--- a/app/components/navbar/navbar_component.html.erb
+++ b/app/components/navbar/navbar_component.html.erb
@@ -1,35 +1,60 @@
-<div class="navbar limit-content-width">
-    <%= render "components/logo" %>
-    <div class="navbar__desktop">
-        <ul>
-            <%= nav_link("Home", '/') %>
-            <% resources.each do |resource| %>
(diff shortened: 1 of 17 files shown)


peteryates (approved):
Looks good and works well for me locally. Nice.

One addition it might be worth making would be to add a few 'hardcoded' items to the `autocomplete_results` so the main 'Events' page can be linked if searching for 'event'; that can easily follow on though.

jebw:
Yeh - the actual search process I think needs more iteration - I'm not really happy with the logic as it stands but this was more just getting a better representation of the results

peteryates (approved):
Looks good and works well for me 👍🏽

PR #875, opened 2021-02-09
NoJS search tweaks

jebw:
### Trello card

https://trello.com/c/1VGYGEd3

### Context

Minor tweaks to improve the NoJS version of search

1. Added missing alt text on search icons
2. Render the search results better when they are hovered over or click - show a background, ensure the whole block is highlighted appropriately, de-emphasize the borders between results
3. Only show search again message on subsequent searches

### DIFF
--- a/app/views/searches/show.html.erb
+++ b/app/views/searches/show.html.erb
@@ -9,7 +9,7 @@
   <div class="govuk-form-group">
     <%= f.button name: nil, class: "request-button" do %>
       Search
-      <%= "again" if @search %>
+      <%= "again" if @search.search&.present? %>
       <%= fas_icon "search" %>
     <% end %>
   </div>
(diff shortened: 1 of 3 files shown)


app/views/searches/show.html.erb
```diff
@@ -9,7 +9,7 @@
   <div class="govuk-form-group">
     <%= f.button name: nil, class: "request-button" do %>
       Search
-      <%= "again" if @search %>
+      <%= "again" if @search.search&.present? %>
```
ethax-ross, line 12: `@search.search` reads a bit odd to me, I don't feel strongly about it but `@search.term` or `@search.text` is more readable IMO

ethax-ross (approved):

app/views/searches/show.html.erb
jebw: Yeh, I think I'll keep it as is in this PR but I'll probably tweak this in the future - I'd like searches to be something like `/search?q=my+search+term`

PR #876, opened 2021-02-09
Fix for nav items showing too high on wide displays with search enabled

jebw:
### Trello card

https://trello.com/c/1VGYGEd3

### Context

Nav bar items are showing too high when highlighted on desktop when there is only a single row of nav items and fonts are rendering with a thinner font/narrower stride - ie quite a wide monitor on Chrome/IE on windows

1. Ensure flexbox is vertically centering the nav items

### Guidance to review

View in chrome on Windows with the page hitting its max-width limits

**Before**

Screenshot from 2021-02-09 16-40-17

**After**

Screenshot from 2021-02-09 16-39-17

### DIFF
--- a/app/webpacker/styles/header.scss
+++ b/app/webpacker/styles/header.scss
@@ -14,6 +14,7 @@
 
   &__desktop {
     display: flex;
+    align-items: center;
     box-sizing: border-box;
     margin-left: 280px;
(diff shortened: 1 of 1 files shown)


ethax-ross (approved):
LGTM

PR #883, opened 2021-02-11
Auto chevroning

peteryates:
### Trello card

N/A (tech debt)

### Context and changes

We currently have to manually add spans around the final text in button-style links so the chevron won't wrap to the next line. This is a pain, adds noise to templates, complicates testing, is easily forgettable and generally a bit of a code smell.

This addition to the link controller automates the process by:

1. looping through the links with a `.button` class (also `.type-description__link` but I'll probably replace this later with a more-descriptive name)
2. splitting the words into an array and wrapping the final one in a `span` tag
3. recombining the text into a string and overwriting the original button text

**At the moment it'll ignore any link that already contains a `<span>`, which should make the upgrade smoother. We can probably remove that step once the `content` repo has been updates (DFE-Digital/get-into-teaching-content/pull/312).**

| Before | After |
| ------ | ------- |
| Screenshot from 2021-02-11 16-02-27 | Screenshot from 2021-02-11 16-02-52 |

### Guidance to review

Concentrate on `link_controller.js`.

Is the approach sensible? If people have JavaScript disabled they'll get no chevrons. The replacement appears to work well.

### DIFF
--- a/app/webpacker/controllers/link_controller.js
+++ b/app/webpacker/controllers/link_controller.js
@@ -56,7 +56,6 @@ export default class extends Controller {
 
     return words
       .map((word, i) => {
-        console.debug(word, i);
         return i === wordCount ? `<span>${word}</span>` : word;
       })
       .join(' ');
--- a/app/webpacker/controllers/link_controller.js
+++ b/app/webpacker/controllers/link_controller.js
@@ -6,6 +6,7 @@ export default class extends Controller {
   connect() {
     this.preventTurboLinksOnJumpLinks();
     this.openExternalContentLinksInNewWindow();
+    this.prepareChevronOnButtonLinks();
   }
 
   preventTurboLinksOnJumpLinks() {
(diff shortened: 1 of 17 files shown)


app/webpacker/controllers/link_controller.js
```diff
@@ -35,4 +36,28 @@ export default class extends Controller {
...
+      .filter((link) => link.querySelector('span') === null) // don't touch buttons that already include a span
+      .forEach((link) => {
+        link.innerHTML = this.wrapFinalWordInSpan(
+          link.textContent.trim().split(' ')
+        );
+      });
+  }
+
+  wrapFinalWordInSpan(words) {
+    const wordCount = words.length - 1;
```
ethax-ross, line 55: I think this could be a bit simpler:

```
return words.push(`<span>${words.pop()}</span>`).join(' ');
```

ethax-ross (dismissed):
Looks good, one suggestion you can take or leave.

Do we have any buttons that just contain an icon/image instead of words? should we account for that potential scenario just in case or is that overkill?

app/webpacker/controllers/link_controller.js
peteryates: Ah, that's much cleaner. Unfortunately JavaScript's `Array.push` returns the new length so it's unchainable; I used `Array.concat` instead.

ethax-ross: Ah, nice 👍

ethax-ross (approved):

peteryates:
Cheers, did a check for buttons with images/icons and the only one we have is the search 🔍 which is marked up differently. Can revisit if we get some.

PR #889, opened 2021-02-12
Move form status styles back inside form rule

peteryates:
The prefix ampersands for the styles at target the form's status data attributes were lost during the linting process (a7e7b01) and have been re-added. As they're flagged by the linter a new config option on the 'selector-no-qualifying-type' rule has been added to make it ignore attributes.

Also fix a typo where the form had the id `cookie-preferences-from`.


jebw (approved):
LGTM

PR #898, opened 2021-02-15
Add intro/context to a mailing list sign up via magic link

ethax-ross:
### Trello card

[Trello-776](https://trello.com/c/HILK0xws/776-mailing-list-sign-up-process-for-crm-data-migration)

### Context

When a user signs up to the mailing list via a magic link we take them straight to the degree status question with no context/intro to the sign up process. We want to add some information so its clear why they are signing up.

- Add intro to mailing list sign up via magic link
- Simplify Authenticate step with access_token_used? helper

The `Wizard::Base` now contains a method to determine if/what kind of authentication method has been used.

Instead of manually tracking this information in the `Authenticate` step we can remove the logic and call out to the wizard.

Ensure `timed_one_time_password` dirty tracking is cleared on `ApiError` so that if a user enters a bad code we only try and validate it with the API once (unless they change the code and try again).

### Guidance to review

Screenshot 2021-02-16 at 15 39 57

### DIFF
--- a/app/models/mailing_list/steps/degree_status.rb
+++ b/app/models/mailing_list/steps/degree_status.rb
@@ -7,6 +7,8 @@ module MailingList
                 presence: true,
                 inclusion: { in: :degree_status_option_ids }
 
+      delegate :magic_link_token_used?, to: :@wizard
+
       def degree_status_options
         @degree_status_options ||= query_degree_status
       end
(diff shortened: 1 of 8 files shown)


peteryates (approved):
Looks good overall. The `#access_token_used?` method name threw me a little, on my first read I expected it to tell me whether it'd been used already, not what was actually in use.

ethax-ross:
Ah that's a good point, I'll have a think and maybe raise a follow-up PR to make that clearer 👍

PR #907, opened 2021-02-15
Hide pre-filled optional fields and steps

ethax-ross:
### Trello card

[Trello-771](https://trello.com/c/tHRHvxLk/771-if-a-matched-back-candidate-pre-fills-an-optional-field-then-clears-it-during-sign-up-we-wont-clear-it-in-the-crm)

### Context

If a candidate is matched back to an existing contact in the CRM we pre-fill the form fields where possible.

Currently, if the user lands on an optional step that is pre-filled with a value and they then clear the value the change won't be reflected in the CRM, as the API only supports sending non-nil values to the CRM.

To make the user-experience consistent with the data that ends up in the CRM for now we are going to skip optional steps that get pre-filled (and hide optional fields on required steps).

In addition, to make it so that the API can write `nil` values to the CRM going forward the app logic has changed so that if a step is `skipped?` we will still export the CRM data for the step _back_ to the API. This means we have a complete request model that we can write back to the CRM and avoids duplicating the 'was this attribute shown/filled in by the candidate' logic in the API. This will ultimately let us write `nil` values back to the CRM.

- Add crm data support to store

We want to hide steps that are optional and have been pre-filled with candidate information.

In order to determine if a step can be pre-filled from the CRM we need to retain the CRM values in the store so that they
can later be queried independently of the data written by the app.

The `fetch` method has also been updated to return values for all keys passed in; this makes the logic to determine if all
values are present in a store simpler (otherwise we would also have to check the `count` to ensure all values are present):

```
fetch(%[...], source: :crm).values.all(&:present?)
```

- Support hidden and optional wizard steps

Currently if a step is `skipped?` it will not show in the wizard and the data will not be exported to the CRM.

We are introducing `optional?` steps, whereby if they are pre-filled from the CRM we will not show them to the user. We
do, however, want to export the data to the CRM still (as this may make partial updates more straight forward in the future). The `skipped?` export logic has been changed in a subsequent PR to facilitate this.

- Hide pre-filled optional steps/fields from wizard

When an optional step is pre-filled from the CRM, or when an optional field within a step is pre-filled, we want to hide it from the user.

- Consolidate hidden/skipped states

Instead of having a separate `hidden?` state - which is not shown to the user but is exported - we combine it with the
`skipped?` state and change so that attributes are exported from skipped steps using the existing CRM data.

This is going to be useful because skipped step values will be `nil` for new candidates and will contain the data already in
the CRM for existing candidates. Ultimately this will make it easier to update the API to support sending `nil` values to
the CRM (it avoids having to duplicate the logic for 'was this step shown to the user' in the API).

### Guidance to review

I chose to do a couple of aspects slightly differently from the suggestions in #884:

- Have the `Wizard` persist the crm data to the store

It worked fine writing to the session instead of the store for the magic link operation as that is initiated in a controller, but the access token exchange is performed in the `Authenticate` step and coupled to the validation of the model; it would be tricky to split that out into a controller action and feed the state back to the model.

- Updated the existing store to contain the logic for the crm/combined data

I felt it was getting overly complex to do this in a subclass; you end up hijacking most of the superclass logic. In addition, the store gets coupled to the `Authenticate` step and any optional steps anyway -- I feel it would be better to combine into the existing store for now and split it out if we ever get a use case for a `Wizard` that isn't backed by the CRM.

- Explicitly set `optional?` in the step subclass

The suggestion was to have optional fields in the `Step` base class that are globally applied, but in the TTA we have a number of instances where an attribute will be optional in one journey and required in another, so doing this explicitly in the subclass feels like it will translate easier when we come to port this across.

### DIFF
--- a/app/controllers/event_steps_controller.rb
+++ b/app/controllers/event_steps_controller.rb
@@ -32,12 +32,17 @@ private
   end
 
   def wizard_store
-    Wizard::Store.new session_store
+    ::Wizard::Store.new app_store, crm_store
   end
 
-  def session_store
+  def app_store
     session[:events] ||= {}
     session[:events][params[:event_id]] ||= {}
+  end
+
+  def crm_store
+    session[:events_crm] ||= {}
+    session[:events_crm][params[:event_id]] ||= {}
   end
 
   def load_event
--- a/app/controllers/mailing_list/steps_controller.rb
+++ b/app/controllers/mailing_list/steps_controller.rb
@@ -14,11 +14,15 @@ module MailingList
     helper_method :step_path
 
     def wizard_store
-      ::Wizard::Store.new session_store
+      ::Wizard::Store.new app_store, crm_store
     end
 
-    def session_store
+    def app_store
       session[:mailinglist] ||= {}
+    end
+
+    def crm_store
+      session[:mailinglist_crm] ||= {}
     end
 
     def set_step_page_title
--- a/app/models/events/steps/contact_details.rb
+++ b/app/models/events/steps/contact_details.rb
@@ -7,6 +7,10 @@ module Events
       before_validation if: :telephone do
         self.telephone = telephone.to_s.strip.presence
       end
+
+      def optional?
+        true
+      end
     end
   end
 end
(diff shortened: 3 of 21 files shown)


jebw (approved):
Thats feels loads neater :smiley: :ok_hand:

PR #909, opened 2021-02-16
Remove hero size capping

peteryates:
### Trello card

N/A

### Context and changes

The hero size capping was introduced before the extra height from the button. This makes it possible for some hero content to be made to 'disappear'. The hero cap has been removed entirely and the image has been changed from an `img` tag to a background-image. This should allow us to just allow the grid to contain all the hero content and let the image resize accordingly:

This change does away with responsive images, the same image will be used throughout. We only have a few content pages with mobile images anyway, we can probably remove these.

https://user-images.githubusercontent.com/128088/108066564-68f0a880-7057-11eb-9406-214150dd95cf.mp4

### Guidance to review

Does it work ok? Are you able to make any of the hero disappear?

### DIFF
--- a/app/components/sections/hero_component.html.erb
+++ b/app/components/sections/hero_component.html.erb
@@ -1,5 +1,5 @@
 <%= tag.header(class: classes) do %>
-  <%= responsive_image %>
+  <%= background_image %>
   <div class="hero__title">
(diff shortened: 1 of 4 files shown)


ethax-ross (approved):
Looks good

PR #911, opened 2021-02-16
Give precedence to shown steps during export

ethax-ross:
It is possible to have steps that write to the same attribute (though we only have this in the TTA currently). With the way we currently export the later of the steps value is used (regardless of if its skipped or not). This can lead to incorrect data being exported; instead, we need to prioritise the shown steps so that they overwrite data exported from skipped steps.


peteryates (approved):
Makes sense. Pity `OtherAge` is only a test attribute :(

PR #918, opened 2021-02-18
Add temp logging to diagnose postcode validation error

ethax-ross:
We are seeing an error where an invalid postcode is being submitted as part of a mailing list or teaching event sign up. I suspect this is an invalid postcode against a candidate already in the CRM, which is then pre-filling the store in the app and being submitted _back_ to the API on completion (which picks it up as a validation error).

Add temp logging to see if this is in fact the case. If it is, I can sanitise the postcodes sent back as match-back data from the API (I would have logged this out in the API but doing it in the Rails app is far easier! ☹️)

### DIFF
--- a/app/models/events/wizard.rb
+++ b/app/models/events/wizard.rb
@@ -21,7 +21,10 @@ module Events
 
     def exchange_access_token(timed_one_time_password, request)
       @api ||= GetIntoTeachingApiClient::TeachingEventsApi.new
-      @api.exchange_access_token_for_teaching_event_add_attendee(timed_one_time_password, request)
+      response = @api.exchange_access_token_for_teaching_event_add_attendee(timed_one_time_password, request)
+      # TEMP: debugging invalid postcode issue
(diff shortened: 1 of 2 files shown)


app/models/events/wizard.rb
```diff
@@ -22,13 +22,18 @@ def complete!
     def exchange_access_token(timed_one_time_password, request)
       @api ||= GetIntoTeachingApiClient::TeachingEventsApi.new
       @api.exchange_access_token_for_teaching_event_add_attendee(timed_one_time_password, request)
+      # TEMP: debugging invalid postcode issue
```
peteryates, line 25: Should `response` be assigned here?

ethax-ross: D'oh, yep that would certainly help :)

peteryates (dismissed):
Looks good!

peteryates (approved):
👍🏽

PR #928, opened 2021-02-22
Groundwork for adding breadcrumbs to pages

ethax-ross:
### Trello card

[Trello-676](https://trello.com/c/U47LYXmX/676-breadcrumbs-for-events-pages-stories-potentially-guidance)

### Context

We want to add breadcrumbs to pages so that users can more easily navigated through the content. This PR lays the groundwork for breadcrumbs by updating the page model to support traversing the content hierarchy, adding Loaf for managing breadcrumbs and a new view component to render the breadcrumbs.

- Render PageNotFoundError if markdown content missing

Currently we return a valid `Page` model for markdown content that does not exist; Rails handles rendering the 404 when it finds the template of the `Page` does not exist.

Instead, we now raise a `PageNotFoundError` and explicitly rescue it in the `ApplicationController` to render 404 not found. We need to do this so that we can more easily put in place methods to traverse the content hierarchy when it comes to building breadcrumbs.

- Add methods to support traversing page hierarchy

To build breadcrumbs of the content pages we need to be able to traverse the page hieracy (from the request page up to the root).

Add `parent` method to get the parent of a page; a path may exist without content so we skip parents that don't exist.

Add `ancestors` method to retrun all ancestors of the current page. These will be the breadcrumbs.

- Add Loaf gem for breadcrumbs

Loaf is a fairly minimal/straight forward breadcumb gem that still receives regular updates.

- Rename components/navbar to components/navigation

We will be introducing a `BreadcrumbComponent` which is similar to the `NavBarComponent` in that they are both navigation-related. Renaming will let us group the components together.

- Add BreadcrumbComponent

Add a `BreadcrumbComponent` that renders Loaf crumbs. The current page crumb is not displayed. Can be specified as `under_hero`, which will render it with a grey background and no margin so it sits flush under the hero section.

### Guidance to review

The breadcrumbs will be added in a follow up PR as it required a fair number of changes to the views.

| Default      | Under Hero |
| ----------- | ----------- |
| Screenshot 2021-02-22 at 08 11 03 | Screenshot 2021-02-22 at 08 12 16 |


app/models/pages/page.rb
```diff
@@ -35,6 +37,31 @@ def data
       @data ||= Pages::Data.new
     end
 
+    def parent
+      pathname = Pathname.new(path)
+
+      loop do
```
jebw, line 43: Open ended loops are best avoided. A for loop with a break (even with a high number like 100 loops) is much preferable.

app/models/pages/page.rb
```diff
@@ -35,6 +37,31 @@ def data
...
+      rescue PageNotFoundError
+        next
+      end
+    end
+
+    def ancestors
+      ancestors = []
+      page = self
+
+      loop do
```
jebw, line 57: See above - better to use a for loop

app/components/navigation/breadcrumb_component.rb
```diff
@@ -0,0 +1,9 @@
+module Navigation
+  class BreadcrumbComponent < ViewComponent::Base
+    attr_accessor :under_hero
+
+    def initialize(under_hero: false)
```
jebw, line 5: As I understand it, this is just to allow configuring the theming - two options I can see

1) We require it to be manually configured, and there with be other scenarios for the alternate styling - so it would be better with a name that reflects what it does, not why its used
2) We only use it for under heros - in which case can this just be handled with css selectors? `.this-class + .that-class .breadcrumb` ?

If possible I think (2) is the better way to go

app/webpacker/styles/components/breadcrumb.scss
```diff
@@ -0,0 +1,49 @@
...
+        margin-right: 0.8em;
+        color: black;
+        font-size: 0.842em;
+          @include mq($until: tablet) {
+            font-size: 0.737em;
+          }
+        &:after {
+          content: "";
+          display: inline-block;
+          width: 7px;
```
jebw, line 25: I think the arrow size should reflect the font-size - so this should probably use `em` units to reflect the sizing of the breadcrumb font - likewise the border size

An alternative might be to use `&raquo;` or `&laquo;` in `content` but that might not achieve the desired visual style

jebw (changes requested):
Looks great but the `loop`s should go - `for`'s with `break`'s will achieve the same but be safer

Arrow mechanism is more of a suggestion

app/components/navigation/breadcrumb_component.rb
ethax-ross: Yeah that's a far better idea 👍

jebw (approved):
LGTM :+1:

PR #929, opened 2021-02-22
Add Breadcrumbs to various layouts

ethax-ross:
### Trello card

[Trello-676](https://trello.com/c/U47LYXmX/676-breadcrumbs-for-events-pages-stories-potentially-guidance)

### Context

We want to add breadcrumbs to pages so that users can more easily navigated through the content. This PR builds on a previous PR to render breadcrumbs in the various layouts, removing redundant back links and tweaking the CSS to make everything consistent.

- Remove back links

These back links will be replaced by the `BreadcrumbComponent`.

- Update MailingList::StepsController layout

The `Events::StepsController` uses the `registration` layout. I'm not sure why the `MailingList::StepsController` uses a different one, but it makes sense for them to be consistent.

It will also mean we can omit the `BreadcrumbComponent` from the registration pages more easily if they both use the same layout.

- Remove no-hero class

This was added so that we can apply a margin above the content when the hero section is not present. It will no longer be needed as the `BreadcrumbComponent` will add the relevant margin in.

The only place we still need to add a top-margin in is in the wizard completion steps; these have a title that would otherwise overlap the logo.

Add breadcrumbs to layouts

Add the `BreadcrumbComponent` to each applicable layout.

Minor CSS tweaks to ensure everything displays consistently with breadcrumbs.

### Guidance to review

| Default      | Under Hero |
| ----------- | ----------- |
| Screenshot 2021-02-22 at 09 23 06 | Screenshot 2021-02-22 at 09 23 18 |


peteryates:
Overall I think this looks good. Codewise it all feels like it makes sense.

I definitely think we should only use them on pages where there's a hierarchy though (stories/events). The dangling 'Home' on the flatter content feels out of place.

sigmaben:
@peteryates @ethax-ross  I'm on the fence about that. on the one hand, there are pages with a hanging home, but it does add consistency. I'd suggest we ship as-is, then I can get a consensus from others on the design channels about the best way forward, and change if needed.

peteryates (approved):
Looks good and works well for me.

PR #932, opened 2021-02-22
Remove the semantic scaffolding class

peteryates:
Now the non-semantic main styles are gone we don't need to refer to the semantic ones by name any more.

I've checked this quite thoroughly and everything looks identical. We now only have `main` CSS rules in `layout.scss` and `home.scss` and the only class on any of the `main` elements in the app is `home` on the homepage layout.


ethax-ross (approved):
Gave it a click about as well and all looks good 👍

PR #933, opened 2021-02-23
Track js searches

jebw:
### Trello card

https://trello.com/c/7OD34EtH

### Context

We want to log search queries to Google Analytics

1. Include the GA Ids on the body tag - currently they are in Tag Manager but inaccessible to the application itself
2. Added a JS class to redact dates, emails and full postcodes
3. Changed the Searchbox stimulus controller to submit searches as a `/search?` url to google analytics


app/webpacker/controllers/searchbox_controller.js
```diff
@@ -64,8 +74,15 @@ export default class extends Controller {
   }
 
   onConfirm(chosen) {
-    if (chosen && chosen.link)
+    if (chosen && chosen.link) {
+      if (this.analyticsSubmitter) {
+        this.clearAnalyticsSubmitter()
+
+        this.sendToAnalytics()
```
peteryates, line 81: `clearAnalyticsSubmitter()` is also called at the top of `sendToAnalytics()`, I think it makes more sense there than here?

peteryates (changes requested):
Looks good overall, just noticed that there's subsequent calls to `clearAnalyticsSubmitter()` in one place.

app/webpacker/controllers/searchbox_controller.js
```diff
@@ -77,4 +94,40 @@ export default class extends Controller {
...
+  scheduleSubmissionToAnalytics() {
+    this.clearAnalyticsSubmitter() ;
+
+    this.analyticsSubmitter = window.setTimeout(this.sendToAnalytics.bind(this), 2000)
+  }
+
+  sendToAnalytics() {
+    this.clearAnalyticsSubmitter() ;
+
+    if (window.ga && this.google_analytics_id) {
```
ethax-ross, line 118: Is there a reason we're using google analytics directly here instead of tag manager? I'm _fairly_ sure you can create and write page view events to `dataLayer` directly and GTM will pick it up instead of going directly to GA.

ethax-ross (commented):
Looks good overall, just one question around sending page views to GA.

app/webpacker/controllers/searchbox_controller.js
jebw: Yes, in theory - that was my thought but I was never able to get it to work. The way GA operates is you have watch for things happening, and hook those observers up to actions. So in theory you can create an event in GTag, than get that event to send the page view.

I couldn't get that to work and just sending straight to GA was _a lot_ easier - it should be lower effort if we need to send other stuff through to GA in the future. Tbh I feel GTag just adds to the complexity rather than reducing it.

peteryates (approved):
Looks good to me.

PR #935, opened 2021-02-23
Rename online events to online Q&As

jebw:
### Trello card

https://trello.com/c/qFdcdHa2

### Context

We are now calling online events, online Q&As

1. Some refactoring - moving variable names and simplifying tests
2. Changed the category name
3. Redirect the old category name

### Review notes

There's a few different refactors in here - they are each split out as their own commit so I can submit them as separate PRs if thats easier?

The `EventBoxComponent` spec changes do drop 3 tests. When I expanded it out I found 1 test checking for an empty string, 1 test running, but the expectation inside it not running because of the conditional, and 2 tests both check for the presence of the 'Online event' text - instead I've change that to be a check that theres only 1 footer item for online events

### DIFF
--- a/app/components/events/event_box_component.rb
+++ b/app/components/events/event_box_component.rb
@@ -7,6 +7,14 @@ module Events
     alias_method :virtual?, :virtual
 
     def initialize(event)
+      # An event can have three different 'states':
+      # - Offline (in person)
+      #   - is_online == false && is_virtual == false
+      # - Moved online (online, but has an associated building - aka virtual)
+      #   - is_online == true && is_virtual == true
+      # - Online (fully online with no building)
+      #   - is_online == true && is_virtual == false
+
       @event       = event
       @title       = event.name
       @type        = event.type_id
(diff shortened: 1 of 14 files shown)


app/components/events/event_box_component.rb
```diff
@@ -7,6 +7,14 @@ class EventBoxComponent < ViewComponent::Base
     alias_method :virtual?, :virtual
 
     def initialize(event)
+      # An event can have three different 'states':
+      # - Offline (in person)
+      #   - is_online == false && is_virtual == false
+      # - Moved online (online, but has an associated building - aka virtual)
+      #   - is_online == false && is_virtual == true
```
ethax-ross, line 14: If an event has moved online it should be is_online == true && is_virtual == true

ethax-ross (commented):
Looks great - one thing I think I remember Myles mentioning at some point is that we want the URL to remain as `/online-events` because `/online-q-as` looks a bit odd; that may have changed since but one to check maybe?

app/components/events/event_box_component.rb
jebw: Ah - good call

jebw:
Regarding the URL - I'd discussed this with @MylesJarvis yesterday and it was decided its fine to change it, there's a redirect in there as well

I've got a separate commit to rejig the URL handling so its easier to split this out in the future, I'd abandoned that but I'll fetch it out again - at the moment the `:id` for the url is determined by the caller of url helper, really it needs to be the url helper deciding what the url should be.

That being said if there's discussions around tickets its worth capturing conclusions in the notes on the ticket for whoever picks it up.

ethax-ross (approved):
LGTM

PR #936, opened 2021-02-24
Add CTA injector to markdown handler

peteryates:
Currently we're breaking large documents up into little chunks so we can place _rich content_, aka calls to action, buttons and other bits of HTML amongst them.

This makes working with documents tricky as it's difficult to see them all at once, we end up flitting between several files.

This approach make the following improvement:

* CTA's can be _registered_ on a page by adding a `calls_to_action`
  section to the front matter
* the `calls_to_action` section will be a list of keys of arbitrary
  names where the value is the arguments required to generate the CTA
* the arbitrary names will need to be matched by placeholders in the
  document in order for the CTA to be injected in the right place
* placeholders will take the form: `$get-an-adviser$`

### DIFF
--- a/app/components/calls_to_action/renderer_component.rb
+++ b/app/components/calls_to_action/renderer_component.rb
@@ -34,8 +34,6 @@ module CallsToAction
     #
     # cta: my_call_to_action
     def basic_call_to_action_component
-      return if @params.blank?
-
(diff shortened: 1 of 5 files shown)


ethax-ross (approved):
Looks good to me 👍

PR #938, opened 2021-02-25
Upgrade to Ruby 2.7.2

peteryates:
Upgrade to Ruby 2.7.2

2.6.6 is getting quite long in the tooth. I only noticed we were behind when I ran into a problem caused by the lack of separation between positional and keyword args, which was implemented in 2.7.0.


ethax-ross (approved):
Gave it a quick smoke test and all looks good 👍

PR #939, opened 2021-02-26
Make the component renderer more generic

peteryates:
Now it can render types of component other than calls to action and it can handle arrays of arguments in addition to hashes of keyword args.


ethax-ross (approved):
Looks good 👍

PR #947, opened 2021-03-01
Circuit breaker redirect

jkempster34:
### Trello card
https://trello.com/c/e9vBMLJE

### Context
When repeated API failures occur, circuit breaker middleware in the Ruby client will raise a `GetIntoTeachingApiClient::CircuitBrokenError`. This is an attempt to reduce the impact of any future service failures.

- Redirect to service unavailable pages when there is a `GetIntoTeachingApiClient::CircuitBrokenError`

### Guidance to review
This PR is paired with https://github.com/DFE-Digital/get-into-teaching-api-ruby-client/pull/37

### DIFF
--- a/app/controllers/concerns/circuit_breaker.rb
+++ b/app/controllers/concerns/circuit_breaker.rb
@@ -1,5 +1,4 @@
 module CircuitBreaker
-  class CircuitBrokenError < RuntimeError; end
   class NotAvailablePathMissingError < RuntimeError; end
 
   extend ActiveSupport::Concern
--- a/app/controllers/concerns/circuit_breaker.rb
+++ b/app/controllers/concerns/circuit_breaker.rb
@@ -1,0 +1,20 @@
+module CircuitBreaker
+  class CircuitBrokenError < RuntimeError; end
+  class NotAvailablePathMissingError < RuntimeError; end
+
+  extend ActiveSupport::Concern
+
+  included do
+    rescue_from GetIntoTeachingApiClient::CircuitBrokenError, with: :redirect_to_not_available
+  end
+
+protected
+
+  def redirect_to_not_available
+    redirect_to(not_available_path)
+  end
+
+  def not_available_path
+    raise NotAvailablePathMissingError
+  end
+end
--- a/app/controllers/concerns/circuit_breaker.rb
+++ b/app/controllers/concerns/circuit_breaker.rb
@@ -14,6 +14,6 @@ protected
   end
 
   def not_available_path
-    raise NotAvailablePathMissingError
+    raise NotAvailablePathMissingError, "#not_available_path hasn't been overridden"
   end
 end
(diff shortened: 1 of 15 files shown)


spec/components/service_unavailable_component_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "rails_helper"
```
jkempster34, line 1: Not sure if this is useful?

spec/requests/circuit_breaker_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  before do
+    api_constants = GetIntoTeachingApiClient.constants.select { |klass| klass.to_s.end_with?("Api") }
+    apis_to_mock = api_constants.map { |const| GetIntoTeachingApiClient.const_get(const) }
+    apis_to_mock.each do |api|
+      api.instance_methods(false).each do |method|
+        allow_any_instance_of(api).to receive(method).and_raise(error)
+      end
+    end
+  end
+
```
jkempster34, line 15: I wanted to parameterise these, but wasn't sure how to pass in the route helpers

spec/components/service_unavailable_component_spec.rb
```diff
@@ -0,0 +1,10 @@
...
+
+RSpec.describe ServiceUnavailableComponent, type: :component do
```
jkempster34, line 3: Not sure if this test is useful

peteryates: Makes sense to have it, even if it's minimal.

spec/requests/circuit_breaker_spec.rb
peteryates: You should be able to do it by setting up outside of the RSpec block and looping through the scenarios you want to test. There's an example in the https://github.com/DFE-Digital/get-into-teaching-app/blob/master/spec/requests/redirects_spec.rb

app/controllers/concerns/circuit_breaker.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+protected
+
+  def redirect_to_not_available
+    redirect_to(not_available_path)
+  end
+
+  def not_available_path
+    raise NotAvailablePathMissingError
```
peteryates, line 17: Should this be returning a path rather than re-raising?

jkempster34: I think this was meant as a way to ensure any class that extends the CircuitBreaker concern implements `not_available_path`

peteryates: Ah, fair enough - maybe a tiny explanation in the error message could help here:

```ruby
raise NotAvailablePathMissingError, "#not_available_path hasn't been overridden"
```

peteryates (approved):
Everything here looks fine to me.

PR #950, opened 2021-03-02
Convert front matter hash keys to strings in specs

peteryates:
### Trello Card

https://trello.com/c/TUQV9Gtq/669-normalize-keys-format-for-frontmatter

### Context and changes

Parsing front matter *always* results in a hash with string keys. We're working around this in the specs by building hashes with symbol keys and then calling `#with_indifferent_access`

This obfuscates what's going on and will add confusion for future developers.

`ag with_indifferent_access spec/` should yield no results.


ethax-ross (approved):
LGTM

PR #953, opened 2021-03-02
Allow mailing list sign up if already signed up for TTA

ethax-ross:
### Trello card

[Trello-1327](https://trello.com/c/2hX9pdEz/1327-spike-consider-unblocking-mailing-list-for-ppl-who-have-signed-up-for-an-adviser)

### Context

Previously we blocked someone signing up for the mailing list if they had already signed up for a TTA (as having a TTA meant they were further along in their journey/the mailing list wouldn't provide value).

As it turn s out, candidates (especially students), engage with an EEA in years 1 and 2. EEAs want to push them to the mailing list so they are kept engaged, so we need to unblock that journey.

- Allow mailing list sign up if already sigend up for TTA

### DIFF
--- a/app/models/mailing_list/steps/already_subscribed.rb
+++ b/app/models/mailing_list/steps/already_subscribed.rb
@@ -2,7 +2,7 @@ module MailingList
   module Steps
     class AlreadySubscribed < ::Wizard::Step
       def skipped?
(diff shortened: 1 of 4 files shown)


peteryates (approved):
lgtm

PR #956, opened 2021-03-02
Remove nowrap from button mixin

jkempster34:
### Trello card
N/A

### Context
Long buttons were not wrapping on mobile.

- Remove nowrap from button mixin
- Add class for when nowrap is needed


peteryates (approved):
Looks good

peteryates:
Actually I just gave this a second run-through and noticed one place where we'd need to prevent it from wrapping:

Screenshot from 2021-03-03 10-13-07

peteryates (changes requested):
The white 'Sign up here' button looks like it could do from a `--nowrap`, the rest all looks fine though.

peteryates (approved):
Nice!

PR #958, opened 2021-03-03
Black site footer background to match rest of footer when wide

jkempster34:
### Trello card
N/A

### Context
See pics. Black footer background doesn't match rest of footer past 2000px

- Move background colour to footer vs width limited wrapper

### Guidance to review
image
image


peteryates (approved):
Yep, looks much better 👍🏽

PR #959, opened 2021-03-03
Try downgrading webmock to make content specs pass

peteryates:
We've seen some failures on the content repo content page specs and webmock looks like the only recently-updated related gem, so this is just testing the theory that it's affected the test suite in some way.

Reverting #943


peteryates:
Scratch this, there were successful builds after this was merged.

PR #966, opened 2021-03-08
Floating mailing list CTA

peteryates:
## Trello Ticket

https://trello.com/c/1bO81rCk/1382-convert-floating-feedback-feature-which-appears-after-3-pages-into-link-to-mailing-list

## Context and changes

Replace the floating feedback banner with a mailing list one.

The appearing after visiting three pages behaviour and dismissal is still in place. The white button will take users to the first page of the mailing list wizard.

| Desktop | Mobile |
| ------ | ------- |
| Screenshot from 2021-03-08 13-15-44 | Screenshot from 2021-03-08 13-16-14 |


app/webpacker/controllers/mailing-list_controller.js
```diff
@@ -50,11 +50,11 @@ export default class extends Controller {
     return parseInt(count);
   }
 
-  dismissFeedback() {
-    return window.localStorage.setItem(feedbackDismissedKey, 'true');
+  dismissMailingList() {
+    return window.localStorage.setItem(mailingListDismissedKey, 'true');
```
ethax-ross, line 54: May be one for the future, but would it be better to set this on the mailing list completion page? I'm thinking if someone hits the CTA then things 'I'll do this later' and comes back to the website then they have no way of getting back to the mailing list sign up? (unless they happen across one of the CTAs in the content somewhere?)

ethax-ross (approved):
One question/comment but looks good code-wise 👍

app/webpacker/controllers/mailing-list_controller.js
peteryates: Good point. Perhaps a ever-present link in the footer might make sense here?

ethax-ross (approved):

PR #969, opened 2021-03-09
Paginate event search results per category

ethax-ross:
### Trello card

[Trello-1389](https://trello.com/c/SHyE7HUc/1389-paginate-event-search-results)

### Context

When a user searches for events in a given month we currently only show 9 per category. Whilst this can show the majority of some event types there are often more School and University events, resulting in the user only seeing events for the next few days (even though they have the whole month selected).

It's not clear that there are more events matching their search parameters - they would need to click on `Explore School and University Events` and then re-filter by their month to see them, which is a lot to expect users to work out.

Instead, we want to add a pagination control per category so that users can quickly cycle through to future events matching their search terms. We want the user's position on the page to remain static so they don't have to keep scrolling back down to the category they are paginating through, which would be jarring.

- Add ScrollController to preserve position on page change

We want to be able to preserve the scroll position when selecting pagination pages (so the transition is not jarring to the user).

```
1. Add `ScrollController` to preserve/restore scroll position.
2. Add controller to all layouts.
3. Update pagination partials to preserve scroll position.
```

- Extract localStorage stub into module

We will need to re-use this for the `ScrollController` specs potentially, so making it reusable.

- Update GroupPresenter to handle pagination

We want to be able to paginate the events sorted by type if they are part of search results (as opposed to the upcoming events).

Add `paginated_events_by_type` method that paginates each set of events according to the arguments (`pages_by_type` contains the active page for each type and `per_page` is how many events per page we want to display).

Add `paginated_events_of_type` convenience method so that we can update `EventCategoriesController` to use the presenter for pagination as well.

- Display search results paginated

When we land on the search page we want to continue to show the next 3 events per category. When we search for events we want to show 9 per category and offer the user a page control to navigate to more events matching their search criteria.

- Use GroupPresenter pagination methods

We now paginate in the `GroupPresenter` for the main event search results; this updates EventCategoriesController to use the presenter methods.

### Guidance to review

I think the event controllers are probably a bit busier than they need to be; I may follow this up with another PR to try and DRY them up.

https://user-images.githubusercontent.com/29867726/110491459-359aba00-80e9-11eb-9409-64fed937f003.mov

### DIFF
--- a/app/controllers/event_categories_controller.rb
+++ b/app/controllers/event_categories_controller.rb
@@ -6,7 +6,6 @@ class EventCategoriesController < ApplicationController
 
   breadcrumb "events.search", :events_path
 
-  EVENTS_PER_PAGE = 9
   MAXIMUM_EVENTS_IN_CATEGORY = 1_000
 
   def show
@@ -46,8 +45,7 @@ private
     @event_search = Events::Search.new(event_filter_params.merge(type: @type.id, period: period))
     events_by_type = @event_search.query_events(MAXIMUM_EVENTS_IN_CATEGORY)
     group_presenter = Events::GroupPresenter.new(events_by_type, false, @event_search.future?)
-    events_of_type = group_presenter.sorted_events_of_type(@type.id)
-    @events = paginate(events_of_type)
+    @events = group_presenter.paginated_events_of_type(@type.id, params[:page]) || []
   end
 
   def load_type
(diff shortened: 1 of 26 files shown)


peteryates (approved):
Code-wise this looks fine and it works well, so I'm approving it.

I think having multiple sets of pagination on a single page is likely to confuse users, and in general the search results page doesn't feel like it has focus. I think we need to revisit it generally.

PR #971, opened 2021-03-10
Add the mailing list banner to stories layouts

peteryates:
It was missing as the mailing list bars were dropped in as a replacement for the feedback ones, which were also absent from the stories pages.


MylesJarvis (approved):
Looks good

PR #972, opened 2021-03-10
Change to Common Action

sfawcett123:
### Context
We use the code three times in different repositories, and we also might want to use the code in other projects. so we have decided to move it to a common repository

Github action [AddTrelloComment](https://github.com/DFE-Digital/github-actions/tree/master/AddTrelloComment) created

### Guidance to review
Does not effect application
Should not impact workflows

### DIFF
--- a/.github/workflows/trello.yml
+++ b/.github/workflows/trello.yml
@@ -2,14 +2,14 @@ name: Link Trello card
 
 on:
   pull_request:
-    types: [ opened ]
+    types: [ opened , edited]
 
 jobs:
   attach-to-trello:
     name: Link Trello card to this PR
     runs-on: ubuntu-latest
+    environment: Development
     steps:
-
       - uses: Azure/login@v1
         with:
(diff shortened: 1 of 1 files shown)


peteryates (approved):

.github/workflows/trello.yml
```diff
@@ -2,14 +2,14 @@ name: Link Trello card
 
 on:
   pull_request:
-    types: [ opened ]
+    types: [ opened , edited]
```
jebw, line 5: This is a late review, but when I wrote this I'd excluded `edited` by choice not accident

Every time a PR is edited the action will run and create another link to Trello - I'd decided to avoid this because the code in the action doesn't check the earlier Trello card comments for existing links?

peteryates: Ah - my bad on this, I assumed it'd check first. It seemed like this was to cover the case where a PR was created without a Trello link _then_ one was added.

PR #973, opened 2021-03-10
Add a basic sitemap generator

peteryates:
### Trello card

https://trello.com/c/CEFa8IGk/1401-create-or-update-our-sitemap

### Context and changes

It uses both the combined frontmatter from all content (via `Pages::Frontmatter`) and a hard-coded list of extra paths, which aren't generated from markdown files (like /events) to build a [sitemap](https://www.sitemaps.org/protocol.html).

### Guidance to review

Find it at `/sitemap.xml`

### DIFF
--- a/app/controllers/sitemap_controller.rb
+++ b/app/controllers/sitemap_controller.rb
@@ -1,0 +1,46 @@
+class SitemapController < ApplicationController
+  DEFAULT_LASTMOD = "2021-03-01".freeze
+
+  ALIASES = {
+    "/home" => "/",
+  }.freeze
+
+  OTHER_PATHS = %w[
+    /events
+    /event-categories/train-to-teach-events
+    /event-categories/online-q-as
+    /event-categories/school-and-university-events
+  ].freeze
+
+  def show
+    render xml: build.to_xml
+  end
+
+private
+
+  def build
+    Nokogiri::XML::Builder.new(encoding: "UTF-8") do |xml|
+      xml.urlset(xmlns: "http://www.sitemaps.org/schemas/sitemap/0.9") do
+        Pages::Frontmatter.list.each do |path, metadata|
+          xml.url do
+            xml.loc(request.base_url + page_location(path))
+            xml.lastmod(metadata.fetch("date") { DEFAULT_LASTMOD })
+            xml.priority(metadata.fetch("priority")) if metadata.key?("priority")
+          end
+        end
+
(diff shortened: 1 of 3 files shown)


spec/requests/sitemap_controller_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      },
+      "/my-story-into-teaching/what-a-great-story" => {
+        "name" => "What a great story",
+        "priority" => 0.7,
+        "date" => "2020-10-10",
+      },
+    }
+  end
+
+  before { allow(Pages::Frontmatter).to receive(:list) { nodes } }
```
ethax-ross, line 25: It took me a second to realise what `nodes` were... would `pages` be clearer?

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  ALIASES = {
+    "/home" => "/",
+  }.freeze
+
+  OTHER_PATHS = %w[
+    /events
+    /event-categories/train-to-teach-events
+    /event-categories/online-q-as
+    /event-categories/school-and-university-events
```
ethax-ross, line 12: Should the privacy policy/cookies/guidance pages be in here? and would it be worth indexing the sign up for the mailing list initial page?

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,46 @@
+class SitemapController < ApplicationController
+  DEFAULT_LASTMOD = "2021-03-01".freeze
```
ethax-ross, line 2: Is there a particular rational behind this date, or is it just a case of it needs to be static and in the past? I guess it makes sense as a sort of 'when the website launched' date?

ethax-ross (approved):
Looks good, couple of observations but they are take it or leave it

app/controllers/sitemap_controller.rb
peteryates: Yeah, I'm planning to do a sweep of the content repo and add `priority` and `date` entries to the markdown, I just picked a past date here for the time being.

spec/requests/sitemap_controller_spec.rb
peteryates: Ah yeah, good call 👍🏽

ethax-ross (approved):

PR #975, opened 2021-03-10
Make markdown h3 headings bigger than h4

peteryates:
A couple of our guidance pages need some distinction. To stop this from bleeding into the events pages the .markdown class has been removed from the event list container.

Here's a `h3` followed by a `h4`:

Screenshot from 2021-03-10 15-42-15

### DIFF
--- a/app/views/layouts/events.html.erb
+++ b/app/views/layouts/events.html.erb
@@ -21,7 +21,7 @@
     <%= render Navigation::BreadcrumbComponent.new %>
 
     <main role="main" id="main-content">
-      <section class="container">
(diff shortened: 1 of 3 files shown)


ethax-ross:
I think this heading has lost styling now as it was being applied via the `markdown` class?

Screenshot 2021-03-10 at 16 13 25

peteryates:
Ah yeah - thanks for that! I've re-pushed and added an `events` class to the events containers (easier to review without whitespace!)

ethax-ross (approved):

PR #983, opened 2021-03-12
Update Sentry gem

ethax-ross:
The sentry-raven gem we were using is deprecated/in maintenance mode. Sentry now have two gems that replace it (sentry-ruby and sentry-rails).

Update to new gems and migrate existing code (migration guide: https://docs.sentry.io/platforms/ruby/guides/rails/migration/)

The `Sentry::BackgroundWorker` sends events to Sentry async by default, using `ActiveRecord::Base.connection_pool`. We don't use `ActiveRecord` in the app, so this was failing silently with an `unrecognized constant ActiveRecord`. Instead of pulling in `ActiveRecord` just for this (which has its own difficulties as it tries to connect to a DB by default) we override the `async` handler to just send everything synchronously (which is how `sentry-raven` was working).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,7 +47,8 @@ gem "loaf"
 
 gem "prometheus-client"
 
-gem "sentry-raven"
+gem "sentry-rails"
+gem "sentry-ruby"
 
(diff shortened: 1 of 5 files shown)


peteryates (dismissed):

peteryates (approved):
Looks good

PR #992, opened 2021-03-16
Keep Artefacts Even on Failure 

sfawcett123:
### Trello card
[Keep Lint Failure Outputs ](https://trello.com/c/C2jiSt2r/1394-keep-lint-failure-output)

### Context
When a LINT fails the following step to save  the results does not run, changed the process to always try to save the results.

added ``` If: failure() ```

### Guidance to review
No changes to the application


jebw (approved):
LGTM

PR #998, opened 2021-03-16
Improved handling of missing non-html pages

jebw:
### Trello card

https://trello.com/c/X2pGWgj8

### Context

Currently we trigger an unhandled exception for missing page requests when the page format is not html

1. Added common 'error renderer' method - this ensures the layout is set to application and handles non-html formats by returning just the status code with an empty body
2. Handle unknown file types in the default error controller as well.

### Guidance to review

The `render_error` method uses the supplied parameter to select the template to render - I've tried to protect against incorrect usage - it requires a symbol, and if its not on of the status code symbols that will in turn trigger an error from the `render` method it uses.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -68,6 +68,7 @@ private
       format.html do
         render \
           template: "errors/not_found",
+          layout: "application",
           status: :not_found
       end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -36,11 +36,11 @@ private
   end
 
   def render_not_found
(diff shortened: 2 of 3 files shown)


app/controllers/pages_controller.rb
```diff
@@ -68,6 +68,7 @@ def rescue_missing_template
       format.html do
```
ethax-ross, line 68: Can be one for later, but I think this could be simplified to:

```
rescue_from *MISSING_TEMPLATE_EXCEPTIONS, with: :render_not_found
```

ethax-ross (approved):

PR #999, opened 2021-03-16
Upgraded to latest simplecov

jebw:
### Trello card

https://trello.com/c/gabkUApE

### Context

We are currently running an out of date copy of simplecov because SonarCloud is using an internal SimpleCov JSON file, and that files format got changed in SimpleCov v0.18

According to https://jira.sonarsource.com/browse/SONARSLANG-477 sonarcloud
should now support the `coverage.json` output which added in SimpleCov 0.20

1. This change adds the new JSONFormatter to create a `coverage.json` file
2. Switches SonarCloud to use that `coverage.json` file

### Guidance to review

I wont know this is right until its merge I think but its easiest enough to revert if it breaks coverage reporting

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,7 +86,7 @@ jobs:
             rspec --format RspecSonarqubeFormatter --out /app/out/test-report.xml --format documentation
 
       - name: Fixup report file paths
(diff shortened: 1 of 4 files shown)


ethax-ross (approved):
LGTM

PR #1000, opened 2021-03-17
Ensure the 404 is rendered in HTML

peteryates:
We've seen an influx of bots poking around for pages that don't exist.

These cause a ActionView::MissingTemplate error because we're trying to serve a 404 in the format they requested the docuemnt in, so `/some-file.md` is expecting us to have a `errors/not_found.md` file, which we don't.

Screenshot from 2021-03-17 12-13-58

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -36,7 +36,7 @@ private
   end
 
   def render_not_found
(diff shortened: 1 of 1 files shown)


peteryates:
Ah, there's a more extensive fix in #998. Closing this one.

PR #1011, opened 2021-03-22
Show either search bar or menu on mobile

jebw:
### Trello card

https://trello.com/c/Nj1GdT7d

### Context

To avoid confusion when we open a searchbar, we should hide the mobile menu and vice versa.

1. Hide mobile menu when opening the search bar
2. Hide search bar when opening the mobile menu


peteryates (approved):
Looks good and works well for me locally.

PR #1014, opened 2021-03-22
Added DfE logo to navbar

sigmaben:
Added DfE Logo to navbar, updated footer logo to svg.


peteryates (approved):
I'm happy with this as-is, I think the whole header is worth a reorganise though, some of the styles look a bit hard to follow.

PR #1018, opened 2021-03-23
Send Google Page Speed insights to Prometheus

ethax-ross:
### Trello card

### Context

We want to monitor and potentially alert on Google Page Speed insights (as they offer a good indication of website performance, accessibility and SEO). In order to monitor the scores we will execute the page speed insights against each web page in the `sitemap.xml` and send the metrics to Prometheus (so that we can build a dashboard in Grafana and monitor with Alert Manager). We should be able to run the page speed insights periodically, most likely daily via GitHub Actions.

- Add google-api-client gem

Enable querying the page speed score API; not required by default.

- Add metrics for page speed scores

We want to record the page speed score for performance, accessibility and SEO in Prometheus.

Adds the metrics as gauges (as they will go up/down) with labels for strategy (mobile/desktop) and path (page under test).

The path label will have a fairly high cardinality, but it is finite (the number of pages on the site) and will give us page-level insights.

- Add PageSpeedScore class

Retrieve page speed scores via the google-api-client for each page listed in the `sitemap.xml`, querying performance, accessibility and SEO categories for both mobile and desktop strategies.

The fields have been restricted to make what is a slow request as fast as possible and the pages are queried asynchronously using up to 10 threads at a time.

- Add rake task to fetch page speed scores

Enable us to query the page speed scores periodically from a GitHub action.

Usage:

```
bundle exec rake 'page_speed_score:fetch_all[http://host.com/sitemap.xml]'
```

- Cache scores and publish metrics on application start

Initially we intended to use the Prometheus Push Gateway to send the metrics directly from the Rake task that retrieves the page speed scores. It turns out that it would be difficult to secure the push gateway, so we need to use a different method.

In this commit the scores are written to Redis from the Rake task and when the application boots in the rolling environment the scores are published for Prometheus to scrape.

### Guidance to review

I haven't gone all out and handled quota exceeded/rate limit exceeded responses or failures from the Google API. It's not a critical bit of production code and the batching should keep us well within usage limits. If we start to see errors I'd rather just address them if/when they happen.


peteryates (approved):
All looks good to me, sensible approach.

`google-api-client` is pulling a lot in, but not sure there's much we can do about that.

ethax-ross:
Yeah, we could go without it but its more work to call the API manually... its not `required` by default as well so hopefully shouldn't have an impact on non-rolling envs

PR #1019, opened 2021-03-23
External link checker

jebw:
### Trello card

https://trello.com/c/pUu6GogQ

### Context

We want a script to check all the external links within the site.

1. Added an rspec test, it is tagged as part of the `onschedule` specs and excluded from runs by default - this can be subsequently run from CI.

### Guidance to review

I've used an RSpec test because I needed a mechanism to render the internal pages we are scanning for links. Its in a single test because we want to share the results across pages though this could potentially be achieved via a singleton.

The checker tries to 'behave well' - it uses HEAD requests, follows redirects, handles connection failures and only queries any given link once per run. The output is just a hash of the failure results but it lists the status code returned (or nil if no response). It also lists the pages which the broken link appears on.

There are several broken links on the assessment providers page it seems.

LinkedIn responds with a status code 999 so I've just excluded it the checks. There is also an (empty) array of domains to use GET requests for if that proves necessary.

I'll add a corresponding PR to content to run this on a schedule once its passed review and been merged

### DIFF
--- a/spec/features/external_links_spec.rb
+++ b/spec/features/external_links_spec.rb
@@ -1,0 +1,60 @@
+require "rails_helper"
+
+RSpec.feature "external links check", :onschedule do
+  describe "all pages" do
+    it "will have no broken links" do
+      results = {}
+      faraday = Faraday.new do |f|
+        f.request :retry, max: 2
+        f.use FaradayMiddleware::FollowRedirects
+      end
+
+      PageLister.content_urls.each do |current_page|
+        puts "\nCHECKING #{current_page}"
+        visit(current_page)
+        document = Nokogiri.parse(page.body)
+        links = document
+                  .css("a")
+                  .pluck("href")
+                  .select { |href| href.starts_with? %r{https?://} }
+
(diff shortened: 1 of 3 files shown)


spec/features/external_links_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+        WebMock.disable_net_connect!
+      end
+
+      failures = results.reject do |_page, response|
+        (200..299).include? response[:status]
+      end
+
+      # comparing to hash instead of checking .empty? so the failures dump
+      # the difference in the hashes
+      expect(failures).to eql({})
```
peteryates, line 23: 🧠

peteryates (approved):
Looks good and work well for me locally.

Only minor thing is there's not much feedback when it's running, looping outside the `it` and doing an assertion per URL could help - but as this'll be run (I assume) by a GitHub action it's not really consequential.

jebw:
Agree there's not much feedback - originally there was a load of debug info in there but if its all good we don't really want feedback since its being run via a job, we just want to know when its a problem

PR #1022, opened 2021-03-25
Instagram doesn't support HEAD requests

jebw:
### Trello card

https://trello.com/c/pUu6GogQ

### Context

The link check uses HEAD requests by default which instagram does not support.

1. Use GET requests instead for Instagram


leoapost (approved):
Looks good

PR #1042, opened 2021-03-30
Add robots controller and beta redirect logic

peteryates:
This is an experiment to see if dealing with the switchover from the
beta domain (beta-getintoteaching.education.gov.uk) to the 'live' domain
(getintoteaching.education.gov.uk) can be handled within Rails while
ensuring the following things happen:

 * people who click 'old' links in their email should be redirected to
   the intended page
 * robots.txt is also redirected, ensuring that search engines won't
   index the `beta-` version
 * query params are maintained during the redirect process

The behaviour can be toggled using the `x.enable_beta_redirects` flag.


spec/routing/beta_redirect_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "rails_helper"
+
+describe "Redirecting beta-getintoteaching.education.gov.uk to getintoteaching.education.gov.uk", type: "routing" do
+  before { allow(Rails.configuration.x).to receive(:enable_beta_redirects).and_return(true) }
+
+  specify "redirects content pages" do
```
ethax-ross, line 6: Is this actually checking that it redirects or just that its routing to the correct controller/action? I would have tested this with a request spec that checks it `redirects_to`?

app/controllers/robots_controller.rb
```diff
@@ -0,0 +1,13 @@
+class RobotsController < ApplicationController
+  layout :none
+
+  def show
+    render plain: <<~ROBOTS
```
ethax-ross, line 5: Worth a comment to say this is so we can redirect requests to the robots.txt on the beta domain to avoid them getting indexed?

ethax-ross (commented):
Looks good, just a couple questions

spec/routing/beta_redirect_spec.rb
peteryates: It was a bit of a pain to test, this is just ensuring the path remains correct and we're relying on Rails's `#redirect` method to work properly. I tested quite a bit locally by setting up the domains in `/etc/hosts` and it all works as expected.

I'll have a look at doing it via a request spec 👍🏽

app/controllers/robots_controller.rb
peteryates: Good point, will do.

ethax-ross (approved):

PR #1055, opened 2021-03-31
Make the external link mixin more modular

peteryates:
This allows it to be dropped in anywhere, making it easier to target buttons and other places where we never want it to appear.

Screenshot from 2021-03-31 15-48-08
Screenshot from 2021-03-31 15-46-06


MylesJarvis (approved):
Happy with the screenshots - but get a dev review if you want to

PR #1063, opened 2021-04-01
Fingerprint content assets

ethax-ross:
### Trello card

### Context

We are not currently able to have an efficient caching strategy on the assets in the content repository due to them not having fingerprints.

By adding fingerprints to these assets we can update the Rails static cache to be much longer than the current 15 minutes (we could change it so that the assets get cached forever).

- Fingerprint content assets

Add a `Fingerprinter` class that fingerprints the assets in the given `asset_dir` and updates references to them in the given `content_dir` (inspired by Middleman's `:asset_hash` extension).

Add a rake task to run the fingerprinting against the content repository assets/content, which could be called as part of the Docker build process:

```
bundle exec rake fingerprinter:run
```

If this gets merged down the change to the content repo Dockerfile can be made and it can all be tested in the generated review app.

The matching is fairly liberal given all the weird and wonderful ways you can link to an image between markdown, ERB and HTML. A side-effect of this is that if the content text contains the path to a valid image it will also get the fingerprint added in - that may or may not be desirable, but its hard to avoid and I can't see a time when we would display the path to an asset directly to the user, so I haven't tried to avoid it for now.

### Guidance to review

This isn't going to run anywhere until the content repo is updated, and I'm fairly certain the existing test to check all internal image URLs are valid will fail if this doesn't work as expected. I've tested it locally and it appears to work well.

The only real issue I ran into is that we appear to link to some webpacker assets from the content repo where the filenames are the same. Luckily they are in different public paths so Rails adds the fingerprint to these assets correctly still. I haven't came across an instance of the Rails app linking to an asset in the content repo, but if that exists it will not work yet -- I'm hoping we don't do this and can just avoid doing it going forward otherwise it will get messy trying to add the fingerprint to any Rails content.


peteryates (approved):
This works well and fingerprints everything correctly.

PR #1072, opened 2021-04-06
Reduce covid banner to one line

sigmaben:
Reduce the size of the covid banner to have less impact on the page on the first load.


peteryates:
Ah sorry @sigmaben - a change that implements this (I've not checked them side by side but I guess they're similar) was merged on Friday #1060 

If you check the live site now you can give it a check and if you still want this as a refinement we can go ahead.

peteryates (approved):
Looks good and works for me locally.

PR #1074, opened 2021-04-06
Increase cache TTL to 1 year

ethax-ross:
### Trello card

### Context

We now fingerprint all the assets in the content repository as well as in the Rails app, so its safe to have them cached
indefinitely going forward (as the fingerprint will change when the contents of the asset changes).

- Increase cache TTL to 1 year


peteryates (approved):
Looks good, quite a jump!

PR #1077, opened 2021-04-06
Feature/feedback UI component

leoapost:
### Trello card
https://trello.com/c/v2QcGzbb/699-move-is-this-page-useful-out-of-the-way

### Context
The 'is this page useful?' is of limited use and not the tidiest of components. We want to replacing it with a new one that will be display across the site.

Create a new feedback section, place it below the 'Talk to us' and stop rendering the 'is this page useful?' component.

### Guidance to review
|Before|After|
|---|---|
|before|after

### DIFF
--- a/app/views/sections/_feedback_bar.html.erb
+++ b/app/views/sections/_feedback_bar.html.erb
@@ -1,0 +1,20 @@
+<div class="feedback-bar">
+  <section class="feedback-bar__inner limit-content-width">
+    <div class="feedback-bar__left">
+      <h2 class="blue small">FEEDBACK</h2>
+    </div>
+    <div class="feedback-bar__right">
+      <div id="page-helpful" data-controller="page-helpful">
+        <div>
+          <strong data-page-helpful-target="text">Is this page helpful?</strong>
+          <a href="javascript:void(0)" class="page-question__answer strong" data-page-helpful-target="link" data-action="page-helpful#answer">Yes</a>
+          <a href="javascript:void(0)" class="page-question__answer strong" data-page-helpful-target="link" data-action="page-helpful#answer">No</a>
+        </div>
+      </div>
+      <div>
+        We'd like your feedback on our new Get into teaching website.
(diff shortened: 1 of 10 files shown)


app/views/sections/_feedback_bar.html.erb
```diff
@@ -0,0 +1,20 @@
...
+      <div id="page-helpful" data-controller="page-helpful">
+        <div>
+          <strong data-page-helpful-target="text">Is this page helpful?</strong>
+          <a href="javascript:void(0)" class="page-question__answer strong" data-page-helpful-target="link" data-action="page-helpful#answer">Yes</a>
+          <a href="javascript:void(0)" class="page-question__answer strong" data-page-helpful-target="link" data-action="page-helpful#answer">No</a>
+        </div>
+      </div>
+      <div>
+        We'd like your feedback on our new Get into teaching website.
+        <a target="_blank" rel="noopener noreferrer" href="https://docs.google.com/forms/d/e/1FAIpQLSdBXbU5nwP6_3TH7HY5rB4ehkSDNzfKqB5X2G7wG4K6LY97-g/viewform">Tell us what you think</a>
```
peteryates, line 16:
```suggestion
        <a target="_blank" rel="noopener noreferrer" href="https://docs.google.com/forms/d/e/1FAIpQLSdBXbU5nwP6_3TH7HY5rB4ehkSDNzfKqB5X2G7wG4K6LY97-g/viewform">Tell us what you think</a>.
```

peteryates (commented):
Overall this looks great. I left one suggestion with a full stop.

Finally I wonder if this should be the width of the 'content' rather than the header. Having had a quick look, adding `max-width: $content-max-width` to `feedback-bar__inner` makes the page look nicer in my opinion, like this:

Screenshot from 2021-04-06 15-14-09

app/views/sections/_feedback_bar.html.erb
leoapost: Both are good suggestions - I've push the updates.

peteryates (dismissed):
All looks good, will reapprove when the final spec's fixed. Nice!

peteryates (approved):
Nice, looks good and works well for me.

PR #1078, opened 2021-04-06
Draft internal events section

jkempster34:


.env.test
```diff
@@ -2,3 +2,8 @@ GIT_API_TOKEN=test
 GIT_API_ENDPOINT=https://host.api/endpoint
 TTA_SERVICE_URL=http://tta-service.test
 PREVIEW_PAGES=1
+
+PUBLISHER_USERNAME=publisher
```
ethax-ross, line 6: We should put any non-local development secrets into Azure, even for the test site

app/controllers/internal/events_controller.rb
```diff
@@ -0,0 +1,140 @@
...
+    layout "internal"
+
+    def index
+      @no_results = load_pending_events.blank?
+    end
+
+    def show
+      @event = GetIntoTeachingApiClient::TeachingEventsApi.new.get_teaching_event(params[:id])
+      @page_title = @event.name
+      if @event.status_id != GetIntoTeachingApiClient::Constants::EVENT_STATUS["Pending"]
```
ethax-ross, line 13: Do we need something similar to ensure pending events 404 on the main events#show path?

app/controllers/internal/events_controller.rb
```diff
@@ -0,0 +1,140 @@
...
+        raise ActionController::RoutingError, "Not Found"
```
ethax-ross, line 14: There's a helper for this you should just be able to call:

```
render_not_found unless @event.status_id == pending_event_status_id
```

^ worth pulling out the constant into a private helper method IMO as its reused and quite a mouthful

app/controllers/internal/events_controller.rb
```diff
@@ -0,0 +1,140 @@
...
+      end
+    end
+
+    def new
+      @event = Event.new
+      @event.building = EventBuilding.new
+      @event.building.fieldset = "existing"
```
ethax-ross, line 21: `fieldset` isn't a very informative name; I would also move the options into an enum or similar so its easy to see what values are available

app/controllers/internal/events_controller.rb
```diff
@@ -0,0 +1,140 @@
...
+    end
+
+    def final_submit
```
ethax-ross, line 24: `approve`?

app/controllers/internal/events_controller.rb
```diff
@@ -0,0 +1,140 @@
...
+      event = GetIntoTeachingApiClient::TeachingEventsApi.new.get_teaching_event(params[:format])
+      @event = transform_event(event)
```
ethax-ross, line 26: I would be inclined to move the mapping to/from an API event to your local event type into the `Internal::Event` so you have something like:

```
@event = Internal::Event.new(api_event)
...
@api_event = @internal_event.to_api_event
```

You could then unit test it

app/controllers/internal/events_controller.rb
```diff
@@ -0,0 +1,140 @@
...
+      @event.start_at = Time.zone.parse(@event.start_at.to_s)
```
ethax-ross, line 27: Could this not be encapsulated in `transform_event`?

app/controllers/internal/events_controller.rb
```diff
@@ -0,0 +1,140 @@
...
+      else
+        render action: :new
+      end
+    end
+
+    def edit
+      event = GetIntoTeachingApiClient::TeachingEventsApi.new.get_teaching_event(params[:id])
+      @event = transform_event(event)
+      if @event.building.nil?
+        @event.building = EventBuilding.new(fieldset: "none")
```
ethax-ross, line 50: I wonder if the `fieldset` attribute should live in the `event` rather than the `building`? having `EventBuilding.new(fieldset: "none")` feels contradictory

app/controllers/internal/events_controller.rb
```diff
@@ -0,0 +1,140 @@
...
+    end
+
+    def load_pending_events
+      opts = {
+        type_id: GetIntoTeachingApiClient::Constants::EVENT_TYPES["School or University event"],
+        status_ids: [GetIntoTeachingApiClient::Constants::EVENT_STATUS["Pending"]],
+        quantity_per_type: 1_000,
+      }
+
+      @events = GetIntoTeachingApiClient::TeachingEventsApi
```
ethax-ross, line 131: Are we re-inventing the wheel a bit here and could the existing `Event::Search` and `Events::GroupPresenter` be used instead?

app/controllers/internal_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+protected
+
+  def publisher?
+    session[:publisher]
+  end
+
+private
+
+  def set_account_role(role)
+    session[role] = true
```
ethax-ross, line 14: I think having multiple session names for each role could lead to weirdness if you log in/out as a publisher/author as you could end up with `session[:author]` and `session[:publisher]` both `true` if you're not careful/clean up. I think having the session attribute static and the value variable would be less error prone:

```
session[:role] = :author
```

app/models/internal/event.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    def end_after_start
+      return if end_at.blank? || start_at.blank?
+
+      if end_at.floor <= start_at.floor
+        errors.add(:end_at, "must be after the start date")
+      end
+    end
+
+    def submit_to_api?
```
ethax-ross, line 74: I think having this in the model is muddying the waters a bit. I would be inclined to have the event/api_event mapping logic in the model but the submission either in the controller or a separate class that handles internal event submissions, This also feels very verbose; I think the mapping to/from an internal event to an api event could be more dynamic (similar to what you have in the `transform_event` controller method already). It feels odd its partly done in the controller one way and then the model the other way; needs making consistent.

ethax-ross (commented):
Lots of good stuff in there, just needs shuffling about and reorganising a bit 😄 I've mainly looked at the controller/models and not so much the views yet

PR #1084, opened 2021-04-07
Serve images in next-gen image formats where possible

ethax-ross:
### Trello card

### Context

We use basic JPEG/PNG images for the most part; a recommendation is to use newer image formats that enable better compression, which improves the page load speed. Instead of manually saving each image in the various formats, we can generate the different formats during the Docker build process and dynamically serve them when they exist from Rails.

We should be able to build on this with a couple of further optimisations; only loading images 'above the fold' on the first paint and serving device-appropriate images (so smaller images for mobile, for example).

It would be nice to serve background images in the different formats as well, but I'm not sure how possible that is without relying on JS - I'll investigate in another PR.

- Add NextGenImages class

Introduces a class that replaces `` tags with `<picture>` tags and applicable `<source>` tags for any next-gen images that exist with the same file name as the original image.

This will allow us to generate next-gen images as part of the Docker build process and then Rails will introduce them into the responses dynamically. As most of our pages are static and cached it shouldn't have a big performance overhead.

- Add next-gen images to all HTML responses

Any HTML responses are intercepted and the next-gen images are introduced when they exist.

Enabled only in preprod initially.

- Tweak CSS to support picture instead of img tags

Some of the CSS targets a specific `img` tag with the `>` selector; these need updating to target the `img` within the new `picture` tag.

- Correct HTML closing tag

It was getting messed up running through Nokogiri (even though it rendered fine by browsers like this).

### Guidance to review

As it stands the existing images will be wrapped in a `<picture>` tag with a single `<source>` for the original image. When we update the content Docker build process to generate images in next-gen formats it should start serving those as well.

I've tested this locally with a decent amount of clicking about and it doesn't seem to break anything. Its only enabled in preprod for now so that I can get some other people to give it a decent smoke test.

### DIFF
--- a/lib/next_gen_images.rb
+++ b/lib/next_gen_images.rb
@@ -47,6 +47,7 @@ private
 
     case ext
     when ".webp"
+      # Rack::Mime.mime_type is returning the wrong mime type for webp.
       "image/webp"
     else
       Rack::Mime.mime_type(ext)
--- a/lib/next_gen_images.rb
+++ b/lib/next_gen_images.rb
@@ -1,0 +1,55 @@
+class NextGenImages
+  # As we control the conversion we only need to look for
+  # a subset of all the possible extensions.
+  NEXT_GEN_IMAGE_EXTS = %w[.jp2 .webp].freeze
+
+  def initialize(page)
+    @page = page
+  end
+
+  def html
+    doc = Nokogiri::HTML(@page)
+    doc.css("img").each do |img|
+      img.replace(picture(img, doc)) unless img.parent.name == "picture"
+    end
+
+    doc.to_html(encoding: "UTF-8", indent: 2)
+  end
+
+private
+
+  def picture(img, doc)
+    src = img.attribute("src")
+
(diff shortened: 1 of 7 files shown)


lib/next_gen_images.rb
```diff
@@ -0,0 +1,55 @@
...
+      source.set_attribute("type", mime_type(src))
+    end
+  end
+
+  def mime_type(src)
+    ext = File.extname(src)
+
+    case ext
+    when ".webp"
+      "image/webp"
```
peteryates, line 50: Is there a reason why `.webp` is a special case here (vs `.jp2`)?

ethax-ross: Yeah for some reason Rack::Mime doesn't know about webp so gives it the wrong mime type, but jp2 it gets right 🙄

peteryates (dismissed):
Nice. Logic-wise it seems sound. Left a question on the case statement, perhaps it just needs a small comment to explain.

peteryates (approved):
Nice! 🖼️

PR #1085, opened 2021-04-07
Support lazy loading images that are below the fold

ethax-ross:
### Trello card

### Context

If an image is below the fold we don't want to load it as part of the first paint. Instead, we only want to load images when the user scrolls the viewport to approach them. This will make the page load faster and give us a better page speed score.

We also want to ensure images still work when Javascript is disabled.

- Support lazy loading images that are below the fold

Intercepts HTML before rendering and removes the `src` attribute, retaining the path in `data-src`.

Appends a `noscript` tag containing the original `img` to support when Javascript is disabled.

The original images are hidden by default, to avoid styling issues when JS disabled (we don't want to show both images as the lazy image will not have a src set).

Lazysizes picks up any img tags with a `lazyload` class and populates the `src` when the user is scrolling towards the image. It also supports lazy-loading the picture source tags with `data-srcset`.

A side-effect of this is that the webp/jp2 image sources will no longer be available if the browser has JS disabled. Whilst we could fix this I don't think its worth the effort given the % of users in this scenario.

Enabled only in preprod environment for now.

### Guidance to review

https://user-images.githubusercontent.com/29867726/113887905-c97ba680-97b9-11eb-8159-c361fcc0f50a.mov

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,5 @@
 require "next_gen_images"
+require "lazy_load_images"
 
 class ApplicationController < ActionController::Base
   include UtmCodes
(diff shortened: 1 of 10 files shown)


peteryates (approved):
Nice! 💨

PR #1086, opened 2021-04-07
Disable VWO except for specific pages

MylesJarvis:
### Trello card

[Trello-1491](https://trello.com/c/4gpMaUPW/1491-remove-vwo-after-cutover-ready-to-implement-on-individual-pages-for-a-b-testing)

### Context

By default we want to disable VWO on all pages, as we are no longer using it for redirections/splitting traffic.

We will be using VWO in the future to run a/b tests, however, so its useful to keep it around but have a mechanism to enable it on specific paths.

Updates `AnalyticsController` to work when VWO is disabled.

- Conditionally enable VWO on pages


peteryates (dismissed):
⏏️

peteryates (approved):

PR #1087, opened 2021-04-08
Post cutover: Update .env.production to reflect new TTA service URL (without beta-)

MylesJarvis:
### Trello card
N/A
### Context
Updating our TTA service URL so that our links on the site point to the new site.

Update `TTA_SERVICE_URL` to remove 'beta-'

### Guidance to review
Do the specs need updating?

### DIFF
--- a/.env.production
+++ b/.env.production
@@ -10,4 +10,4 @@ BAM_ID=1
 LID_ID=1
 HOTJAR_ID=1871524
 VWO_ID=524595
-TTA_SERVICE_URL="https://beta-adviser-getintoteaching.education.gov.uk/"
+TTA_SERVICE_URL="https://adviser-getintoteaching.education.gov.uk/"


MylesJarvis:
@ethax-ross can I just check that we've tested a record being created via the new URL? Before I move this over.

peteryates (approved):

ethax-ross:
@MylesJarvis I haven't but I can ask Ian if we can do a test run (as it'll go to prod)

MylesJarvis:
> @MylesJarvis I haven't but I can ask Ian if we can do a test run (as it'll go to prod)

Yes please @ethax-ross. If you could let me know when we're confirmed, we can roll this out. Converted to draft in the meantime.

ethax-ross (approved):

PR #1088, opened 2021-04-08
Bug fix check if public asset exists

ethax-ross:
I'm not yet sure why, but `File.exist?` is returning `false` for assets in `/public`. When I test locally `File.file?` returns correctly.

Updating to see if this also works in the hosted environments.


peteryates (approved):
Look good to me.

PR #1093, opened 2021-04-08
Correct ordering of picture sources

ethax-ross:
It turns out the order of the sources and img inside the `picture` element is important - we want to put the next-gen formats first and the `img` tag last, so that the browser correctly serves the next-gen image with priority over the jpeg/png.

I've put `webp` first as that seems to have the broadest support.


peteryates (approved):
Wow, learnt something new today!

PR #1098, opened 2021-04-09
Lazy load images correctly with Turbolinks

ethax-ross:
- Lazy load images correctly with Turbolinks

The lazy images weren't being correctly loaded on page change due to the library not being initialised on `turbolinks:load`.

Also improves the code around making the images visible once the `src` is set.

- Disable lazy loading header images

The images in the navigation don't benefit from being lazy loaded and there is a very slight flash when transitioning between pages because of the lazy loading. Its only noticeable on these images as they are present/in the same place on every page.

Cleans up some of the Nokogiri syntax to be consistent.

### DIFF
--- a/app/components/navigation/navbar_component.html.erb
+++ b/app/components/navigation/navbar_component.html.erb
@@ -11,12 +11,12 @@
             </ul>
 
             <div class="navbar__desktop__search">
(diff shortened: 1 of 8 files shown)


peteryates (approved):
Looks good to me

PR #1108, opened 2021-04-12
Set cookies SameSite attribute to 'Lax'

leoapost:
### Trello card
https://trello.com/c/3iFacmIh

### Context
Browsers will reject cookies which have set the SameSite attribute to "None", without also setting the Secure attribute. 

Set the SameSite to "Lax" instead.

### Guidance to review
Firefox shouldn't display the `Some cookies are misusing the recommended “SameSite“ attribute` warning in the console.

### DIFF
--- a/app/webpacker/javascript/cookie_preferences.js
+++ b/app/webpacker/javascript/cookie_preferences.js
@@ -41,6 +41,7 @@ export default class CookiePreferences {
     const serialized = JSON.stringify(categories);
     Cookies.set(CookiePreferences.cookieName, serialized, {
       expires: CookiePreferences.cookieLifetimeInDays,
+      sameSite: 'Lax',
     });
 
     this.cookieSet = true;


app/webpacker/javascript/cookie_preferences.js
```diff
@@ -41,6 +41,7 @@ export default class CookiePreferences {
     const serialized = JSON.stringify(categories);
     Cookies.set(CookiePreferences.cookieName, serialized, {
       expires: CookiePreferences.cookieLifetimeInDays,
+      sameSite: 'Lax',
```
ethax-ross, line 44: Can we update the tests to check for this? and is there a reason we're going `Lax` instead of `Secure`? I don't think we support http requests so can't see why we wouldn't be able to use `Secure`?

leoapost: We can definitely use SameSite 'None' with Secure, but I couldn't see why we would need it.

ethax-ross: Ah reading into it more I think a `Lax` is probably safer for us and `None` with `Secure` may break some of our tracking pixels... it also looks like we don't test the cookie attributes currently either so I'll leave it up to you if you want to add that in or not now.

ethax-ross (approved):

app/webpacker/javascript/cookie_preferences.js
leoapost: I believe we don't test the cookie attributes because they are not available to javascript. For example, `Cookies.get('_foo')` will return only the cookie values.

ethax-ross: Ah yes you're right - I had always assumed the attributes were part of the cookie that was available. I guess we could spy on `Cookies.set` but its probably more hassle than its worth...

PR #1112, opened 2021-04-12
Add includes polyfill for IE11

ethax-ross:
IE11 complains as `include` is not available; this polyfills it. At some point (or if there's many more of these!) I'll have a look at setting up webpacker to transpile JS so its compatible for IE.


peteryates (approved):
The saga continues...

PR #1118, opened 2021-04-13
Internal show

jkempster34:
# Trello card
https://trello.com/c/eDrH76ST

# Context
It was requested that the pending events should be able to be viewed as they appear on the main site, so they can be properly reviewed.

# Changes proposed in this pull request
- Add views for index and show
- Add request specs 
- Add actions for `index` and `show` to the internal events controller
- Do not show `pending` events on the main site when visiting them directly (they shouldn't be visible on index because there is a default filter on the API `get events` endpoint that returns `open` and `closed` events only)

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -25,7 +25,7 @@ class EventsController < ApplicationController
 
   def show
     @event = GetIntoTeachingApiClient::TeachingEventsApi.new.get_teaching_event(params[:id])
-    raise_not_found if @event.status_id == GetIntoTeachingApiClient::Constants::EVENT_STATUS["Pending"]
+    raise_not_found if @event.status_id == pending_event_status_id
 
     @page_title = @event.name
     @front_matter = { "description" => @event.summary }
@@ -79,5 +79,9 @@ private
 
     (params[Events::Search.model_name.param_key] || defaults)
       .permit(:type, :distance, :postcode, :month)
+  end
+
+  def pending_event_status_id
+    GetIntoTeachingApiClient::Constants::EVENT_STATUS["Pending"]
   end
 end
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -25,6 +25,7 @@ class EventsController < ApplicationController
 
   def show
     @event = GetIntoTeachingApiClient::TeachingEventsApi.new.get_teaching_event(params[:id])
+    raise_not_found if @event.status_id == GetIntoTeachingApiClient::Constants::EVENT_STATUS["Pending"]
 
     @page_title = @event.name
     @front_matter = { "description" => @event.summary }
--- a/app/controllers/internal/events_controller.rb
+++ b/app/controllers/internal/events_controller.rb
@@ -17,21 +17,21 @@ module Internal
   private
 
     def load_pending_events
-      opts = {
-        type_id: GetIntoTeachingApiClient::Constants::EVENT_TYPES["School or University event"],
-        status_ids: [pending_event_status_id],
-        quantity_per_type: 1_000,
-      }
-
       search_results = GetIntoTeachingApiClient::TeachingEventsApi
                          .new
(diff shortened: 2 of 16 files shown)


app/controllers/events_controller.rb
```diff
@@ -25,6 +25,7 @@ def search
 
   def show
     @event = GetIntoTeachingApiClient::TeachingEventsApi.new.get_teaching_event(params[:id])
+    raise_not_found if @event.status_id == GetIntoTeachingApiClient::Constants::EVENT_STATUS["Pending"]
```
ethax-ross, line 28: This reads nicer the way you have it in the internal controller with `pending_event_status_id`

app/controllers/internal/events_controller.rb
```diff
@@ -1,8 +1,40 @@
 module Internal
   class EventsController < ::InternalController
+    layout "internal"
+
     def index
-      # stubbed for now
-      head :ok
+      @no_results = load_pending_events.blank?
```
ethax-ross, line 6: I would move the `load_pending_events` to a `before_action` and have:

```
@no_results = @events.blank?
```

Otherwise someone may inadvertently remove the no results variable and not realise that the events will stop displaying

app/helpers/events_helper.rb
```diff
@@ -25,6 +25,10 @@ def event_status_open?(event)
     event.status_id == GetIntoTeachingApiClient::Constants::EVENT_STATUS["Open"]
   end
 
+  def event_status_pending?(event)
```
ethax-ross, line 28: I don't think this is used?

app/views/internal/_provider_event.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= link_to internal_event_path(id: provider_event.readable_id), class: "events-featured__list__item" do %>
```
ethax-ross, line 1: Should this partial just be `_event.html.erb`? I don't think we use the term provider event elsewhere?

app/views/internal/index.html.erb
```diff
@@ -0,0 +1,35 @@
+<div class="markdown">
+  <% if params[:success] %>
```
ethax-ross, line 2: I would avoid having the views access `params` directly and instead set `@success` and `@pending` instance vars in the controller

spec/requests/internal/events_controller_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+          assert_response :not_found
+          expect(response.body).to include "Page not found"
+        end
+      end
+    end
+  end
+
+private
+
+  def generate_auth_headers(user_type)
```
ethax-ross, line 85: Will this be reused in other specs? if so you could drop it into `spec/support/spec_helpers/internal.rb` and include it in `rails_helper`

spec/requests/internal/events_controller_spec.rb
```diff
@@ -0,0 +1,100 @@
+require "rails_helper"
+
+describe Internal::EventsController do
+  let(:types) { Events::Search.available_event_type_ids }
+  let(:events) do
+    2.times.collect do |index|
+      start_at = Time.zone.today.at_end_of_month - index.days
+      type_id = types[index % types.count]
```
ethax-ross, line 8: This is a bit convoluted... would it be clearer to just have:

```
let(:pending_event) { build() }
let(:open_event) { build() }
let(:events) { [pending_event, open_event] }
```

ethax-ross (commented):
Looks good, just a couple of things

app/helpers/events_helper.rb
jkempster34: I forgot a change here: https://github.com/DFE-Digital/get-into-teaching-app/pull/1118/commits/cd92bc621ec3129697242c9ceb27201bbbd4faec

It didn't look great having the banner showing on event pending event

ethax-ross (dismissed):
Looks good to me

ethax-ross (approved):

PR #1127, opened 2021-04-13
Internal create edit

jkempster34:
# Trello card
https://trello.com/c/eDrH76ST

# Context
There needs to be a way to create a pending event, edit a pending event, and approve a pending event (update its status to `open`)

# Changes proposed in this pull request
- Add controller actions and utilities for create/update
- Add view for new 
- Move all internal views into `events` folder to match controller
- Update request specs with the new actions
- Add features specs
- Add JavaScript to enable the gov styling, and to enable `accessible autocomplete` dropdown
- Add CSS styling to make form consistent


app/controllers/internal/events_controller.rb
```diff
@@ -14,8 +18,75 @@ def show
       @page_title = @event.name
     end
 
+    def new
+      @event = Event.new(venue_type: Event::VENUE_TYPES[:existing])
+      @event.building = EventBuilding.new
+    end
+
+    def approve
+      api_event = GetIntoTeachingApiClient::TeachingEventsApi.new.get_teaching_event(params[:format])
```
ethax-ross, line 27: Why is this `params[:format]`? I would expect it to be something like `params[:id]`

app/controllers/internal/events_controller.rb
```diff
@@ -14,8 +18,75 @@ def show
       @page_title = @event.name
     end
 
+    def new
+      @event = Event.new(venue_type: Event::VENUE_TYPES[:existing])
+      @event.building = EventBuilding.new
+    end
+
+    def approve
```
ethax-ross, line 26: Should this not check the role to ensure only a `publisher` is approving?

app/controllers/internal/events_controller.rb
```diff
@@ -14,8 +18,75 @@ def show
...
+    def edit
+      api_event = GetIntoTeachingApiClient::TeachingEventsApi.new.get_teaching_event(params[:id])
+      @event = Event.initialize_with_api_event(api_event)
+      render :new
+    end
+
   private
 
+    def submit_pending
+      return if invalid?
```
ethax-ross, line 52: I think this should explicitly `return false if invalid?`

app/controllers/internal/events_controller.rb
```diff
@@ -14,8 +18,75 @@ def show
...
+
+      @event.status_id = GetIntoTeachingApiClient::Constants::EVENT_STATUS["Pending"]
+      api_event = @event.to_api_event
+      submit_to_api(api_event)
+    end
+
+    def invalid?
+      invalid_building = @event.building.nil? ? false : @event.building.invalid?
```
ethax-ross, line 60: Can this not be encapsulated in the `Internal::Event` class so you can just call `@event.invalid?`?

app/controllers/internal/events_controller.rb
```diff
@@ -14,8 +18,75 @@ def show
...
+
+    def submit_to_api(body)
+      GetIntoTeachingApiClient::TeachingEventsApi.new.upsert_teaching_event(body)
+      true
+    rescue GetIntoTeachingApiClient::ApiError => e
+      @event.map_api_errors_to_attributes(e) if e.code == 400
+      false
+    end
+
+    def format_building(event_params)
```
ethax-ross, line 72: This method is doing a lot and its a bit hard to follow. I think it would benefit from refactoring into a couple of methods and I also think this should encapsulate this in the `Internal::Event` so that you can just do `Event.new(event_params, @buildings)` and it takes care of constructing the related building?

app/webpacker/packs/internal.js
```diff
@@ -0,0 +1,16 @@
+import { initAll } from 'govuk-frontend';
+import { enhanceSelectElement } from 'accessible-autocomplete';
+import 'trix';
+
+initAll();
+
+document.body.className = document.body.className
+  ? document.body.className + ' js-enabled'
```
ethax-ross, line 8: What's this used for?

spec/features/internal_spec.rb
```diff
@@ -0,0 +1,179 @@
...
+    navigate_to_new_submission
+
+    enter_valid_event_details
+
+    click_button "Submit for review"
+    expect(page).to have_text "Event submitted for review"
+  end
+
+  scenario "Edit a pending event with no building" do
+    allow_any_instance_of(GetIntoTeachingApiClient::TeachingEventsApi).to \
```
ethax-ross, line 39: It looks like these two allow calls are repeated, worth wrapping the scenarios in a context and putting them in a before block to dry it up?

ethax-ross (commented):
Given it a quick review - my minds a bit melted this afternoon I'll maybe do another pass tomorrow morning

app/webpacker/packs/internal.js
jkempster34: https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#add-the-javascript-file-to-your-html

For the GovUK JS to work

ethax-ross: Ah ok, may be worth adding a comment in that case; its the kind of thing that someone would likely remove at some point as theres no direct reference to it in the app

peteryates (approved):
Overall looks good to me.

`Event#save` looks a tiny bit 'wordy', not sure there's need to explicitly return `true` when `upsert_teaching_event` should return it, but it's not doing any harm.

PR #1129, opened 2021-04-14
Improve abbreviation usability

peteryates:
### Context and changes

Currently our abbreviations aren't usable for people on mobile. The way abbr tags usually work is that the content of the `title` attribute is displayed in a 'popover' window when hovering over an abbreviation. Mobile users don't have a cursor so can't really hover. They _can_ trigger the title to display by long pressing, on Android/Chrome at least, but the behaviour isn't exactly intuitive.

Here, when an abbr is clicked, we replace the content with the explanation. This should make the sentence more explicit without changing its meaning.

The regex that detects abbreviations to be abbreviated has been amended so it won't target those surrounded by parentheses, ensuring we don't end up with text like:

> Postgraduate Certificate in Education (Postgraduate Certificate in Education), lorem ipsum...

https://user-images.githubusercontent.com/128088/114697372-bf0b6080-9d15-11eb-80b0-224aeeda9810.mp4

### DIFF
--- a/app/views/layouts/content.html.erb
+++ b/app/views/layouts/content.html.erb
@@ -55,7 +55,7 @@
         </aside>
         <% end %>
 
-        <%= tag.article(class: article_classes(@front_matter)) do %>
+        <%= tag.article(class: article_classes(@front_matter), data: { controller: 'abbreviation' }) do %>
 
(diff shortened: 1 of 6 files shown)


leoapost (approved):
Looks good

PR #1134, opened 2021-04-15
Enhance the error summary component

leoapost:
### Trello card
https://trello.com/c/VOwsgTBe

### Context
By default, the browser scrolls the target into view. That results in presenting an input tag without context when users use the error summary anchors, because we display the label or legend right above it.

Using the govuk-frontend components will manually handle the error summary scrolling and will not hide the label or legend.


ethax-ross (approved):
Looks good and works well. Not an issue as such, but I wasn't sure why the new JS is in a separate pack and required into the application pack? 

I've always used packs such that the pack is required into the layout, but that's maybe just a convention in my own head 😄 if you did do it that way you could just include this pack in the registration layout though, which may be beneficial:

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/views/layouts/registration.html.erb

leoapost:
I wasn't sure if we are using the error summary component only in the registration page, so I thought to compile the custom JS in the application.js, which is loaded across the app. To be fair though, the docs do recommend to place custom business logic outside the packs folder - perhaps I could move it in the `webpacker/javascript/` folder?

As for the registration layout, it renders the 'head' section which also every other layout does, so putting a `javascript_pack_tag` in that section would still load it across the app. We could use `content_for` in the layout and yield it in the head section, but not sure if it worth it. Finally, correct me if I'm wrong, but webpacker will compile the JS in a separate file if we use the `javascript_pack_tag` so I am not sure there's any real benefit.

ethax-ross:
Yeah so we only use the error summary component on the registration pages. The `webpacker/javascript` folder is where we've tended to put other custom JS and then just imported it into the application pack.

I was thinking if you used `javascript_pack_tag` in the `registration` layout (the `head` layout already yields so you should be able to use `content_for(:head)`) then it means we're only loading that JS on the pages that need it and not every page. As I said though it's not a big deal, I don't feel strongly about it either way.

leoapost:
> Yeah so we only use the error summary component on the registration pages. The `webpacker/javascript` folder is where we've tended to put other custom JS and then just imported it into the application pack.
> 
> I was thinking if you used `javascript_pack_tag` in the `registration` layout (the `head` layout already yields so you should be able to use `content_for(:head)`) then it means we're only loading that JS on the pages that need it and not every page. As I said though it's not a big deal, I don't feel strongly about it either way.

In this case it makes sense to use the `javascript_pack_tag`.

I renamed the file to be more specific and included it in the registration layout.

ethax-ross (approved):
Looks good 👍

PR #1135, opened 2021-04-15
Log to Sentry when CircuitBrokenError is thrown in client Gem

jkempster34:
### Trello card
https://trello.com/c/Z9YPUZ5Q

### Context
The API client will raise a CircuitBrokenError if there are repeated API failures. See PR [here](https://github.com/DFE-Digital/get-into-teaching-api-ruby-client/pull/37)

When this happens, the error should be logged to Sentry

- Log to Sentry when CircuitBrokenError is raised in the client Gem

### Guidance to review
This was tested by going into the client Gem and raising a CircuitBrokenError on any request, see the Sentry log that was produced [here](https://sentry.io/organizations/dfe-bat/issues/2340553013/?project=5276949&query=is%3Aunresolved). The app still navigates to not available pages such as `/events/not-available` as it should, see [here](https://github.com/DFE-Digital/get-into-teaching-app/pull/947).

If this is a suitable solution, I will do the same in the TTA app.

### DIFF
--- a/app/controllers/concerns/circuit_breaker.rb
+++ b/app/controllers/concerns/circuit_breaker.rb
@@ -4,7 +4,10 @@ module CircuitBreaker
   extend ActiveSupport::Concern
 
   included do
-    rescue_from GetIntoTeachingApiClient::CircuitBrokenError, with: :redirect_to_not_available
+    rescue_from GetIntoTeachingApiClient::CircuitBrokenError do |exception|
+      Sentry.capture_exception(exception)
+      redirect_to_not_available
+    end
   end
 
 protected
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/circuit_breaker.rb
```diff
@@ -4,7 +4,10 @@ class NotAvailablePathMissingError < RuntimeError; end
   extend ActiveSupport::Concern
 
   included do
-    rescue_from GetIntoTeachingApiClient::CircuitBrokenError, with: :redirect_to_not_available
+    rescue_from GetIntoTeachingApiClient::CircuitBrokenError do |exception|
+      Sentry.capture_exception(exception)
```
ethax-ross, line 8: Can we update the spec to `expect(Sentry).to receive(:capture_exception).with(exception)`?

ethax-ross (approved):

PR #1147, opened 2021-04-19
Skip site wide authentication in internal section

jkempster34:
### Context
There are two implementations of HTTP Basic Authentication:
- Site wide which is used to protect the Dev and Test environments
- In the `internal` section (`/internal/events`). This section is accessible on the live site by permitted people only.

However, these don't work together when both active (which they will be on the Dev and Test environments)

- Skip site wide auth in internal controller
- Rename to `site_wide_authentication` to state more explicitly what it does
 
### Guidance to review
To see this working set the following in `.env.develpment`
```
HTTPAUTH_USERNAME = something
HTTPAUTH_PASSWORD = something
```

You will be prompted to authenticate when first accessing either section, or when swapping from one section to another, which I think is the behaviour we want?

### DIFF
--- a/app/controllers/internal_controller.rb
+++ b/app/controllers/internal_controller.rb
@@ -1,5 +1,6 @@
 class InternalController < ApplicationController
   before_action :authenticate
+  skip_before_action :site_wide_authentication
   helper_method :publisher?
 
 protected
@@ -20,11 +21,13 @@ private
 
       if username == publisher[:username] && password == publisher[:password]
         set_account_role(:publisher)
+        return true
       elsif username == author[:username] && password == author[:password]
         set_account_role(:author)
+        return true
       end
 
-      session[:role].present?
+      false
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/internal_controller.rb
```diff
@@ -16,15 +17,17 @@ def set_account_role(role_type)
 
   def authenticate
     authenticate_or_request_with_http_basic do |username, password|
-      return false if username.blank? || password.blank?
```
ethax-ross, line 19: How come we're losing this check? I think its a good idea to keep it in case the ENV vars are not set to avoid allowing a blank username/password pass authentication?

jkempster34: The method always returns false if the password and username don't match the environment variables

jkempster34: Oh I see, I'll put it back in

ethax-ross: But what if the environment variables are not set/therefore are `nil`/blank? would it match username `""` and password `""`?

ethax-ross (approved):

PR #1157, opened 2021-04-21
Add length validation to match CRM

jkempster34:
### Context
The fields of Teaching events and Teaching event buildings have maximum character count in the CRM. If an event/building field that exceed that character count is posted, the API will throw an error.

- Add maximum character counts to the field of events/buildings, so that bad events/buildings are not posted.


ethax-ross (approved):
Looks good, we should reflect these in the API validation as well at some point

PR #1160, opened 2021-04-22
Place holder actions

sfawcett123:
# [Monorepo Add Placeholders](https://trello.com/c/s2kKErpa/1425-explore-moving-to-a-monorepo)
Adding place holder actions into master, these will not be run , but to be able to develop monorepo they are required so that github knows there is an action with that name on the branch


peteryates (approved):
Makes sense

PR #1162, opened 2021-04-22
Add spacing beneath buttons in markdown paragraphs

peteryates:
This gives the button a little more breathing room and makes documents
flow more nicely.

Screenshot from 2021-04-22 15-42-20

Also improve the spacing around headers so there's more above than below. Old on left, new on right.

Screenshot from 2021-04-22 16-57-50


EmmaFrith (approved):

ethax-ross (approved):
Looks good

PR #1165, opened 2021-04-23
Alternate mailing list signup process

peteryates:
## Trello

https://trello.com/c/3HESZ1z6/1524-redesign-the-email-sign-up-form-register-your-interest-on-one-page

## Context and changes

The current mailing list signup wizard might be too long. The plan is to test a shorter version.

This PR implements a one page signup wizard. It duplicates much of the logic from the wizard because one will be deleted once the testing's over, so there's no need to share code between them.

* [x] Create a model, controller and basic signup form
* [x] Hook the form dropdowns up so they show real data from the API
* [x] Borrow the validation rules from wizard steps
* [x] Make the form submission submit data to the CRM API
* [x] Deal with the API response appropriately
  * [x] Show success if a matching record **doesn't exists**
  * [x] Show an extra verification page if a matching record **does exist**
* [x] Improve the routes, should be `mailing-list/signup` or similar
* [x] Add specs, maybe (not sure at the commitment level yet)
  * [x] Test the model and validation
  * [x] Test the controller and calling the API

### Final bits of tidying up

* [x] Link to privacy policy from the checkbox label
* [x] More specs covering methods in `MailingList::Signup`
* [x] Move postcode down to the bottom of the form
* [x] Ensure relevant values are present in unverified submission
* [x] Ensure the `candidate_id` and `qualification_id` are present in verified submission

## Progress

| Form | Form with [errors](https://github.com/DFE-Digital/get-into-teaching-app/pull/1165/files#diff-44438ce218f5287c58d0017f965d888715635d94280669896f75841fbd7b4cd7R184) |
| ------ | ------- |
| Screenshot from 2021-04-26 16-46-12  | Screenshot from 2021-04-26 16-47-43 |


app/models/mailing_list/signup.rb
```diff
@@ -0,0 +1,141 @@
+require "attribute_filter"
+
+module MailingList
+  class Signup
+    include ActiveModel::Model
+    include ActiveModel::Validations::Callbacks
+    include ::Wizard::ApiClientSupport
```
ethax-ross, line 7: Can be one for later, but I think we should rename this to something more generic like `Form::ApiClientSupport`

app/models/mailing_list/signup.rb
```diff
@@ -0,0 +1,141 @@
...
+    end
+
+    def strip_whitespace
+      email&.strip!
+      first_name&.strip!
+      last_name&.strip!
+    end
+
+    def convert_string_ids_to_integers
+      self.degree_status_id = degree_status_id.to_i
```
ethax-ross, line 133: If we uses `attribute` instead of `attr_accessor` we wouldn't need these as we can type it as `:integer`?

ethax-ross (dismissed):
Looks great -- couple of take it or leave it comments; I would perhaps exclude the route from prod before this is shipped though/until the remaining implementation is done? I know its not going to be indexed but you never know... 🙄

app/models/mailing_list/signup.rb
peteryates: Ah good call, will do.

MylesJarvis:
@peteryates I think we should perhaps keep the order the same as the current mailing list sign up, which leaves postcode to the end. Thoughts?

app/controllers/mailing_list_signups_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+    @signup = MailingList::Signup.new
+  end
+
+  def create
+    @signup = MailingList::Signup.new(new_signup_params)
+
+    if @signup.valid?
+      save_signup_to_session
+
+      return redirect_to(edit_mailing_list_signup_path) if @signup.already_signed_up?
```
ethax-ross, line 17: This is a bit misleading I think - `already_signed_up?` will return `true` if the candidate already exists, not if they exist _and_ they have already signed up to the mailing list

app/controllers/mailing_list_signups_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+    session.delete(SESSION_STORE_KEY)
+    redirect_to mailing_list_step_path(path)
+  end
+
+  def restore_signup_from_session
+    signup_data = session[SESSION_STORE_KEY]
+
+    if signup_data.nil?
+      Rails.logger.warn("tried to confirm identity with blank session")
+      return redirect_to(new_mailing_list_signup_path)
```
ethax-ross, line 59: Could this not result in odd bugs if we return a `redirect_to` the caller will continue to process the code after the call to `restore_signup_from_session` (so line 33 onward in this file)?

app/models/mailing_list/signup.rb
```diff
@@ -0,0 +1,195 @@
...
+
+    validate :timed_one_time_password_is_correct,
+             on: :verify
+
+    def degree_status_options
+      @degree_status_options ||= query_degree_status
+    end
+
+    def teaching_subjects
+      @teaching_subjects ||= [OpenStruct.new(id: nil, value: "Please select")] + query_teaching_subjects
```
ethax-ross, line 78: Not a big deal, but could avoid having to do this by using `prompt: { "Please select" }` in the view I think

app/models/mailing_list/signup.rb
```diff
@@ -0,0 +1,195 @@
...
+      GetIntoTeachingApiClient::MailingListApi.new
+    end
+
+    def timed_one_time_password_is_correct
+      request = GetIntoTeachingApiClient::ExistingCandidateRequest.new(identity_data)
+      response = api.exchange_access_token_for_mailing_list_add_member(timed_one_time_password, request)
+
+      Rails.logger.info("MailingList::Wizard#exchange_access_token: #{AttributeFilter.filtered_json(response)}")
+
+      response.tap do |r|
```
ethax-ross, line 186: I don't feel strongly about it, but this method is doing quite a lot and more than the method name suggests. I wonder if it would be clearer to split this and have an `after_validation :prepopulate` that sets the necessary ids etc from the totp response? You then have to track the response as an instance variable, so maybe what you have is more pragmatic 🤷‍♂️

spec/models/mailing_list/signup_spec.rb
```diff
@@ -0,0 +1,210 @@
...
+    describe "#degree_status_id" do
+      let(:degree_statuses) { GetIntoTeachingApiClient::PickListItemsApi.new.get_qualification_degree_status }
+      let(:message) { "Select your degree stage from the list" }
+
+      it { is_expected.to validate_presence_of(:degree_status_id).with_message(message) }
+
+      describe "values" do
+        specify "allows teaching subject values" do
+          degree_statuses.map(&:id).each do |ds|
+            expect(subject).to allow_value(ds).for(:degree_status_id)
```
ethax-ross, line 97: You can pass multiple values to these shoulda matchers with `allow_values` to make them more concise:

`it { is_expected.to allow_values(degree_statuses.map(&:id)).for(:degree_status_id) }`

spec/rails_helper.rb
```diff
@@ -8,6 +8,7 @@
 end
 
 require "rspec/rails"
+require "active_record"
```
ethax-ross, line 11: Why was this necessary?

ethax-ross (commented):
Looks good overall, just some minor questions/observations, mostly take it or leave it

app/controllers/mailing_list_signups_controller.rb
peteryates: Yeah - I named it when I didn't fully understand what was going on. Renamed to `#exists_in_crm?`.

app/controllers/mailing_list_signups_controller.rb
peteryates: Yep agree 🤦🏽 

Reworked with a VerifyWithEmptySession error

app/controllers/mailing_list_signups_controller.rb
ethax-ross: That kind of reads like the same thing when you call it on `@signup.exists_in_crm?` 😬 maybe something like `@signup.existing_candidate?`

spec/rails_helper.rb
peteryates: When switching from `attr_accessor` to `attribute` I'm getting errors from shoulda (originating [here](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/rails_shim.rb#L51)) matchers unless ActiveRecord is present.

ethax-ross: I think you're maybe just missing an `include ActiveModel::Attributes` on your signup class?

ethax-ross: Ah fair enough, weird the step models work but this one doesn't 🤔 I can't see any differences between them though 🤷

app/models/mailing_list/signup.rb
peteryates: Yeah I did try with a separate method but the logic became a bit hard to follow. I'll rename the method to add a bit more clarity

spec/rails_helper.rb
peteryates: Nor me, I spent ages comparing them side by side 🤷🏽‍♂️

ethax-ross (approved):
Looks good

ethax-ross (approved):

PR #1166, opened 2021-04-23
Update API cache store

jkempster34:
### Context

Currently, when a teaching event is posted to the API, it will not be immediately available on the hosted environments because of HTTP caching.

The cache will be cleared when events are posted in the API client Gem. However, we only want to clear the API cache, so here we update the API client cache store to use a new Redis namespace.

-Update the API client cache store

### Guidance to review
Linked to this [PR](https://github.com/DFE-Digital/get-into-teaching-api-ruby-client/pull/49)

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -66,4 +66,6 @@ Rails.application.configure do
     Rails.application.credentials.author_username.presence
   config.x.author_password = ENV["AUTHOR_PASSWORD"].presence || \
     Rails.application.credentials.author_password.presence
+
+  config.x.api_client_cache_store = ActiveSupport::Cache::MemoryStore.new
 end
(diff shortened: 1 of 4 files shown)


config/environments/development.rb
```diff
@@ -66,4 +66,6 @@
     Rails.application.credentials.author_username.presence
   config.x.author_password = ENV["AUTHOR_PASSWORD"].presence || \
     Rails.application.credentials.author_password.presence
+
+  config.x.api_client_cache_store = ActiveSupport::Cache::MemoryStore.new
```
ethax-ross, line 70: This is causing a spec to fail due to testing for `cache_read` on active support but its never called now that we've isolated this to a separate cache.

ethax-ross (approved):

PR #1168, opened 2021-04-26
Log when duplicate candidate is potentially being created

ethax-ross:
If the API responds with a non-404 error on the match back endpoint then we are potentially creating a duplicate Candidate record in the CRM (as we have been unable to verify against an existing Candidate record).

Add logging so that we can easily search for these occurrences.

Add logging so that we can easily identify requests for access tokens by transaction (mailing list/event sign up).


leoapost (approved):
Looks good

PR #1174, opened 2021-04-27
Building must be present if venue_type is :existing

ethax-ross:
We were not showing an error/catching the scenario when a user selects an existing venue type but does not pick a building from the dropdown (instead we created the event with no building, which has been flagged as confusing by the CRM team).

Adds validation to ensure a building with an id is present if the `venue_type` is `:existing`.

Minor clean up of other tests.

Screenshot 2021-04-27 at 11 00 21

### DIFF
--- a/app/views/internal/events/new.html.erb
+++ b/app/views/internal/events/new.html.erb
@@ -40,7 +40,7 @@
       <%= b.govuk_error_summary unless @event.building.nil? %>
       <%= f.govuk_radio_buttons_fieldset(:venue_type, legend: { size: 'm', text: 'Event venue' }) do %>
         <%= f.govuk_radio_button :venue_type, Internal::Event::VENUE_TYPES[:none], label: { text: 'No venue' }, link_errors: true %>
-        <%= f.govuk_radio_button :venue_type, Internal::Event::VENUE_TYPES[:existing], label: { text: 'Search existing venues' }, link_errors: true do %>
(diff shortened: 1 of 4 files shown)


app/views/internal/events/new.html.erb
```diff
@@ -39,8 +39,8 @@
     <%= f.fields_for :building, @event.building do |b| %>
       <%= b.govuk_error_summary unless @event.building.nil? %>
       <%= f.govuk_radio_buttons_fieldset(:venue_type, legend: { size: 'm', text: 'Event venue' }) do %>
-        <%= f.govuk_radio_button :venue_type, Internal::Event::VENUE_TYPES[:none], label: { text: 'No venue' } %>
-        <%= f.govuk_radio_button :venue_type, Internal::Event::VENUE_TYPES[:existing], label: { text: 'Search existing venues' } do %>
+        <%= f.govuk_radio_button :venue_type, Internal::Event::VENUE_TYPES[:none], label: { text: 'No venue' }, link_errors: true %>
+        <%= f.govuk_radio_button :venue_type, Internal::Event::VENUE_TYPES[:existing], label: { text: 'Search existing venues' }, link_errors: true do %>
```
peteryates, line 43: Only the first item in the collection should have `link_errors: true`.

```suggestion
        <%= f.govuk_radio_button :venue_type, Internal::Event::VENUE_TYPES[:existing], label: { text: 'Search existing venues' } do %>
```

peteryates (commented):
Looks good, just the `link_errors:` thing to adjust.

peteryates (approved):
👍🏽

PR #1175, opened 2021-04-27
Run specs in random order

ethax-ross:
At the moment if we hit a flakey spec in the CI pipeline we have no way to recreate the same test order locally in order to debug it.

By running in a random order RSpec will print out the `seed` value used for the run, which we can then use to perform an identical run locally.


sfawcett123 (approved):
Looking Good

PR #1186, opened 2021-04-29
Add a few more wildcard redirects

ethax-ross:
These appear to have been the most frequent 404s recently.

### DIFF
--- a/config/redirects.yml
+++ b/config/redirects.yml
@@ -287,7 +287,11 @@ redirects:
   "/life-as-a-teacher/real-life-experiences/(*story)": "/my-story-into-teaching"
   "/funding-and-salary(*section)": "/funding-your-training"
   "/regional-make-a-difference-in-(*area)": "/"
-
+  "/how-to-apply(*method)": "/tips-on-applying-for-teacher-training"
+  "/lp(*category)": "/"
+  "/frequently-asked-questions": "/"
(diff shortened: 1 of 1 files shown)


leoapost (approved):

config/redirects.yml
```diff
@@ -287,7 +287,11 @@ redirects:
   "/life-as-a-teacher/real-life-experiences/(*story)": "/my-story-into-teaching"
   "/funding-and-salary(*section)": "/funding-your-training"
   "/regional-make-a-difference-in-(*area)": "/"
-
+  "/how-to-apply(*method)": "/steps-to-become-a-teacher"
```
MylesJarvis, line 290: /tips-on-applying-for-teacher-training might be a better one here

ethax-ross: Same as well for this one? `/premier-plus-adviser(*tip)`

MylesJarvis: Think that one might be best suited as /three-things-to-help-you-get-into-teaching. I just used wayback machine to check its origins, it's solely focussed on TTAs

peteryates: Sounds like someone who [advises on hotel deals at Premier Inn](https://www.premierinn.com/gb/en/sleep/premier-plus.html) 😅

peteryates (approved):

PR #1201, opened 2021-05-04
Add a funding widget in the funding page

leoapost:
### Trello card
https://trello.com/c/NE9ALBAW

### Context
Currently, the users need to go through the funding page and read a lot of text to see what funding options they are eligible for the subject they are interested at.

Adding this simple widget at the top of the funding page, allows them to quickly see their funding options by simply choosing a subject. 

Create a view component to render a funding widget with a form to look up funding options per subject and display the results. The responsible controller method (pages#show) is not meant to support such interactivity, so this is not the most elegant implementation.  

This PR also includes Capybara configurations for running feature tests with `js: true` on headless chrome/chromium. The CI will be using chromium which has smaller footprint, but had to replace the `webdrivers` gem with the `selenium-webdriver` gem because the drivers the `webdrivers` gem is downloading are not working very well in Alpine.

### Guidance to review
| Before | After |
| --- | --- |
|funding-page-before|funding-page-after|


app/controllers/pages_controller.rb
```diff
@@ -37,6 +37,15 @@ def show
       breadcrumb page.title, page.path if @page.title.present?
     end
 
+    if @page.path.match?(/funding-your-training/)
```
ethax-ross, line 40: I would maybe call this out into its own route, feels wrong having it in the generic `pages#show` action

app/models/funding_widget.rb
```diff
@@ -0,0 +1,218 @@
+class FundingWidget
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  SUBJECTS = {
```
ethax-ross, line 5: Worth moving this to the translations file so its not polluting the model, then you can just have a set of keys here?

app/components/funding_widget_component.rb
```diff
@@ -0,0 +1,34 @@
...
+class FundingWidgetComponent < ViewComponent::Base
+  attr_reader :path, :form
+
+  def initialize(form, path)
+    @form = form
+    @path = path
+  end
+
+  def grouped_options
+    {}.tap do |options|
```
ethax-ross, line 12: Found this quite hard to follow... not sure if this is any better/if its right, but:

```
form.subjects.index_by { |_, v| v[:education] }.transform_values { |v| v.first }
```

app/views/sections/_head.html.erb
```diff
@@ -1,6 +1,10 @@
 <head>
   <meta charset="utf-8">
   <title><%= prefix_title page_title(@page_title, @front_matter) %></title>
+  <script type="module">
+    document.documentElement.classList.toggle('js-enabled');
```
ethax-ross, line 5: Can this go in to `/javascript` instead of being inline?

app/webpacker/styles/layout.scss
```diff
@@ -31,7 +31,7 @@ main,
   width: 1px;
 }
 
-.hidden {
+.js-enabled .hidden {
```
ethax-ross, line 34: Will changing this selector alter the behaviour of any other controllers that toggle this `hidden` class? (they may have their own selectors dealing with JS enabled/disabled?)

ethax-ross (commented):
Looks great, just a couple of comments/suggestions.

app/webpacker/styles/layout.scss
leoapost: No this is safe, it will not change the behaviour of any other controllers.

app/components/funding_widget_component.rb
leoapost: I've refactored this into an array and extracted it to smaller methods to make it simpler to understand.

ethax-ross (approved):
LGTM

ethax-ross (dismissed):

MylesJarvis:
To ship Monday 10 May

ethax-ross (dismissed):

MylesJarvis:
Now awaiting further review from policy colleagues

MylesJarvis (dismissed):
Testing approving it to see if review app builds - NOT TO MERGE

app/views/content/funding-your-training.md
```diff
@@ -16,13 +16,14 @@
...
+  before-content:
+    - content/funding-your-training/funding-widget
   calls_to_action:
     chat:
       name: chat_online
       arguments:
         text: |-
-          If you're unsure about whether you qualify for a bursary or scholarship
-          you can chat to us.
+          If you're unsure about whether you qualify for a bursary or scholarship you can chat to us, or call us on 0800 389 2500.
```
MylesJarvis, line 26: @peteryates here's the component in need of an <a href="tel:08003892500" class="telephone-number"
    aria-label="Telephone">0800 389 2500</a>

MylesJarvis:
@peteryates are you okay to resolve conflicts too?

peteryates (approved):
All looks fine to me.

The addition of JavaScript testing from our specs is an interesting one - the overlap between internal tests and the Cypress ones is increasing.

I'm not against it, just wonder whether we should look at combining them at some point.

MylesJarvis (approved):
LGTM

PR #1202, opened 2021-05-04
Make Myles Codeowner

sfawcett123:
## Change 
Add a new codeowners file with Myles name in it.


ethax-ross (approved):
Think we will want the other content editors in here too?

sfawcett123:
> Think we will want the other content editors in here too?

I am not sure, the codeowner basically is the mandatory reviewer.

PR #1214, opened 2021-05-05
Add button to generate partial URL

jkempster34:
### Trello card
https://trello.com/c/l8pD2kl7

### Context
The CRM team requested a way to generate partial URLs based on the event name and start date.

- Add button to generate partial URL

### Guidance to review
- I originally did this on name/date change. But I thought that could be annoying if you choose a custom name, then change name or date.
- Tested this by testing the individual methods, do I need to set up the HTML template like some of the other controllers instead/as well?
- Should this have some feedback when name or date is not set? Would this just be an error message (suppose I could just hide the button until they are set)?
- I replace whitespace with hyphens to match the other events, are there any other characters that shouldn't appear in a URL?

### DIFF
--- a/app/views/internal/events/new.html.erb
+++ b/app/views/internal/events/new.html.erb
@@ -13,7 +13,7 @@
   <%= f.hidden_field :id %>
   <%= f.govuk_text_field :name,
                          label: { text: 'Event name' },
-                         data: { "internal-events-target": "name"  } %>
+                         data: { "internal-events-target": "name" } %>
   <%= f.govuk_text_area :summary, label: { text: 'Event summary' } %>
   <%= render "form/govuk_error_message", label: "Event description", field_errors: @event.errors[:description] do %>
     <%= f.hidden_field :description, class: "event-description" %>
@@ -43,7 +43,11 @@
   <%= f.govuk_text_field :readable_id,
                          label: { text: 'Event partial URL' },
                          hint: { text: 'You can generate the partial URL from the event name and start at date' },
-                         data: { target: "internal-events.partialUrl" } %>
+                         data: { "internal-events-target": "partialUrl" } %>
+
+  <span class="govuk-error-message hidden" id="internal-generate-url-error" data-internal-events-target="errorMessage" }>
+    <span class="govuk-visually-hidden">Error: </span>Name and Start at fields must have values in order to generate partial URL
+  </span>
   <%= f.button "Generate partial URL",
                class: "button form-button",
                type: "button",
--- a/app/views/internal/events/new.html.erb
+++ b/app/views/internal/events/new.html.erb
@@ -45,7 +45,7 @@
                          hint: { text: 'You can generate the partial URL from the event name and start at date' },
                          data: { target: "internal-events.partialUrl" } %>
   <%= f.button "Generate partial URL",
-               class: "call-to-action-button button form-button",
+               class: "button form-button",
                data: { action: "click->internal-events#generateButtonClicked" } %>
 
   <div class="no-left-margin">
@@ -72,5 +72,5 @@
       <% end %>
     <% end %>
   </div>
-  <%= f.button "Submit for review", class: "call-to-action-button button form-button" %>
+  <%= f.button "Submit for review", class: "button form-button" %>
 <% end %>
--- a/app/views/internal/events/new.html.erb
+++ b/app/views/internal/events/new.html.erb
@@ -13,7 +13,7 @@
   <%= f.hidden_field :id %>
   <%= f.govuk_text_field :name,
                          label: { text: 'Event name' },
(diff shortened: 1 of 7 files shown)


app/webpacker/controllers/internal_events_controller.js
```diff
@@ -0,0 +1,47 @@
+import { Controller } from 'stimulus';
+
+export default class extends Controller {
+  static targets = ['name', 'startAt', 'partialUrl'];
+
+  get nameValue() {
```
ethax-ross, line 6: I wouldn't bother with these getters, its just a layer of indirection it doesn't add anything IMO

app/webpacker/controllers/internal_events_controller.js
```diff
@@ -0,0 +1,47 @@
...
+
+  get partialUrlValue() {
+    return this.partialUrlTarget.value;
+  }
+
+  set partialUrlValue(value) {
+    this.partialUrlTarget.value = value;
+  }
+
+  generateButtonClicked(event) {
```
ethax-ross, line 22: The function should describe what its doing, that way when you see it in the HTML its easier to understand, for example:

```
click->internal-events#generateButtonClicked

vs

click->internal-events#populatePartialUrl
```

app/views/internal/events/new.html.erb
```diff
@@ -27,14 +31,23 @@
...
     <%= f.datetime_field :end_at, class: class_names("datetime", "govuk-input",
                                                      { "govuk-input--error": @event.errors[:end_at].any? }) %>
   <% end %>
 
+  <%= f.govuk_text_field :readable_id,
+                         label: { text: 'Event partial URL' },
+                         hint: { text: 'You can generate the partial URL from the event name and start at date' },
+                         data: { target: "internal-events.partialUrl" } %>
+  <%= f.button "Generate partial URL",
+               class: "call-to-action-button button form-button",
```
peteryates, line 48: I think you should be ok with just `button` here:

```suggestion
               class: "button",
```

app/views/internal/events/new.html.erb
```diff
@@ -59,5 +72,5 @@
       <% end %>
     <% end %>
   </div>
-  <%= f.button "Submit for review", class: "call-to-action-button button" %>
+  <%= f.button "Submit for review", class: "call-to-action-button button form-button" %>
```
peteryates, line 75: And here

```suggestion
  <%= f.button "Submit for review", class: "button" %>
```

app/webpacker/controllers/internal_events_controller.js
```diff
@@ -0,0 +1,47 @@
...
+
+    return `${date}-${name}`;
+  }
+
+  formatDate(dateTimeString) {
+    if (dateTimeString === '') return '';
+
+    const date = dateTimeString.substring(0, dateTimeString.indexOf('T'));
+    const [year, month, day] = date.split('-');
+    return year.slice(-2) + month + day;
```
ethax-ross, line 41: This is pretty hard to follow - would it be cleaner to construct a new `Date` from the string and then output that in a format?

spec/javascript/controllers/internal_events_controller_spec.js
```diff
@@ -0,0 +1,69 @@
+import InternalEventsController from 'internal_events_controller.js';
+
+describe('InternalEventsController', () => {
+  let controller;
+
+  beforeEach(() => {
+    controller = new InternalEventsController();
```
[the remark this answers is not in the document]
ethax-ross: I would replace them or you'll end up doubling up; if you test the output of the controller actions by asserting against the HTML it will also let you refactor your methods without needing to update your tests, which is beneficial

app/webpacker/controllers/internal_events_controller.js
peteryates, line 41: Could this be simplified by using a date object? Something along the lines of:

```
const date = new Date(dateTimeString);
return date.toLocaleDateString();
```

peteryates: Ignore that, assumed JavaScript had some kind of built-in date formatter 🤦🏽‍♂️

jkempster34: Pete made the same comment, not sure if there is a way to do this.

peteryates: Yeah it's a bit of a faff because JavaScript's date functionality is pretty limited. [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString) has lots of options but it's a total faff compared to `#strftime`.

```suggestion
    const date = new Date(dateTimeString);
    return date.toLocaleString('en-GB', { '🤷🏽‍♂️' });
```

PR #1216, opened 2021-05-05
Add button to duplicate events

jkempster34:
### Trello card
https://trello.com/c/l8pD2kl7

### Context
The CRM team requested a way to duplicate events as a time saving measure. It is very common for multiple events to have the same/similar details.

- Add button to duplicate event

### Guidance to review
I wanted to add a features spec like this. It was a bit awkward though (I was having to turn off WebMock for a Capybara `__identify__` request), and I was getting a false positive so I abandoned it. 
 
Okay to rely on request spec only for this?

```ruby
  before do
    WebMock.allow_net_connect!
  end

  after do
    WebMock.disable_net_connect!
  end

  scenario "Duplicate event", js: true do
    navigate_to_existing_event

    new_tab = window_opened_by { click_button "Duplicate this event" }
    within_window new_tab do
      page.driver.browser.authorize("publisher", "password")
      expect(page).to have_field("Event partial URL", with: "hello")
    end
  end
end

```

### DIFF
--- a/app/controllers/internal/events_controller.rb
+++ b/app/controllers/internal/events_controller.rb
@@ -24,15 +24,13 @@ module Internal
     end
 
     def new
-      @event_type = determine_event_type(params[:event_type])
-      @event = Event.new(venue_type: Event::VENUE_TYPES[:existing], type_id: @event_type)
-      @event.building = EventBuilding.new
       if params[:duplicate]
         @event = get_event_by_id(params[:duplicate])
         @event.id = nil
         @event.readable_id = nil
       else
-        @event = Event.new(venue_type: Event::VENUE_TYPES[:existing])
+        @event_type = determine_event_type(params[:event_type])
+        @event = Event.new(venue_type: Event::VENUE_TYPES[:existing], type_id: @event_type)
         @event.building = EventBuilding.new
       end
     end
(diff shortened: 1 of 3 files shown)


leoapost (approved):
LGTM

The JS feature tests need a bit of extra configuration to work and we are working on it. So it should be ok for now to merge this PR and add the any other tests we need later.

app/controllers/internal/events_controller.rb
```diff
@@ -44,6 +44,11 @@ def create
     def edit
       api_event = GetIntoTeachingApiClient::TeachingEventsApi.new.get_teaching_event(params[:id])
       @event = Event.initialize_with_api_event(api_event)
+      if params[:duplicated]
```
ethax-ross, line 47: I think this is a bit counter-intuitive having it in the `edit` action when you're not editing, you're creating - I would have expected this in the `new` action, something like:

```
/events/new?duplicate=<event_id>
```

jkempster34: @ethax-ross Updated based on this ^

ethax-ross (approved):

peteryates (approved):
This looks ok to me.

I think generally using `#link_to` instead of `#button_to`/`method: :get` is the way to go, but as this is all internal stuff we can revisit that later.

PR #1220, opened 2021-05-06
Run specs in parallel

ethax-ross:
Parallelise test suite across worker cores - the GitHub Action workers only have 2 CPU cores so we don't get a massive benefit, but it brings the test run time down from 3 minutes to around 2 minutes.

I tried a few different things in this PR and I've left the earlier commits in place in case we want to revisit them, but to summarise:

- Enable parallel tests with parallel_tests gem

This was the least improvement speed-wise; it splits the spec files up and runs them on separate CPU cores. As we have a few spec files that contain the longest-running tests it didn't make a great improvement.

- Parallel tests over multiple workers

Ran the tests across multiple GitHub Action workers and cores using `parallel_tests` and a custom rake task. This was the fastest, but there was more work to do in order to get Sonarcube working with the fragmented test output so I don't think its worth pursuing (got the test run down to around 1m 30s).

- Enable parallel tests with parallel_split_tests gem

This was the best trade-off - simple change that just works, splits individual tests (not just spec files) across CPU cores. Drops about a minute from the test runtime.


sfawcett123 (approved):

peteryates (approved):
Nice!

PR #1221, opened 2021-05-06
Delete detailed finance guidance page and update redirects

MylesJarvis:
### Trello card
N/A
### Context
We no longer link to this page because of the fact that we've lifted content higher up on the site.

I've deleted the page, and updated redirects that pointed to this page to go to /funding-your-training

### Guidance to review
Have I caught all redirects? Is the page deleted?


peteryates (approved):
👍🏽

PR #1239, opened 2021-05-10
Fix early exit from http_basic block

jkempster34:
### Context
Bug fix - using `return` in the `authenticate_or_request_with_http_basic` block is actually returning from the enclosing `authenticate` method. This is causing some unusual behaviour.

- Fix early exit from http_basic block


ethax-ross (approved):
LGTM

PR #1245, opened 2021-05-10
Make use of quotes consistent across the site

ethax-ross:
### Trello card

[Trello-1541](https://trello.com/c/EEn4kHgh/1541-design-create-module-for-a-teacher-photo-headshot-and-quote-to-intersperse-through-the-site)

### Context

We have multiple ways that we display blockquotes on the website currently (using markdown syntax and straight up HTML). We want to consolidate these approaches to use the new `QuoteComponent`.

- Rename ComponentInjector to CallToActionComponentInjector

The `CallToActionComponentInjector` is quite specific and more complex than the other component injectors will need to be (as there are multiple types of CTA components).

Renaming to allow a new `ComponentInjector` to take care of the more basic component instantiation from frontmatter/placeholders.

- Update QuoteComponent to support frontmatter

As we are injecting the component via frontmatter we can't rely on symbols and instead need to use strings for `hang` and `inline` (`inline` now also supports left/right instead of just true/false).

- Update quote markdown to component

Instead of using the markdown quote syntax we want to use the `QuoteComponent` for consistency.

Update all references to quotes to use the new component placeholder/injection method.

Remove unused blockquote CSS.

- Support injecting QuoteComponent into markdown

### Guidance to review

The pages with quotes on them currently are day in the life of and the stories; I've checked them all manually and they _seem_ fine with the new component, but worth someone else eyeballing a few of them too.


peteryates (dismissed):
Nice 😎

MylesJarvis (approved):
Covers all stories from a search through - lgtm

PR #1248, opened 2021-05-10
Get events for index from parameter

jkempster34:
### Trello card
https://trello.com/c/l8pD2kl7

### Context
We want to be able to view different event types. This will allow events to be navigated to by page/tab.

- Get events for index from URL parameter


peteryates (approved):
Looks good to me.

PR #1249, opened 2021-05-11
Update Fail2Ban settings

ethax-ross:
[Trello-1568](https://trello.com/c/2fmvfC4A/1568-look-to-increasing-fail2ban-time-for-a-longer-period)

- Update Fail2Ban length to 5 minutes

- Send Fail2Ban violations to the logs

We are receiving a fair amount of exceptions due to Fail2Ban being triggered; its not an exceptional case and shouldn't be treated as an error, so sending these to the logs and setting up a Grafana dashboard to monitor them feels more appropriate.


leoapost (approved):
LGTM

PR #1255, opened 2021-05-11
Revert "Revert "Move mailing list promotion bar outside of footer""

MylesJarvis:
Reverts DFE-Digital/get-into-teaching-app#1253 - This is effectively a revert of a revert. This puts the new and updates green banner into the codebase.

Deploys the Mailing List promotion work - this will need GA tags on it along with other interventions.


peteryates (approved):
Looks good

PR #1263, opened 2021-05-12
Add tabbed navigation

jkempster34:
### Trello card
https://trello.com/c/l8pD2kl7

### Context
The CRM requested the ability to add Scribble Events (a type of online event).

This will enable users to switch between provider and online events. The form will be updated to reflect this in a later PR

- Add tabbed navigation between provider and online event types.

### Guidance to review
- Does it look passable?

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,12 +1,12 @@
 GIT
   remote: https://github.com/DFE-Digital/get-into-teaching-api-ruby-client.git
-  revision: 9e1ee278207b6c3499ca969c8d2e10bc2bb30a97
+  revision: c0bf769898ceabbaaa74790381a44c1e316e4b60
   specs:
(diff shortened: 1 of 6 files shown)


peteryates (approved):
Looks ok to me. I pushed a change that limits the CSS to the internal layout.

I think there are a couple of little bits that could do with a tidy, but no show stoppers. Examples include using `button_to` with `method: :get` where a `link_to` would suffice.

Nothing that can't be improved incrementally later though. Nice.

peteryates (changes requested):
Everything looks fine but I'm getting into a basic auth loop when attempting to edit an event, both locally and on the review app. Will re-approve if it's working fine and I'm just doing something wrong.

ethax-ross (approved):

PR #1264, opened 2021-05-12
Add ids to mailing list bar for tracking purposes

peteryates:
### Trello card

N/A

### Context

The IDs are to make interactions trackable via GA.


ethax-ross (approved):
Worth adding a spec to ensure nobody accidentally deletes the selector in the future, not realising its linked to GTM?

ethax-ross (approved):

PR #1266, opened 2021-05-12
Updated Internship Providers - 12 May

MylesJarvis:
### Context
We've received new teaching internship providers, this PR adds them in.

As per email from team - needs consulting to check. 

### Guidance to review
Are they in the right place? Are they accurate?

### DIFF
--- a/app/views/content/teaching-internship-providers.md
+++ b/app/views/content/teaching-internship-providers.md
@@ -15,6 +15,10 @@ providers:
       link: "https://www.beauchampteach.org.uk/opportunities/internships/"
       name: Emma Lowe
       email: Emma.Lowe@beauchamp.org.uk
+    - header: Bluecoat Aspley Academy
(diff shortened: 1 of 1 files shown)


app/views/content/teaching-internship-providers.md
```diff
@@ -48,10 +52,23 @@ providers:
       link: "https://www.setsa.info/1569/news-1/post/20/teaching-internship-2021"
       name: Emma Lowne
       email: e.lowne@setsa.info
+    - header: The Becket School
```
HettieS, line 55: have you put Becket in East of England instead of East Midlands for a reason? Is the excel spreadsheet wrong?

HettieS (changes requested):

app/views/content/teaching-internship-providers.md
MylesJarvis: Nope, that's an error. ~ood catch.

HettieS (approved):
LGTM!

PR #1268, opened 2021-05-13
Amending to new TPUK opening time

MylesJarvis:
### Context
TPUK have extended their calling times, updating this time from 5pm to 5.30pm

### Guidance to review
Have I caught all instances in the app. TTA still to do.


HettieS (approved):
LGTM

PR #1269, opened 2021-05-13
Rename telephone to address_telephone

ethax-ross:
The API currently supports both field names, but `telephone` is being deprecated in favour of `address_telephone`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,12 +1,12 @@
 GIT
   remote: https://github.com/DFE-Digital/get-into-teaching-api-ruby-client.git
-  revision: fdcee355185dfb8981f44e92aaa9cb24a43312e1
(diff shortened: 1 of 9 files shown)


app/webpacker/controllers/mailing-list-popup_controller.js
```diff
@@ -9,6 +9,7 @@ export default class extends Controller {
   topScrollMinimumDistance = 700;
 
   connect() {
+    console.log("HERE")
```
leoapost, line 12: Probably we don't need this one?

ethax-ross: Oh whoops, left that in from a previous PR I was debugging! Nice spot, I'll remove it now

leoapost (approved):

PR #1273, opened 2021-05-14
remove rogue setence

EmmaFrith:
This sentence doesn't seem to be adding much. We're already saying lots of things in that area like 'talk to us', and 'chat online' - so it's already clear we're help to help the user


peteryates (approved):
I agree it's not particularly useful. Worth getting a consensus on it though.

MylesJarvis:
I agree @EmmaFrith, just compared current site and review app side by side and feels superfluous. Happy for you to merge.

PR #1274, opened 2021-05-14
give taster of what steps covers in banner

EmmaFrith:
.. and make button text consistent with next page title text

### DIFF
--- a/app/views/content/home/_purple-box.html
+++ b/app/views/content/home/_purple-box.html
@@ -7,7 +7,7 @@
   </h2>
 
   <p class="steps-to-become-a-teacher__text">
-    Find out what qualifications you need, what funding you can get and how to train to teach.
+    Find out which qualifications you need, what funding you can get and how to train to teach.
   </p>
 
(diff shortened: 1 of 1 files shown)


app/views/content/home/_purple-box.html
```diff
@@ -3,15 +3,15 @@
   </div>
 
   <h2 class="steps-to-become-a-teacher__heading">
-    Steps to get into teaching.
+    Steps to become a teacher.
   </h2>
 
   <p class="steps-to-become-a-teacher__text">
-    Find out how to become a qualified teacher in England.
+    Find out what qualifications you need, what funding you can get and how to train to teach.
```
peteryates, line 10: The word 'what' being the start of two consecutive things in the list feels a bit funny here. Maybe

```suggestion
    Find out which qualifications you need, what funding you can get and how to train to teach.
```

EmmaFrith: like it

MylesJarvis (approved):
Tasty!

PR #1275, opened 2021-05-14
Allow non-content pages to be searched

peteryates:
The searching mechanism only pays attention to two fields from the frontmatter, the title and keywords. Here we're prepending a list of custom pages to the automatically-loaded frontmatter that will be included in search results.


MylesJarvis (approved):
LGTM

PR #1276, opened 2021-05-14
Ensure chevron applied before external link span

peteryates:
If these happen the other way round the chevron is added to the hidden external link text and isn't displayed.


MylesJarvis (approved):
Fixes the 2 examples fo the bug I found

PR #1288, opened 2021-05-17
Scribble form

jkempster34:
### Trello card
https://trello.com/c/l8pD2kl7

### Context
The CRM requested the ability to add Scribble Events (online events).
This PR adds a "new" form for Scribble events in addition to the Provider events form.

- Conditionally validate and render the event fields based on event type.
- Pass event type as query string to new form
- Update `events/show` view to render section if event pending. Whoever is reviewing the events must be able to accurately assess how they will appear when in an open state.
- Default optional fields to `nil`. If fields are omitted from the API request, they will not be changed in the CRM. For example, this prevents a `Scribble` event from omitting a `provider` field which would lead to it retaining its old value. This shouldn't happen anyway because it isn't possible to change an event type.
- Update some of the wording to work with either event type to reduce logic in the `internal/show` view

### DIFF
--- a/app/controllers/internal/events_controller.rb
+++ b/app/controllers/internal/events_controller.rb
@@ -7,7 +7,7 @@ module Internal
     DEFAULT_EVENT_TYPE = "provider".freeze
 
     def index
-      @event_type = set_event_type(params[:event_type])
+      @event_type = determine_event_type(params[:event_type])
 
       load_pending_events(@event_type)
       @no_results = @events.blank?
@@ -24,7 +24,7 @@ module Internal
     end
 
     def new
-      @event_type = set_event_type(params[:event_type])
+      @event_type = determine_event_type(params[:event_type])
       @event = Event.new(venue_type: Event::VENUE_TYPES[:existing], type_id: @event_type)
       @event.building = EventBuilding.new
     end
@@ -55,7 +55,7 @@ module Internal
 
   private
 
-    def set_event_type(type)
+    def determine_event_type(type)
       event_types[type] || event_types[:provider]
     end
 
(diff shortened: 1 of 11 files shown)


app/views/internal/events/new.html.erb
```diff
@@ -21,15 +22,20 @@
...
+
+  <% if @event.provider_event? %>
+    <%= f.govuk_email_field :provider_contact_email, label: { text: 'Provider email address' } %>
+    <%= f.govuk_text_field :provider_organiser, label: { text: 'Provider organiser' } %>
+    <%= f.govuk_text_field :provider_target_audience, label: { text: 'Target audience' } %>
+    <%= f.govuk_url_field :provider_website_url, label: { text: 'Provider website/registration link' } %>
+
+    <%= f.govuk_radio_buttons_fieldset(:is_online, inline: true,
+                                       legend: { size: 'm', text: 'Online event?' }) do %>
+      <%= f.govuk_radio_button :is_online, true, label: { text: 'Yes' } %>
```
peteryates, line 34:
```suggestion
      <%= f.govuk_radio_button :is_online, true, label: { text: 'Yes' }, link_errors: true %>
```

peteryates (commented):
Looks fine overall, needs a `link_errros` on the radios fieldset.

app/controllers/internal/events_controller.rb
```diff
@@ -54,6 +55,10 @@ def edit
 
   private
 
+    def set_event_type(type)
```
ethax-ross, line 58: First time I read this I thought the method was going to assign something. How about something like:

```
def determine_event_type
  event_types[params.fetch(:event_type, :provider)]
end
```

ethax-ross (approved):
Just one minor comment

ethax-ross (approved):

PR #1289, opened 2021-05-17
add link to page on getting support if you're disabled

EmmaFrith:
* shorten headings so important info is closer to the front 
* add link to page on getting support if you're disabled
* make text on funding if you're disabled consistent with that on page about getting support if you're disabled

PR #1290, opened 2021-05-18
style edits across funding and salaries

EmmaFrith:
mainly adding more spacing, breaking text into bullets or adding extra headings 
putting early career framework in lower case in line with gov.uk
add a call to action to 'Ways to train' from Salaries as it feels that there's no clear way to go after that page


MylesJarvis:
@peteryates I think this needs a rebase following merging of #1289

peteryates (approved):

MylesJarvis (approved):

PR #1296, opened 2021-05-18
Add a single question survey in the event signup exit page

leoapost:
### Trello card
https://trello.com/c/koVyEO7I/675-development-amend-the-exit-page-of-the-events-sign-up-to-include-a-where-did-you-hear-about-this-event

### Context
We want an option to get feedback from users on how do they hear about the events they are singing up for. However, we don't want it to be very distracting for the users, so feels appropriate for now to add it at the exit page of the Events sign up.

Create a view component to render single question surveys. The component accepts two mandatory arguments for the question and answers and an optional hint to add helpful instructions.


ethax-ross (dismissed):
Looks good. I think `page_helpful_controller` needs to be renamed to something more generic though like `feedback_controller` so that it applies to both use cases.

spec/javascript/controllers/feedback_controller_spec.js
```diff
@@ -1,5 +1,5 @@
 import { Application } from 'stimulus';
-import PageHelpfulController from 'page_helpful_controller';
+import PageHelpfulController from 'feedback_controller';
 
 describe('PageHelpfulController', () => {
```
ethax-ross, line 4: Missed a few references to `PageHelpfulController` here

leoapost: Good catch, I've updated the spec file.

ethax-ross (approved):

MylesJarvis (approved):

PR #1298, opened 2021-05-19
more specificity on Steps

EmmaFrith:
* make it clearer who should click on the link to the undergraduate content
* prevent 'false pessimism' about the need for certain GCSEs, following BAT/GIT data sharing UR playback 
* tweak language so it's more geared to people who are studying for a degree 
* remove repetition of 'GCSE grade 4 (C) or above' through bullet lead-in sentence
* move undergrad link to the bottom as we think this content doesn't apply to a large part of our audience
* link from the "if" statement rather than "undergraduate teacher training" so it's more screen reader friendly and so there's less potential to click there when it doesn't apply to you

Screenshot 2021-05-19 at 12 04 53

### DIFF
--- a/app/views/content/steps-to-become-a-teacher/_step_2_check_your_qualifications.html.erb
+++ b/app/views/content/steps-to-become-a-teacher/_step_2_check_your_qualifications.html.erb
@@ -1,15 +1,15 @@
-<p>If you already have a degree, or equivalent qualification, you’re ready for postgraduate primary or secondary initial teacher training.</p>
+<p>Once you have a degree or equivalent qualification you’re ready for postgraduate primary or secondary initial teacher training.</p>
 
-<p>You also need the following or equivalents:</p>
+<p>You also need the following or equivalents, or to show you can meet the same standard:</p>
 
 <ul>
   <li>English GCSE grade 4 (C) or above</li>
   <li>maths GCSE grade 4 (C) or above</li>
 </ul>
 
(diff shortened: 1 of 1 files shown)


peteryates (dismissed):
I like it but am concious that we're mentioning 'equivalents' in each of the first three paragraphs

EmmaFrith:
> I like it but am concious that we're mentioning 'equivalents' in each of the first three paragraphs

we could probably remove it from the 3rd para? maybe the message comes across enough already by that point?

app/views/content/steps-to-become-a-teacher/_step_2_check_your_qualifications.html.erb
```diff
@@ -1,16 +1,17 @@
...
 <ul>
-  <li>English GCSE grade 4 (C) or above</li>
-  <li>maths GCSE grade 4 (C) or above</li>
+  <li>English</li>
+  <li>maths</li>
+  <li>science if you want to teach primary</li>
 </ul>
 
-<p>You also need a science GCSE grade 4 (C) or above to teach primary.</p>
+<p>You may be able to show you meet the standard in another way if you do not have these GCSEs or equivalents.</p>
```
EmmaFrith, line 11:
```suggestion
<p>You may be able to show you meet the standard in another way if you do not have these GCSEs.</p>
```

EmmaFrith:
@peteryates - have removed the 3rd 'or equivalent'

MylesJarvis (approved):
LGTM - this should be clearer

PR #1302, opened 2021-05-20
style changes

EmmaFrith:
* make calls to action more balanced by taking out unnessary words: 

before:
Screenshot 2021-05-20 at 10 22 49

after:
Screenshot 2021-05-20 at 10 23 28

* make link text accessible in places
* remove repeated call to action on Funding page

### DIFF
--- a/app/views/content/funding-your-training.md
+++ b/app/views/content/funding-your-training.md
@@ -42,8 +42,8 @@
       name: simple
       arguments:
         icon: "icon-arrow"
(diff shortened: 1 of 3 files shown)


peteryates (approved):
Another **definite** improvement and think it should go in.

Some general observations on this section in general (not to be fixed now):

* the buttons _almost_ match the headers - do we need both? Could the one or the other be adjusted slightly so it's not so repetitive?
* 'Talk to us' and 'Chat online' are really close and mean more-or-less the same thing

MylesJarvis:
We've done all of these in a new PR

PR #1307, opened 2021-05-21
Rewrite and restyle ways to train page

peteryates:
This is an experimental look at rewriting the ways to train page and using something other than tables to represent the attributes (pros and cons) of each avenue.

Some of the markup is a bit untidy so if we go with it, it's probably worth reimplementing in a neater manner.

## Attempt one - placing the tags beneath the headings

Here the tags are placed right under the appropriate heading. The main drawback of this approach is that policy would prefer their _keyword_, i.e. **School direct (fee-funded)**  to be alongside the tag. As the new content has removed these terms we might be failing on that count.

Screenshot from 2021-05-18 17-00-49

## Attempt two

This time we reintroduce the policy-approived terms alongside the tags.

| A section covering one option | A section covering multiple options |
| ------- | ------- |
| Screenshot from 2021-05-21 10-19-52 | Screenshot from 2021-05-21 10-24-08 |


peteryates:
Added some light dividing lines between the 'rows' of the definition list

| Desktop | Mobile |
| ------ | ------ |
| Screenshot from 2021-05-25 16-12-54 | Screenshot from 2021-05-25 16-12-45 |

app/views/content/ways-to-train.md
```diff
@@ -57,7 +57,7 @@ You need 'qualified teacher status' (QTS) to work in the majority of schools in
 
 Find out how to get QTS if you:
 
-* [have a degree](#if-you-have-a-degree) (or you're doing one)
+* [have a degree](#if-you-have-a-degree) (or you're getting one)
```
peteryates, line 60: definitely better!

'doing one' sounds like legging it 🏃🏽‍♂️

MylesJarvis:
image
@peteryates  we've got the classic external site tag appearing in a button again 😩

peteryates:
Also there should be space between those tags - which browser are you using? I thought even Safari supported `gap` now

app/views/content/ways-to-train.md
```diff
@@ -42,208 +51,150 @@ keywords:
...
-A number of universities and School Centred Initial Teacher Training programmes (SCITTs) offer assessment only routes. Their criteria for entry may vary.
+## If you are already an experienced teacher
 
-You can apply to them directly.
+You may be able to get QTS through a 12-week 'assessment only' programme if you also have a degree.
 
-<p class="call-to-action__action">
-  <a href="/assessment-only-providers">See assessment only <span>providers</span></a>
-</p>  
+You’ll need to show evidence that you already meet the standards for QTS.
```
MylesJarvis, line 185: @EmmaFrith, what do you reckon to adding another line on the 'typical' requirements here for AO, e.g. X years experience, working in 2 schools?

MylesJarvis:
Ah, this was me reviewing it on my DfE device on Edge...

MylesJarvis:
> Also there should be space between those tags - which browser are you using? I thought even Safari supported `gap` now

@peteryates Ah, this was me reviewing it on my DfE device on Edge...

gazcarr:
I think we may need to lose the rounded corners on the tags as well. Usually, design patterns go with one or the other and our branding is straight/non-rounded

EmmaFrith:
> I think we may need to lose the rounded corners on the tags as well. Usually, design patterns go with one or the other and our branding is straight/non-rounded )border-radius: 0;
> 
> I don't think there is any need to bold these items either, they are prominent enough with the background colours. font-weight: normal

won't this make them look too much like buttons tho @gazcarr ?

app/views/content/ways-to-train.md
EmmaFrith: yus deffo. need to pull something together still

EmmaFrith: added a commit for this

gazcarr:
> > I think we may need to lose the rounded corners on the tags as well. Usually, design patterns go with one or the other and our branding is straight/non-rounded )border-radius: 0;
> > I don't think there is any need to bold these items either, they are prominent enough with the background colours. font-weight: normal
> 
> won't this make them look too much like buttons tho @gazcarr ?

I would say that is subjective, I always go back to UR in these scenarios - is a user less likely to try clicking on this because of its corners, probably not... therefore go with the rest of the design. I am happy to be challenged and ultimately its only something like hotjar testing that might tell us over time.

Its also one of the reasons I advised reduce the font-weight, all of our current buttons are using **bold** .

Happy to be challenged on any of these thoughts though.

EmmaFrith:
> > > I think we may need to lose the rounded corners on the tags as well. Usually, design patterns go with one or the other and our branding is straight/non-rounded )border-radius: 0;
> > > I don't think there is any need to bold these items either, they are prominent enough with the background colours. font-weight: normal
> > 
> > 
> > won't this make them look too much like buttons tho @gazcarr ?
> 
> I would say that is subjective, I always go back to UR in these scenarios - is a user less likely to try clicking on this because of its corners, probably not... therefore go with the rest of the design. I am happy to be challenged and ultimately its only something like hotjar testing that might tell us over time.
> 
> Its also one of the reasons I advised reduce the font-weight, all of our current buttons are using **bold** .
> 
> Happy to be challenged on any of these thoughts though.

your call @gazcarr - feel free to agree on a style with @peteryates

MylesJarvis (approved):

PR #1314, opened 2021-05-24
Use the event summary for meta descriptions

peteryates:
### Trello card

https://trello.com/c/AmBOVnD3/1484-improve-setting-meta-descriptions-for-events

### Context

The individual event summary should form its meta description. This will make it dynamic and more-descriptive.


peteryates:
Actually this is already set in the controller 🤦🏽

PR #1327, opened 2021-05-27
Separate international trainees and teachers into 2 separate pages 

EmmaFrith:
Separating international candidates page into 2 separate pages:
* Train to teach in England as an international student
* Come to England to teach if you're a teacher from outside the UK

This means that:

* each page has a clear focus, centred around an action (either training to teach as an international student, or coming to England to teach if you're already a teacher)
* we can use more descriptive and specific page titles with SEO friendly language
* we can have more specific places to link to on certain topics 
* we don't have to be bound by user groups such as 'internationals' if it doesn't centre around the action they're trying to complete


MylesJarvis (approved):
LGTM

PR #1330, opened 2021-05-28
make banner content consistent with newsletter content

EmmaFrith:
use some of the same language in banner and newsletter to create a more coherent ux

Screenshot 2021-05-28 at 15 02 16

Screenshot 2021-05-28 at 15 02 53

Screenshot 2021-05-28 at 15 03 33

### DIFF
--- a/app/views/content/home.md
+++ b/app/views/content/home.md
@@ -1,6 +1,6 @@
 ---
   layout: "layouts/home"
-  title: "Inspire the next generation"
+  title: "Get personalised guidance to your inbox"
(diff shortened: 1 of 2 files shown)


peteryates (approved):
I like "one step closer to the classroom", that fits in really well.

"Sign up" is also a better description of what's involved.

I'm on board with this.

MylesJarvis:
This is going to be overtaken by some new promotion work as we progress towards end of cycle - we're going to use the hero to promote finding a course and TTA

PR #1342, opened 2021-06-01
Ensure we're only using one font family throughout

peteryates:
### Trello card

https://trello.com/c/epj0hvM7/1608-ensure-were-using-the-same-font-throughout

### Context and changes

This attempts to solve the problem in two ways:

#### Move git-font-family declaration above git import

Previously the $git-font-family declaration was made _after_ the `@import './git'`, which meant the setting wouldn't be applied everywhere. Now the font is defined right at the top of `application.scss` so it should be clear that it's global.

#### Separate out the `internal` and `application` stylesheets

In addition to the above, we can ensure the full import of the govuk-frontend styled don't have any impact on the rest of the site by serving another stylesheet for the internal section.

This will have the added benefit of making the page a bit lighter for almost every user.


jkempster34 (approved):
Had a browse through the internal section on the review app. All seems to be working now 👍

PR #1344, opened 2021-06-01
Set `start_at` and `end_at` to nil on duplicate

jkempster34:
### Context
CRM team requested for `start_at` and `end_at` attributes to be emptied when duplicating an event

- Set `start_at` and `end_at` to nil on duplicate

### DIFF
--- a/app/controllers/internal/events_controller.rb
+++ b/app/controllers/internal/events_controller.rb
@@ -5,6 +5,7 @@ module Internal
     helper_method :event_type_name
 
     DEFAULT_EVENT_TYPE = "provider".freeze
+    NILIFY_ON_DUPLICATE = %i[id readable_id start_at end_at].freeze
 
     def index
       @event_type = determine_event_type(params[:event_type])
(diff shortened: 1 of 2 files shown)


app/controllers/internal/events_controller.rb
```diff
@@ -26,8 +27,7 @@ def show
     def new
       if params[:duplicate]
         @event = get_event_by_id(params[:duplicate])
-        @event.id = nil
-        @event.readable_id = nil
+        @event.assign_attributes(NILIFY_ON_DUPLICATE.to_h { |attribute| [attribute, nil] })
```
jkempster34, line 30: Potentially doing something a bit weird here.

Tried all of these methods https://makandracards.com/makandra/42641-different-ways-to-set-attributes-in-activerecord

and the only one that worked was `assign_attributes`

ethax-ross (approved):

app/controllers/internal/events_controller.rb
peteryates: Just wondering, could this be moved to a model method?

Perhaps something along the lines of:

```
def self.copy(existing_event)
  new(existing_event).tap do |e|
    e.starts_at = nil
    e.ends_at = nil
    ...
  end
end
```

That'd make duplicating easy to test on its own and keep everything in one place?

jkempster34: Looking at some more duplicate stuff now, so I'll bear this in mind

PR #1351, opened 2021-06-02
Fix bug where mailing list bar behind hero title

peteryates:
This happens because the hero title has a z-index of 20 and the green bar has none, so it appears between the hero photo and title. We want it to appear above them both, so now it has a z-index of 40.


MylesJarvis (approved):
LGTM - works on mobile when trying to replicate previous issue

PR #1353, opened 2021-06-03
Adding Events back in

MylesJarvis:
### Context
We want to add Events back into the Nav given the recent issues we've seen with sign up volumes and traffic for events.

Adding 'Teaching events' into the nav

### Guidance to review
How can we make the nav less janky given we've now got an extra category in?


peteryates (approved):
🚢

PR #1355, opened 2021-06-03
Migrate the content assets into webpacker

leoapost:
### Trello card
https://trello.com/c/v3cfxpjd

### Context
The GiT app currently fingerprints assets in two ways:

- Webpacker fingerprints Rails assets in `/app/webpacker/images`
- We manually fingerprint content assets in `/public/assets` as part of the Docker build process

This creates an issue when we want to use a single ViewComponent to reference either kind of asset (webpacker assets on a dynamic page and content assets on a markdown page).

This PR moves the assets inside the webpacker folder and configures webpacker to compile them with the rest assets. Also uses the standard [webpacker helper methods](https://www.rubydoc.info/gems/webpacker/Webpacker/Helper) to reference to content assets, which respect file fingerprints.

Replaces all the document link elements in the content pages with a new reusable cta ViewComponent called Attachment. It requires the link text and the file path and it accepts meta info like the file size, file type, published date and updated date.

Finally, it removes the custom Fingerprinter.

### Guidance to review
Nothing should look different.


peteryates:
I think the failing specs are caused by helpers being called too early in the lifecycle. Moving them back to [before_render](https://viewcomponent.org/guide/lifecycle.html) instead of initialize will probably do the trick 🤞🏽

leoapost:
> I think the failing specs are caused by helpers being called too early in the lifecycle. Moving them back to [before_render](https://viewcomponent.org/guide/lifecycle.html) instead of initialize will probably do the trick 🤞🏽

@peteryates mutating the value of the image attribute from a file path to an html element in the initializer felt wrong, so I refactored it a bit to use the helper method in the view template instead.

peteryates (dismissed):
I've had a good look around and can't see anything that looks broken or missing. This is a huge improvement, glad everything'll be in one place 🙌🏽

app/views/content/research.md
```diff
@@ -4,6 +4,87 @@ description: |-
...
 date: "2021-05-17"
 image: false
+calls_to_action:
+  my-favourite-subject:
+    name: attachment
+    arguments:
+      text: Read the summary of findings
+      file_path: media/documents/my-favourite-subject.pdf
+      file_type: PDF
+      file_size: 266KB
```
MylesJarvis, line 14: @leoapost I think we've missed the publication date for this one

MylesJarvis (approved):
LGTM - nothing appears to have regressed visually

PR #1372, opened 2021-06-09
Improve the search logic to prioritise more relevant content

ethax-ross:
### Trello card

[Trello-1321](https://trello.com/c/lTKFqFhK/1321-dev-advanced-search)

### Context

We want to improve the search results so that users get more relevant content higher-up in their search results.

- Add module containing english stopwords

We will be using these to exclude stopwords from the search terms.

- Update search with ranking/levenshtein comparisons

In order to improve the search we want to rank the results (more keyword matches resulting in a higher-ranked page), make use the Levenshtein comparison to reduce spelling errors impacting the search results and remove stopwords to reduce noise.

The search algorithm now consists of:

```
- Tokenize the search term, keywords and page title (removing stopwords, downcase)
- For each page
  - Count keyword matches and multiply by 2 to add weight
  - Count page title word matches
  - Sum to give page rank
- Remove unranked pages (no matches)
- Sort pages by rank, descending
```

A 'match' is defined as:

```
- The word starts with one of the search tokens
- The word is >= 5 characters and Levenshtein distance is <= 1 to a search token
```

The original search tests have been left as unchanged as possible to show that the new search works with all the original test cases. Tests for the new search logic will be in a follow up commit.

- Add test cases for new search logic

Additional test cases to cover:

```
- Ignoring stopwords
- Words containing apostrophe/hypen
- Mispelled words > 5 characters long
- Weighting keyword results by a factor of 2
```

### Guidance to review

I kept the original search tests unchanged to ensure there are no regressions; tests for the new logic are in a separate commit.

We may want to remove some stopwords from the list; I removed 'way', for example, so that 'ways to train' matches better.

The library I've used for the Levenshtein comparison can also calculate word stems; it may be worth trying a combination of both; for example "teaching" and "teacher" would have "teach" as the stem word; we could even use this in a combination with the Levenshtein comparison so that "teech" matched "teacher", "teaching", "teach", etc.

We could use caching in places to speed this up, but testing it locally it doesn't appear to need it so I've opted not to for now to keep it simpler.


MylesJarvis:
@ethax-ross Looking good so far. Only instances that I'd like to improve are around words relating to 'return' and 'international'. I've added some screen grabs which show them ranking lower than other items that I wouldn't expect them to:
**Return**
Screenshot 2021-06-09 at 16 14 28

**International (one set of results, 2 photos)** 
Screenshot 2021-06-09 at 16 14 54
Screenshot 2021-06-09 at 16 14 28

HettieS:
@ethax-ross found a fun one where typing any word, with a space in 

Uploading Screen Recording 2021-06-09 at 16.45.40.mov…

front of it, yields the same list of results:

MylesJarvis:
@ethax-ross other items that might be worth looking at include more conversational queries - we saw users do this, with and without question marks when we could see query strings. An example might be - 'how do I get student finance?'. Student finance is a keyword in the front matter, so is there a way to get this to show up in search results?

ethax-ross:
@HettieS @MylesJarvis that should be all points addressed:

- The international/returner pages its just a case of ensuring they have enough of the correct keywords to rank. They had returners/returning but not returner in some cases and I added a few international-related keywords to make the other page feature higher; the more keywords that match the higher it ranks, so if a page is lower down just add some additional keywords that match, for example 'international', 'international funding', 'international students'. It may be worth exploring word stemming here, which would mean "returnering" would match "returner", "returns", "returners" etc; it would only take an hour or two to explore this.

- The space bug was due to an empty character being treated as a search token and therefore matching everything; I've ensured they get stripped out now.

- There was another small tweak to downcase the tokens before matching stopwords which is now fixed and searching for "how do I get student finance?" now seems to returner more appropriate results.

MylesJarvis (approved):
LGTM

PR #1374, opened 2021-06-09
Add the urgency notice in the funding widget

leoapost:
### Trello card
https://trello.com/c/zkp39vx7

### Context
As part of our recent push on the end of cycle and trying to increase applications before the new cycle in Sept/Oct 2021, we want to add some new copy to the funding widget.

This copy will reflect other initiatives we have in place, namely about how there's 'still time to apply before September'.

Add a section at the bottom of the widget and display it in all subjects.

### Guidance to review
| before | after |
|-|-|
| before | after |

### DIFF
--- a/app/components/funding_widget_component.html.erb
+++ b/app/components/funding_widget_component.html.erb
@@ -62,8 +62,7 @@
         <div class="call-to-action__text">
           <h3>Start your teacher training this September</h3>
           <div>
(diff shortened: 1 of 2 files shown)


peteryates (approved):

MylesJarvis (approved):
LGTM

PR #1378, opened 2021-06-10
Rename accept cookies button id to bypass ad-blockers

ethax-ross:
### Trello card

[Trello-1623](https://trello.com/c/6yvBReQt/1623-investigate-the-impacts-of-privacy-software-on-the-usability-of-the-site)

### Context

It appears that AdBlockers in general allow our cookie modal to appear, however they hide the "Accept all cookies" button (seemingly targetting the `id` because it includes `cookie`).

A workaround is to rename the `id` to something else - I've left a note in the code to prevent this from being changed back in the future.

- Rename accept cookies button id to bypass ad-blockers

### Guidance to review

The only way I was able to replicate this bug was to install uBlock and enable all of the 'annoyances' under filters:

Screenshot 2021-06-10 at 10 52 20

### DIFF
--- a/app/views/sections/_cookie-acceptance.html.erb
+++ b/app/views/sections/_cookie-acceptance.html.erb
@@ -18,7 +18,11 @@
 
 
             <div class="dialog__buttons">
-              <a href="#" class="button call-to-action-button" data-action="click->cookie-acceptance#accept" data-cookie-acceptance-target="agree" id="cookies-agree">
+              <!--
+                The id is `biscuits-agrees` instead of `cookies-agree`
+                to prevent ad blockers from removing the button.
+              -->
+              <a href="#" class="button call-to-action-button" data-action="click->cookie-acceptance#accept" data-cookie-acceptance-target="agree" id="biscuits-agree">
                   <span>Accept all cookies</span>
               </a>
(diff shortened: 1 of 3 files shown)


app/views/sections/_cookie-acceptance.html.erb
```diff
@@ -18,7 +18,11 @@
 
 
             <div class="dialog__buttons">
-              <a href="#" class="button call-to-action-button" data-action="click->cookie-acceptance#accept" data-cookie-acceptance-target="agree" id="cookies-agree">
+              <!--
+                The id is `biscuits-agrees` instead of `biscuits-agree`
+                to prevent ad blockers from removing the button.
```
jkempster34, line 23: Do you mean ` The id is biscuits-agree instead of cookies-agree`?

ethax-ross: Whoops, yep I fixed that but didn't push it 🤦

jkempster34 (approved):
Neat 📸, didn't know about this 👍 

There seems to be a few lists floating about with classes to avoid using https://gist.github.com/spyesx/42fe84c0ef757d1c38a4

PR #1383, opened 2021-06-10
TTA Campaign Page - Draft

MylesJarvis:


MylesJarvis:
I've taken this as far as I think I can get it for now. @peteryates I believe we've got an updated quote module with image (https://github.com/DFE-Digital/get-into-teaching-app/pull/1245), can you help me implement in this page?

PR #1389, opened 2021-06-14
Lazy-load YouTube iFrame

ethax-ross:
The YouTube iFrame tanks our Google Page Speed Score; instead, we can lazy-load it using the same library we use for lazy loading images after the main content paint.

I'm not sure how much of an impact this will actually have; the Lighthouse audit in Chrome doesn't seem to report the YouTube iFrame page speed hit in the same way as their website/API.


MylesJarvis (approved):
LGTM

PR #1390, opened 2021-06-14
Auth basic

jkempster34:
### Trello card
https://trello.com/c/t73uzqkE

### Context
We want the Basic auth of this app to match the Basic auth of the TTA app. Pasting over the TTA code and updating it to include roles.

In the internal path (`/internal/events`), we want to give users roles of `publisher` or `author` to control what permissions they have. Update the TTA implementation to have a `User` class with the required roles.

The site-wide authentication uses the same credentials as it does currently, so there should be no change to the review apps.

- Paste over Basic auth implementation from TTA app
- Add optional user roles to Basic auth
- Update README as it is now outdated. Include new credentials format.
- Update credentials for:
  - `development`: to have a `publisher_user` and `author_user`
  - `rolling` and `preprod`: to have a `publisher_user` and `author_user` plus the usual credentials for site-wide auth.
  - `production`: users for `Naeem` (`publisher`) and `AbramisUser` (`author`), to test whether the update works properly before adding a full list of users.
- Use Rails credentials for local dev (`development`) instead of environment variables to better reflect the deployed applications. 
- Add some feedback about who is logged in

image


app/controllers/healthchecks_controller.rb
```diff
@@ -1,5 +1,5 @@
 class HealthchecksController < ApplicationController
-  skip_before_action :site_wide_authentication
+  skip_before_action :http_basic_authenticate
```
ethax-ross, line 2: Instead of skipping the action we should maybe return `false` from the `authenticate?` method here?

app/controllers/internal_controller.rb
```diff
@@ -1,47 +1,26 @@
 class InternalController < ApplicationController
-  before_action :authenticate
-  skip_before_action :site_wide_authentication
-  helper_method :publisher?
+  helper_method :publisher?, :username
+  before_action :authorize_user
```
ethax-ross, line 3: What's this doing? I would have expected to see something similar to the TTA service, along the lines of:

```
before_action :restrict_access, if: :authenticate?<br class="Apple-interchange-newline">
```

lib/basic_auth.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    def env_requires_auth?
+      # Site-wide authentication present in all production-like
+      # environments, but not in production itself.
+      !Rails.env.production? && !Rails.env.test? && !Rails.env.development?
+    end
+  end
+end
+
+class User
```
ethax-ross, line 31: Can we pull this out into its own file? I would be inclined to move some of the `author?` and `publisher?` methods in your controller into the `User` model as well and expose a `@user` for the view instead of having `helper_method`s in your controller

ethax-ross (commented):
Looks good, just a bit I think needs reworked around the user and some associated methods

lib/basic_auth.rb
jkempster34: Cool, should this live in `app` rather than `lib`?

lib/user.rb
```diff
@@ -0,0 +1,18 @@
+class User
+  attr_reader :username, :role
+
+  ROLES = { publisher: :publisher, author: :author }.freeze
```
ethax-ross, line 4: Is this read anywhere? Should it be:

```
ROILES = { publisher: "publisher", author: "author" }.freeze
```

Then you could do:

```
def publisher?
  @role == ROLES[:publisher]
end
```

spec/requests/internal_controller_spec.rb
```diff
@@ -21,14 +32,34 @@
...
 
+    expect(session[:user].username).to eq(author_username)
+    expect(session[:user].role).to eq("author")
     assert_response :success
-    expect(session[:role]).to be(:author)
+  end
+
+private
+
+  def generate_auth_headers(login_type)
```
ethax-ross, line 50: Split this out into a support module so it can be reused?

ethax-ross (approved):
LGTM

PR #1393, opened 2021-06-15
Navigation data structure

peteryates:
### Trello card

### Context

The work on the navigation has been split up into smaller chunks.

This change introduces the `Pages::Navigation` model.  It's now responsible for building up a collection of nodes (`Pages::Navigation::Node`) that each represent a page on the site in the navigation tree, and is already used to build the existing navigation (both desktop, mobile and footer).

Currently we're only interested in root pages, but this structure will make it easier to find child pages for a particular node.

Unlike the previous iteration, we're now leaning towards using the path for hierarchy and determining whether or not a page is root based on it.  This is going to be easier to understand than setting `parent` metadata in the front matter.

### DIFF
--- a/app/components/navigation/navbar_component.html.erb
+++ b/app/components/navigation/navbar_component.html.erb
@@ -5,7 +5,7 @@
             <ul>
                 <%= nav_link("Home", '/') %>
                 <% resources.each do |resource| %>
-                    <%= nav_link(resource[:front_matter][:title], resource[:path]) %>
+                    <%= nav_link(resource.title, resource.path) %>
                 <% end %>
                 <%= nav_link("Teaching events", events_path) %>
                 <%= nav_link("Register your interest", mailing_list_steps_path) %>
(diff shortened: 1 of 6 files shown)


app/components/navigation/navbar_component.html.erb
```diff
@@ -5,7 +5,7 @@
             <ul>
                 <%= nav_link("Home", '/') %>
                 <% resources.each do |resource| %>
-                    <%= nav_link(resource[:front_matter][:title], resource[:path]) %>
+                    <%= nav_link(resource.title, resource.path) %>
                 <% end %>
                 <%= nav_link("Teaching events", events_path) %>
```
MylesJarvis, line 10: Are these two additional `nav_link` references going to be artefacts that need removing?

peteryates: Yeah eventually, this PR shouldn't have any visual impact at all though.

ethax-ross (approved):
Makes sense 👍

PR #1394, opened 2021-06-15
TTA Page - Rebased

MylesJarvis:
This is the TTA page that we want to use to sell the value of the TTA, and push traffic from social posts, ad traffic etc. This is a first go at it and will no doubt be subject to iteration as we move forward.


peteryates (approved):
Looks good and makes sense.

PR #1399, opened 2021-06-16
Create a new header component

peteryates:
### Trello card

### Context

This is the second part of the work towards building the new navigation functionalty.

Currently each of our layouts implemented the 'top of the page' bit independently, there was lots of repetition and inconsistency.

This PR includes the following changes:

* Introduce the `HeaderComponent`
* Create the `Header` namespace and add all of the subcomponents it renders to it, these are:
  * `Header::BreadcrumbComponent`
  * `Header::HeroComponent`
  * `Header::NavigationComponent`

The HeroComponent is a slot that takes frontmatter as an argument, meaning we can include or exclude it
on a layout-by-layout basis.

The one negative introduced here is that the `content_for :hero` approach used in events no longer works. For the time being I've moved the arguments inside the controller (along with the title). Happy to revisit this later.

### Review guidance

Overall this shouldn't lead to any visual changes. Does the reorg make sense?


app/components/header_component.rb
```diff
@@ -0,0 +1,9 @@
+class HeaderComponent < ViewComponent::Base
```
ethax-ross, line 1: Worth adding a spec for this component?

ethax-ross (commented):
One suggestion, overall this is much nicer 💯

app/components/header_component.rb
peteryates: Oops, good call 😅

ethax-ross (approved):

PR #1403, opened 2021-06-17
Introduce delay before executing search

ethax-ross:
### Trello card

[Trello-345](https://trello.com/c/eDPlKMPS/345-development-add-js-superset-to-search)

### Context

Currently the search is executed after every key press, which both results in a lot of search API calls and also means our analytics include search terms for each character.

Instead, we want to introduce a 500ms delay after the user finishes typing before executing the search.

- Introduce delay before executing search

### Guidance to review

It would have been nicer to use Jest's fake timers for the tests, however the `accessible-autocomplete` lib uses timers to poll the input field internally and this makes using the fake timers not possible (it just hangs, even when you ask it to only run on the pending timers).

Now that there's a delay do we need a 'loading...' entry instead of 'no results' while the user types?

### DIFF
--- a/app/webpacker/controllers/searchbox_controller.js
+++ b/app/webpacker/controllers/searchbox_controller.js
@@ -69,6 +69,7 @@ export default class extends Controller {
       minLength: 2,
       source: this.performXhrSearch.bind(this),
       confirmOnBlur: false,
+      tNoResults: () => this.searching ? "Searching..." : "No results found",
       onConfirm: this.onConfirm.bind(this),
       templates: {
         inputValue: this.inputValueTemplate.bind(this),
@@ -82,6 +83,8 @@ export default class extends Controller {
   }
 
   performXhrSearch(query, callback) {
+    this.searching = true
+
     if (this.delaySearchTimeout) {
       clearTimeout(this.delaySearchTimeout);
     }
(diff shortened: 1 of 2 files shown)


MylesJarvis:
@ethax-ross I think we need something instead of the 'no results found' that appears. Can we repurpose that to 'searching' or WTTE. I guess we'd then need to ensure 'no results found' continues to surface once the delay has been processed if there are in fact no results

app/webpacker/controllers/searchbox_controller.js
```diff
@@ -82,20 +82,26 @@ export default class extends Controller {
...
+    if (this.delaySearchTimeout) {
+      clearTimeout(this.delaySearchTimeout);
     }
 
-    request.send()
+    this.delaySearchTimeout = setTimeout(() => {
+      this.searchQuery = query
+      this.scheduleSubmissionToAnalytics()
+
+      let request = new XMLHttpRequest()
```
peteryates, line 93: Been a while since I saw a `XMLHttpRequest` in the wild 😂

peteryates (approved):
Looks good to me.

app/webpacker/controllers/searchbox_controller.js
ethax-ross: hah yeah we should probably swap that out to use `fetch` at some point

MylesJarvis (approved):
Looks good to me 👍

app/webpacker/controllers/searchbox_controller.js
peteryates: Yeah - sending IE11 users to the non-JS one is fair imo!

PR #1404, opened 2021-06-17
Grey out events search button

jkempster34:
### Trello card
https://trello.com/c/q28oZTps

### Context
Some users were confused by the events search - after searching events again, they were unclear if the UI had updated to show their new search results.

After a chat with @gazcarr, we decided see what a reduced opacity button with a `not-allowed` cursor would look like.

Behaviour is now:
- Landing on `/events` = Button normal
- Make a valid search = Button disabled (in appearance)
- Make an invalid search - Button normal
- Make a valid search, then change any of the `select` elements, or type in the `input` = Button normal
- Land on a categories page, such as `/event-categories/online-q-as` = Button disabled (event search has already taken place)

Because the button only appears disabled when the JavaScript is attached, the behaviour for people with JavaScript turned off will just be the current behaviour.

- Connect Stimulus `event-search-button` controller when search performed
  - The `event-search-button` Stimulus controller will be used to update the appearance of the `Update results` button. This should only happen when a search has been performed, so conditionally connect the controller using the existing `search_performed` variable. 
- Reduce opacity of `Update results` button when parameters unchanged
  - When the search parameters have not changed since the last search, then reduce to opacity of the `Update results` button to suggest to users that it won't do anything. When any of the parameters are updated, turn opacity back up to 100%.
- Don't set `performed_search` when postcode is invalid
  - Currently, the events search sets a `performed_search` variable when postcode is invalid. Don't set this when search is invalid so that I can control the behaviour of the `Update results` button.
- Update use of search component. 
  - Pass `search component` to view with `content_for` - We want to avoid setting instance variables in the view and passing them to the layout. When we look at the layout we assume any instance variable is set in the controller, which makes the logic harder to follow if it is set in the view instead.

### DIFF
--- a/app/components/events/search_component.html.erb
+++ b/app/components/events/search_component.html.erb
@@ -8,25 +8,25 @@
             <div class="search__controls" data-controller="conditional-field" data-conditional-field-mode="hide" data-conditional-field-expected="">
               <% if include_type %>
                 <%= tag.div(class: input_field_classes(:type)) do %>
-                    <%= f.label :type, "Event type" %>
-                    <%= f.collection_select :type, search.class.available_event_types, :id, :value, { include_blank: "All events" } %>
+                  <%= f.label :type, "Event type" %>
+                  <%= f.collection_select :type, search.class.available_event_types, :id, :value, { include_blank: "All events" }, data: { action: "event-search-button#parameterChanged" } %>
                 <% end %>
               <% end %>
 
               <%= tag.div(class: input_field_classes(:distance)) do %>
-                  <%= f.label :distance, "Location" %>
-                  <%= f.select :distance, search.class.available_distances, {},
-                        data: { action: "conditional-field#toggle", "conditional-field-target": "source" } %>
+                <%= f.label :distance, "Location" %>
+                <%= f.select :distance, search.class.available_distances, {},
+                             data: { action: "conditional-field#toggle event-search-button#parameterChanged", "conditional-field-target": "source" } %>
               <% end %>
 
               <%= tag.div(class: input_field_classes(:postcode), data: { "conditional-field-target": "showhide" }) do %>
                 <%= f.label :postcode, "Postcode" %>
-                <%= f.text_field :postcode %>
+                <%= f.text_field :postcode, data: { action: "event-search-button#parameterChanged" } %>
               <% end %>
 
               <%= tag.div(class: input_field_classes(:month)) do %>
-                  <%= f.label :month, "Month" %>
-                  <%= f.select :month, search.available_months, **month_args %>
+                <%= f.label :month, "Month" %>
+                <%= f.select :month, search.available_months, month_args, data: { action: "event-search-button#parameterChanged" } %>
               <% end %>
             </div>
 
@@ -38,7 +38,8 @@
               </div>
             <% end %>
 
-            <button class="request-button">Update results <%= helpers.fas_icon "sync" %></button>
+            <button class="request-button" id="updateResultsButton" data-event-search-button-target="updateResultsButton">Update
+              results <%= helpers.fas_icon "sync" %></button>
           <% end %>
         </div>
       </div>
--- a/app/components/events/search_component.html.erb
+++ b/app/components/events/search_component.html.erb
@@ -3,7 +3,7 @@
     <article class="markdown overhang fullwidth">
       <h2 id="searchforevents" class="colored-stripe__title purple"><%= heading %></h2>
       <div class="colored-stripe__content">
-        <div class="search">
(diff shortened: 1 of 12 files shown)


app/controllers/events_controller.rb
```diff
@@ -62,7 +62,7 @@ def search_events
     search_results = @event_search.query_events(MAXIMUM_EVENTS_PER_CATEGORY)
 
     @display_empty_types = @event_search.type.nil?
-    @performed_search = true
+    @performed_search = true if @event_search.valid?
```
peteryates, line 65:
```suggestion
    @performed_search = @event_search.valid?
```

peteryates:
Looking good so far, will review i full when the review app's up.

app/controllers/events_controller.rb
jkempster34: 🤦 Thanks

app/views/event_categories/show.html.erb
```diff
@@ -1,6 +1,7 @@
 <% @search_component = Events::SearchComponent.new(
         @event_search,
         @form_action.to_s,
+        @performed_search,
         include_type: false,
         heading: "Search for #{@category_name}",
         allow_blank_month: true,
```
peteryates, line 7: Rather than using an instance variable here I feel like it should be a `content_for :search_component` block, with a corresponding `yield :search_component` in the layout.

peteryates (commented):
This looks good overall and worked for me. I think a `content_for` block could make the logic easier to follow wrt rendering stuff in the view and placing it in the layout.

app/components/events/search_component.html.erb
```diff
@@ -3,30 +3,30 @@
...
                 <%= f.label :postcode, "Postcode" %>
-                <%= f.text_field :postcode %>
+                <%= f.text_field :postcode, data: { action: "event-search-button#parameterChanged" } %>
               <% end %>
 
               <%= tag.div(class: input_field_classes(:month)) do %>
-                  <%= f.label :month, "Month" %>
-                  <%= f.select :month, search.available_months, **month_args %>
+                <%= f.label :month, "Month" %>
+                <%= f.select :month, search.available_months, { **month_args }, data: { action: "event-search-button#parameterChanged" } %>
```
peteryates, line 29:
```suggestion
                <%= f.select :month, search.available_months, month_args, data: { action: "event-search-button#parameterChanged" } %>
```

peteryates (commented):
This looks good to me and seems to work. Left one extra suggestion on there, not sure there's any need to double splat inside a hash. Once that's done this is good to go imo 👍🏽

jkempster34:
> This looks good to me and seems to work. Left one extra suggestion on there, not sure there's any need to double splat inside a hash. Once that's done this is good to go imo 👍🏽

Added to an older commit. Learned `interactive rebase` too which seems useful

peteryates:
>  Learned interactive rebase too which seems useful

**Achievement unlocked 🏆**

peteryates (approved):
Looks good to me.

MylesJarvis:
image

Is it worth having a greyed out box when typing in an error generating postcode too @jkempster34 ?

HettieS (changes requested):
Looks good, but is there any way we can prevent the button from 'flashing yellow' when it is clicked, but greyed out. We don't want to disable the button but the yellow state when it's clicked whilst greyed-out is potentially confusing to the user - it makes it look like it's broken

Repo:
- perform a valid event search
- don't change parameters and click the search button again
- notice button is greyed out but still flashes yellow as if executing something


https://user-images.githubusercontent.com/43522239/122964788-87590e80-d37f-11eb-9f09-7a54546e278e.mov

HettieS:
In fact @MylesJarvis @jkempster34 - I personally wouldn't grey out the search button when landing on a specific event type page e.g. Q&A - I think that is potentially confusing for the user if the button is greyed on entry to a page, not sure what your thoughts are

peteryates:
>  is there any way we can prevent the button from 'flashing yellow' when it is clicked

Well-spotted @HettieS - spot the closet QA! 😂 

@jkempster34, to do that you're best adding a `disabled` attr to your button.

jkempster34:
@HettieS @MylesJarvis 

Maybe this is the wrong approach? How would you feel about updating the search button text to `Search for events` and `Search again`, or something like that?

As it is, the disabled button represents the state of the search. If disabled, that means that clicking it will return the same results. But, it seems to be even more confusing :anguished:

- When landing on a categories page, the search appears before the results. So if it is enabled, users might click it to search (despite the fact that by going to that page, a search has already taken place)
- Once you have corrected an invalid postcode, the button will return new results because you have changed the search parameters.

jkempster34:
Closing this PR due to accessibility issues.

PR #1408, opened 2021-06-17
Split the inline search out to its own partial

peteryates:
This is the second stab at fixing search. It could do with some proper tests and they will follow, but this just gets both JS and non-JS searches working.

The problem arose because the format for search results is different (and has been for some time) between the JS and non-JS searches, yet they share the same partial. This wasn't caught by our tests because they mock the results rather than actually running searches against the site.

The previous attempt #1406 fixed the non-JS search and broke the JS one.

## Reviewing suggestions

Ensure both search types work in the review app.

### DIFF
--- a/app/controllers/searches_controller.rb
+++ b/app/controllers/searches_controller.rb
@@ -33,10 +33,10 @@ private
     end
   end
 
-  def render_result(*result)
+  def render_result(*inline_result)
     render_to_string \
-      partial: "result",
-      object: result,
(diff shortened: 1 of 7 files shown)


ethax-ross:
@peteryates I'm getting a 500 on the JS disabled search this time, only seems to happen when there are results - works fine for no results

ethax-ross (approved):
Looks good and works well 👍

PR #1409, opened 2021-06-17
Event upsert rate limiting

jkempster34:
### Trello card
https://trello.com/c/t73uzqkE

### Context
The CRM team requested that rate limiting is applied to the upsert endpoint for additional security. We have rate limiting of 100 upserts per day in the API, also add rate limiting of 5 upserts per IP per minute here.

- Apply rate limiting to event upsert endpoints (`POST /internal/events` and `PUT /internal/approve`)

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -55,6 +55,16 @@ module Rack
 
         req.ip if (req.patch? || req.put?) && path_performs_event_sign_up
       end
+
+      # Throttle event upsert by IP (5rpm)
+      throttle("event upsert", limit: 5, period: 1.minute) do |req|
+        event_upsert_paths = [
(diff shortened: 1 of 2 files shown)


config/initializers/rack_attack.rb
```diff
@@ -55,6 +55,15 @@ class Attack
 
         req.ip if (req.patch? || req.put?) && path_performs_event_sign_up
       end
+
+      throttle("event upsert", limit: 100, period: 1.day) do |req|
```
ethax-ross, line 59: As the API will already be rate limiting to 100req/day this isn't going to add any value; instead you would be better factoring in the client IP and dropping the rate limit to something like 5req/min/ip

ethax-ross (approved):

PR #1410, opened 2021-06-18
New navigation 

peteryates:
### Trello card

### Context and changes

This is the ~~latest~~ (final 🤞) attempt at re-writing the nav. This time we're building on top of #1399 and  #1393, which do away with much of the tech debt that got in the way of previos attempts.

* [x] modularise and reorganise the CSS
* [x] use separate components for all the header parts
* [x] replace the current navigation bar with a new one
* [x] make the new navigation bar support a secondary level
* [x] enhance the `Pages::Navigation` model so it looks for subpages by path
* [x] add a `menu` flag to front matter for nav entries that are lists of other pages
* [x] general reworking of the search styles and other top-of-page furniture


peteryates:
WIP preview.

https://user-images.githubusercontent.com/128088/123107649-a3b58380-d431-11eb-9a26-e28f927cc838.mp4

peteryates:
WIP mobile nav.

Screenshot from 2021-06-24 10-46-50

gazcarr:
@peteryates @MylesJarvis 

**Top bar**

- Icons needed, search icon can go to the right of the search input (can it go inside the search input re. accessibility)
- Placeholder text to be the same size as 'register your interest', could we shorten this to 'Site search' or 'Search teaching'
- The grey bar needs a little more padding top and bottom now we have the input there

**Primary Nav**

- Simple text-underline on hover- Do we need home? Can it just be 'Events'?
- When screen size is reduced and the text wraps onto 2 likes, can we vertically align the text to the top? ... Currently, the home is the only one that doesn't wrap but it sits at the bottom and alignment looks a bit odd.

**Mobile layout Nav**

- Should have a full grey background
- White border at the bottom of each link- More padding top and bottom
- See the image mockup below

mobilenav

MylesJarvis (approved):
LGTM - few optimisations still to do in a new PR, but let's see how this does in Prod

spec/javascript/controllers/navigation_controller_spec.js
```diff
@@ -2,89 +2,59 @@ import { Application } from 'stimulus';
...
 
 describe('NavigationController', () => {
-  const navigationTemplate = `<div class="navbar__mobile" data-controller="navigation">
-        <div class="navbar__mobile__buttons">
-            <a data-action="click->navigation#navToggle" href="javascript:void(0);" class="icon">
-                <div data-navigation-target="hamburger" id='hamburger' class="icon-close"></div>
-                <div data-navigation-target="label" id="navbar-label" class="icon-hamburger-label">Close</div>
+  describe('opening and closing the nav menu', () => {
+    document.body.innerHTML = `
+    <header data-controller="navigation">
```
MylesJarvis, line 7: This used to be a div class, would that have any impact?

MylesJarvis: @peteryates

app/components/header/hero_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= tag.header(class: classes) do %>
+<%= tag.div(class: classes) do %>
```
MylesJarvis, line 1: why has this changed from a header to a div?

PR #1411, opened 2021-06-21
Bake in seo terms to Salaries page

MylesJarvis:
### Trello card
https://trello.com/c/45hUahfm/1637-content-rework-existing-pages-to-include-seo-organic-search-friendly-terms
### Context
This is a starter for 10 on baking in SEO terms. Making use of @Tomcowan1's initial go at salaries and benefits.


peteryates (approved):
_You won't believe these ten secret facts about teaching salaries!_

PR #1415, opened 2021-06-22
Add check for static asset availability

ethax-ross:
### Trello card

[Trello-1622](https://trello.com/c/leu6mmtZ/1622-investigate-packcs-js-applicationjs-404ing)

### Context

We are seeing sporadic 404s being returned for assets that contain a digest from an old deployment. We suspect CloudFront is not caching these assets according to the 1 year cache-control TTL we have set, but we want to determine if users are running into this error during deployments when the asset digests are changing, or if its just bots poking about.

This introduces an `AssetChecker` that loads the home page, pulls out the CSS/JS pack URLS and then attempts to download them; if the requests fail it will be returned as a 404 to the client with a list of the failed URLs, if it passes then we return an empty 204.

We are using the page speed environment for ease; this doesn't really warrant its own environment and it can be torn out once we get to the root cause of the 404ing.

- Add check for static asset availability

### DIFF
--- a/.github/workflows/static_asset_checker.yml
+++ b/.github/workflows/static_asset_checker.yml
@@ -1,0 +1,25 @@
+name: Run Static Asset Checker
+on:
+  schedule:
+    - cron:  '* * * * *' # every minute
+  workflow_dispatch:
+
+jobs:
+  pagespeed:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: Azure/login@v1
+      with:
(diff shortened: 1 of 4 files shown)


peteryates (approved):
Looks good to me. If it's running minutely are we likely to hit any quotas or anything?

ethax-ross:
It shouldn't be an issue I don't think we have any quotas/limits on our GitHub action usage 👍

PR #1423, opened 2021-06-23
Prevent double clicking form buttons

ethax-ross:
### Trello

[Trello-1659](https://trello.com/c/Cl0hYG6a/1659-duplicate-crm-record)

### Context

As we are not using the `govuk_submit` helper (due to our own submit button styling) the default prevent double click behaviour was not being applied.

Update to include the `data-prevent-double-click` and `data-disable-with` for correct double click prevention behaviour.

Update disabled button styling to fade it out slightly so it looks disabled.

- Prevent double clicking form buttons

### Guidance to review

I could have swapped it out to use the `govuk_submit` and re-style it but I think that may have ended up messier than applying the double click attributes manually.

It would be worth DRYing up the `_form` partials but I'd prefer to do that in a separate PR.

### DIFF
--- a/app/views/mailing_list/steps/_form.html.erb
+++ b/app/views/mailing_list/steps/_form.html.erb
@@ -7,10 +7,8 @@
     <%= render current_step.key, current_step: current_step, f: f %>
 
     <% if wizard.can_proceed? %>
-      <%= f.button class: "call-to-action-button button" do %>
-        <%= wizard.last_step? ? "Complete sign up" : "Next Step" %>
-        <span></span>
-      <% end %>
+      <% button_text = safe_html_format(wizard.last_step? ? "Complete sign up <span></span>" : "Next Step <span></span>") %>
+      <%= f.button button_text, class: "call-to-action-button button", data: { "prevent-double-click": true, "disable-with": button_text } %>
     <% end %>
   <% end %>
 </section>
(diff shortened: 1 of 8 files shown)


app/views/mailing_list/steps/_form.html.erb
```diff
@@ -7,9 +7,9 @@
     <%= render current_step.key, current_step: current_step, f: f %>
 
     <% if wizard.can_proceed? %>
-      <%= f.button class: "call-to-action-button button" do %>
-        <%= wizard.last_step? ? "Complete sign up" : "Next Step" %>
-        <span></span>
+      <% button_text = safe_html_format(wizard.last_step? ? "Complete sign up <span></span>" : "Next Step <span></span>") %>
+      <%= f.button class: "call-to-action-button button", data: { "prevent-double-click": true, "disable-with": button_text } do %>
+        <%= button_text %>
       <% end %>
```
peteryates, line 13: This could be shortened a bit

```suggestion
      <%= f.button button_text, class: "call-to-action-button button", data: { "prevent-double-click": true, "disable-with": button_text } %>
```

peteryates (dismissed):

app/views/mailing_list/steps/_form.html.erb
ethax-ross: 👍 updated

ethax-ross:
@jkempster34 good spot this is because we toggle that button dynamically from JS as well 😩 I'll put in a fix

jkempster34:
> @jkempster34 good spot this is because we toggle that button dynamically from JS as well 😩 I'll put in a fix

Latest push seems to have fixed this 👍

peteryates (approved):

PR #1430, opened 2021-06-25
Set `is_online` to true for `online` type events on creation

jkempster34:
### Context
Some bad events were created via the CRM recently which alerted me to the fact that all `online` type events must have the `is_online` property set to true. Currently, there is a potential for `online` type events to be posted to the API with `is_online` set to `nil`. An event object is always initialised with a `type_id`, so set `is_online` there.

The behaviour for Provider events will remain the same, where the user chooses via the form.

Set `is_online` to true for `online` type events on creation

### DIFF
--- a/app/models/internal/event.rb
+++ b/app/models/internal/event.rb
@@ -106,6 +106,12 @@ module Internal
                        .new.get_teaching_event_buildings
     end
 
+    def type_id=(value)
+      super(value)
+
+      self.is_online = true if online_event?
+    end
+
     def invalid?
       invalid_building = building.present? && building.invalid?
       super || invalid_building
--- a/spec/models/internal/event_spec.rb
+++ b/spec/models/internal/event_spec.rb
@@ -234,6 +234,15 @@ describe Internal::Event do
     end
   end
 
+  describe "#type_id=" do
+    context "when event is initialised with 'online' event_type" do
+      subject { described_class.new({ type_id: GetIntoTeachingApiClient::Constants::EVENT_TYPES["Online event"] }) }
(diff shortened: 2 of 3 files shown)


app/models/internal/event.rb
```diff
@@ -106,6 +106,12 @@ def buildings
                        .new.get_teaching_event_buildings
     end
 
+    def type_id=(value)
+      super(value)
+
+      assign_attributes({ is_online: true }) if online_event? && is_online.nil?
```
ethax-ross, line 112: Does this need to use `assign_attributes`? will it not work with the accessor (the `is_online.nil?` is redundant as well IMO - if someone sets an `is_online` to `false` and then changes the type to online event we should just toggle it to `true`:

```
self.is_online = true if online_event?
```

spec/models/internal/event_spec.rb
```diff
@@ -234,6 +234,15 @@
     end
   end
 
+  describe "#type_id=" do
+    context "when event is initialised with 'online' event_type" do
+      subject { described_class.new({ type_id: GetIntoTeachingApiClient::Constants::EVENT_TYPES["Online event"] }) }
+      it "sets 'is_online' to true" do
+        expect(subject.is_online).to be true
```
ethax-ross, line 241: Could be simplified to:

```
it { is_expected.to be_is_online }
```

app/models/internal/event.rb
jkempster34: Ah, I was missing the `self`

ethax-ross (approved):

PR #1435, opened 2021-06-25
Revamp the events page

leoapost:
### Trello card
https://trello.com/c/n80P96ba

### Context
Users need to scroll a lot to see the events list in the current Events page, especially when on mobile devices. We want to remove the "Types of events" and the "Some events have moved online" blocks to declutter the page and make it easier for the users to see the available events.

Also, the TTT (Train to Teach) events are ending today and we want to display a message when users are searching for events, that more TTT events will be added in September and remind them that they can sign up for events notifications. It's a temp message we want to display until September and then it will be removed.

Remove the 
- "Types of events" block
- "Some events have moved online" block
- "Discover event" title

Add a "no Trains to Teach events" message with a call-to-action for the users to sign up for notifications. The message is displayed only when there are no TTT events available and the users are searching either for `all events` or `TTT events` .

Add subheadings in event category displayed in the grey blocks. This will offers some quick description to the users, since we removed the "Types of Events" block.

Resize the NoResultsComponent from 70% to full-size in order to make it thinner.

### DIFF
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -116,7 +116,7 @@ module EventsHelper
   # Currently the button needs to be hidden only in
   # the TTT block and when the are no TTT events.
   def show_see_all_events_button?(type_id, events)
-    events.blank? && type_id == ttt_event_type_id ? false : true
+    events.present? || type_id != ttt_event_type_id
   end
 
   def ttt_event_type_id
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -102,10 +102,11 @@ module EventsHelper
 
   def display_no_ttt_events_message?(performed_search, events, event_search_type)
     train_to_teach_id = GetIntoTeachingApiClient::Constants::EVENT_TYPES["Train to Teach event"]
+    searching_for_ttt = train_to_teach_id.to_s == event_search_type
+    searching_for_all = event_search_type.blank?
 
-    searching_for_ttt_or_all_events = [train_to_teach_id.to_s, ""].include? event_search_type
     no_ttt_events = events.map(&:first).exclude?(train_to_teach_id)
 
-    performed_search && searching_for_ttt_or_all_events && no_ttt_events
+    performed_search && (searching_for_ttt || searching_for_all) && no_ttt_events
   end
 end
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -84,7 +84,7 @@ module EventsHelper
 
   def event_type_color(type_id)
     case type_id
-    when GetIntoTeachingApiClient::Constants::EVENT_TYPES["Train to Teach event"]
+    when ttt_event_type_id
       "purple"
     else
       "blue"
@@ -101,12 +101,25 @@ module EventsHelper
   end
 
   def display_no_ttt_events_message?(performed_search, events, event_search_type)
-    train_to_teach_id = GetIntoTeachingApiClient::Constants::EVENT_TYPES["Train to Teach event"]
+    train_to_teach_id = ttt_event_type_id
     searching_for_ttt = train_to_teach_id.to_s == event_search_type
     searching_for_all = event_search_type.blank?
 
     no_ttt_events = events.map(&:first).exclude?(train_to_teach_id)
 
     performed_search && (searching_for_ttt || searching_for_all) && no_ttt_events
+  end
+
+  # Determines if the "see all events" button should
+  # be shown in the events blocks or not.
+  #
+  # Currently the button needs to be hidden only in
+  # the TTT block and when the are no TTT events.
+  def show_see_all_events_button?(type_id, events)
+    events.blank? && type_id == ttt_event_type_id ? false : true
+  end
+
+  def ttt_event_type_id
+    GetIntoTeachingApiClient::Constants::EVENT_TYPES["Train to Teach event"]
   end
 end
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -99,4 +99,13 @@ module EventsHelper
     t "event_types.#{type_id}.name.past",
       default: "Past " + pluralised_category_name(type_id)
   end
+
+  def display_no_ttt_events_message?(performed_search, events, event_search_type)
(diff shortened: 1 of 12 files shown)


app/helpers/events_helper.rb
```diff
@@ -99,4 +99,13 @@ def past_category_name(type_id)
     t "event_types.#{type_id}.name.past",
       default: "Past " + pluralised_category_name(type_id)
   end
+
+  def display_no_ttt_events_message?(performed_search, events, event_search_type)
+    train_to_teach_id = GetIntoTeachingApiClient::Constants::EVENT_TYPES["Train to Teach event"]
+
+    searching_for_ttt_or_all_events = [train_to_teach_id.to_s, ""].include? event_search_type
```
ethax-ross, line 106: I don't feel that strongly about it, but I had to do a double take to figure out why you're checking for an empty string here (its clearer when you know that `event_search_type` comes from a query string param, or lack thereof). IMO it may be easier to read if you split them:

```
searching_for_ttt = train_to_teach_id == event_search_type
searching_for_all = event_search_type.blank?
```

ethax-ross (approved):
One take it or leave it comment but looks good overall - like going back to the original event results design 🙈

MylesJarvis:
> One take it or leave it comment but looks good overall - like going back to the original event results design 🙈

Indeed - turns out that we had

app/views/events/index.html.erb
```diff
@@ -41,8 +22,10 @@
     %>
   <% end %>
 <% else %>
-  <%= render(Events::NoResultsComponent.new) do %>
-    Sorry your search has not found any events, try a different type, location or month.
+  <% if @performed_search && @events_search_type != "222750001" %>
+    <%= render(Events::NoResultsComponent.new) do %>
+      Sorry your search has not found any events, try a different type, location or month.
```
MylesJarvis, line 27: @leoapost is it a matter of just updating this to include the sign up link?

app/helpers/events_helper.rb
leoapost: I've changed it as suggested :)

app/views/events/index.html.erb
leoapost: @MylesJarvis we need that as is for the other event types, but you were not that far off. We just need to check if the event type is a TTT one and render the NoResultsComponent with the proper content.

ethax-ross (dismissed):

peteryates (dismissed):
Looks good overall - definite improvement.

I wonder if it'd make sense (in a later PR) to create a presenter and offload some of the heavier helpers to it?

leoapost:
> Looks good overall - definite improvement.
> 
> I wonder if it'd make sense (in a later PR) to create a presenter and offload some of the heavier helpers to it?

Yeah, I agree. 

We are going to revisit the page just before September when more TTT events will be added. Half of the code introduced in this PR will be removed and it will probably be a good opportunity to see what other improvements we can do.

MylesJarvis:
@leoapost just waiting for some initial views from Events and then we can ship it - I'll do it once it's ready! Thanks for all your work on this.

app/views/events/index.html.erb
```diff
@@ -1,48 +1,33 @@
...
-  <% unless @performed_search %>
-  <h2>Discover events</h2>
-  <% end %>
   <% @events_by_type.each do |type_id, events| %>
     <%= render "event_group",
       type_id: type_id,
       events: events,
       page_param_name: @performed_search ? @group_presenter.page_param_name(type_id) : nil,
-      show_see_all_events: true
+      show_see_all_events: events.blank? && type_id == 222750001 ? false : true
```
ethax-ross, line 23: Determined to get that ternary in aren't you? 😅 

```
show_see_all_events: events.present? && type_id != 222750001
```

Worth having the integer as a constant as its not clear which type it is? same for the conditional below

MylesJarvis (approved):

app/views/events/index.html.erb
leoapost: > Determined to get that ternary in aren't you? 

Hahaha, I hope I'm not wrong this time 
 
> ```
> show_see_all_events: events.present? && type_id != 222750001
> ```
So, the `events.present? && type_id != 222750001` will evaluate to false and hide the "see all events" button when there are events to show in the TTT block, but what we want is to hide the button only when there are no events to show in the TTT block. 

This table may help to illustrate the condition
TTT category | Has no events | Hide the button
-- | -- | --
TRUE | TRUE | FALSE
TRUE | FALSE | TRUE
FALSE | TRUE | FALSE
FALSE | FALSE | FALSE

May worth exploring the idea of hiding the button in any category (not just TTT) that doesn't have events. In another iteration though as that would be out of this PR's scope.

> Worth having the integer as a constant as its not clear which type it is? same for the conditional below

Actually, we do have the `GetIntoTeachingApiClient::Constants::EVENT_TYPES["Train to Teach event"]`. I could use that instead.

peteryates: The table definitely helps. Perhaps it should be in a helper method with an explanation of the logic/reasoning?

peteryates (commented):
Looks good to me overall. I left a comment though, I think a helper method is worth it here.

app/views/events/index.html.erb
ethax-ross: Ah forgot to switch it to an or when I negated it, so I think this would work:

```
show_see_all_events: events.present? || type_id != 222750001
```

But yeah as Pete mentioned a helper is probably the way to go

leoapost: >The table definitely helps. Perhaps it should be in a helper method with an explanation of the logic/reasoning?

Good call. I've moved it in the `show_see_all_events_button?()` helper method and added some comments. I hope it's more clear now.

> > Worth having the integer as a constant as its not clear which type it is? same for the conditional below
>
> Actually, we do have the GetIntoTeachingApiClient::Constants::EVENT_TYPES["Train to Teach event"]. I could use that instead.

I replaced that code with the `ttt_event_type_id()` helper method.

peteryates (dismissed):
Looks good, works well for me too.

ethax-ross (approved):

ethax-ross (approved):

peteryates (approved):

PR #1442, opened 2021-06-30
Adding keyword 'average'

Tomcowan1:
Adding average salary information for SEO purposes.


MylesJarvis (approved):

MylesJarvis:
@Tomcowan1 I've removed the lead practitioner salary statement as there's some difference it seems between leadership scales and leading practitioner scale. I want to look into that more for accuracy before we put something out there.

PR #1455, opened 2021-07-01
Add and configure Skylight profiler

peteryates:
[Skylight](https://www.skylight.io) should provide us with some more in-depth profiling and hopefully give us a hint as to why CPU usage is incrementally rising.


ethax-ross (dismissed):

leoapost (dismissed):
Looks good

PR #1460, opened 2021-07-02
Allow users to paste a date and time into `start at` and `end at` inputs

jkempster34:
### Context
The users of the `internal/events` forms receive events via email from the Advertise a Teaching Event form on the main website. They would like the ability to paste a date and time into the `start at` and `end at` fields in the format of the email (e.g., `01/07/2021 17:00`), as it is tedious to type the full date out or to use the selector. This updates the `internal/events` form to use `flatpickr` for its datetime fields, which lets us paste into the input and specify custom formatting.

The API accepts a full ISO format datetime such as `2021-07-01T09:00:00.000Z`, or a localised format such as `2021-07-02T09:00:00`. However, the CRM seems to store the localised version without any additional time zone information (it also stores a separate time zone code), which in turn will update the API cache after a sync. This means (without changing how the API works) it is easier to work with a localised datetime in the APP. Otherwise, the format of the returned date will unexpectedly 
change when a sync runs, and both formats will need to be handled.

We were previously using `<input type="datetime-local">` which also stores a localised datetime with no timezone information and was working nicely.

- Swap inputs to `flatpickr` inputs
- Move `must be in the future` validation into private method to match the other validation.
- Add a hint above the date fields.
- Small tidy up of `internal.js`

### DIFF
--- a/app/models/internal/event.rb
+++ b/app/models/internal/event.rb
@@ -45,11 +45,7 @@ module Internal
     validates :provider_website_url, presence: true, allow_blank: false, length: { maximum: 300 }, if: -> { provider_event? }
     validates :scribble_id, presence: true, allow_blank: false, length: { maximum: 300 }, if: -> { online_event? }
     validates :venue_type, inclusion: { in: VENUE_TYPES.values }
-    validates_each :start_at, :end_at do |record, attr, value|
-      unless value.nil?
-        record.errors.add attr, "Must be in the future" if value <= Time.zone.now
-      end
-    end
+    validate :dates_in_future
     validate :end_after_start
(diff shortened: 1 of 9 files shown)


app/models/internal/event.rb
```diff
@@ -141,6 +137,17 @@ def end_after_start
       end
     end
 
+    def dates_in_future
```
ethax-ross, line 140: I don't feel strongly about it but using something like the [validates_timeliness](https://github.com/adzap/validates_timeliness) gem would make this cleaner

ethax-ross (approved):

PR #1479, opened 2021-07-07
Switch global font family to HK Grotesk

peteryates:
### Trello card

https://trello.com/c/KbwlgkUP/1691-test-font-options-on-site

### Context

The Get Into Teaching styleguide states that [AkzidenzGrotesk Grotesk](https://www.bertholdtypes.com/font/akzidenz-grotesk/proplus/) is the font of choice for all Get Into Teaching marketing activities. We've been unable to source the official font because of licencing reasons.

- Remove AkzidenzGrotesk fonts
- Added HKGrotesk fonts
- Configure the global font face to be "HK Grotesk"

### Checks

* [ ] Are we within the bounds of the [usage licence](https://hanken.co/pages/web-fonts-eula)? I _believe_ we are, but because our website's code is publically available on GitHub we _might_ fall foul of this clause:
  >  Licensee must provide the necessary protection against unauthorized access to the Web Font(s).

  However, every user who visits the site will be downloading the same `.woff` and `.woff2` files so I'm not sure if it's actually an issue.


peteryates:
Not quite close enough to Akzidenz, the 'g' and 't' differ too much.

PR #1486, opened 2021-07-08
Ensure `events/search` with no params does not raise error

jkempster34:
### Context
Fix - The events search is attempting to access params. When they are not there, we are getting a nil exception. 

- Safely access params with dig.


leoapost (approved):

peteryates (commented):
Nice!

PR #1487, opened 2021-07-08
Standardise the bottom-of-page furniture with a single component

peteryates:
This is a precursor to the upcoming 'Welcome guide' work.

The welcome guide will need a footer that doesn't contain a "Talk to us" section and shouldn't show the mailing list bar. As almost every layout has a list of 4-6 partials that are included, usually the same ones, it makes sense for these to be consolidated into a `FooterComponent`. The initializer takes some arguments that toggles the rendering of the new `MailingListComponent`, `TalkToUsComponent`, etc.

This is a much cleaner approach.


jkempster34 (approved):
Looks really neat 📸

jkempster34 (changes requested):
Wait, no. Unapprove ❌ 

You have left a Hi in here

image

peteryates:
Ooops!

jkempster34 (approved):

PR #1491, opened 2021-07-09
Add alt text to blog images

peteryates:
### Trello card

https://trello.com/c/mSGvxZoZ/1651-design-develop-position-new-blog-for-website

### Context

The images didn't have alt text, now it's enforced.

- Replace the png image with a jpg, much smaller
- Make the ImageComponent require alt text
- Add a helper to ease finding the first blog image
- Update blog posts to include alt text for photos
- Extend the markdown handler to pass on all args


app/helpers/blog_helper.rb
```diff
@@ -2,4 +2,10 @@ module BlogHelper
   def format_blog_date(date_string)
     Date.parse(date_string).to_formatted_s(:govuk_zero_pad)
   end
+
+  def first_image_from_post(images)
```
leoapost, line 6: I was wondering if it would be better to be more explicit as to which is the first image, to avoid issues when reordering the images in the frontmatter?

Could do something like

```Ruby
def header_image(img)
  img_params = img.symbolize_keys

  render Content::ImageComponent.new(**img_params)
end
```
and then in the `_post_preview` partial
```Ruby
header_image(fm[:images][:header_image])
```

Just a though though. Not a biggie.

leoapost (approved):
Looks good to me. Left a comment but nothing big.

app/helpers/blog_helper.rb
peteryates: Good question. I think that would build an expectation that each post would need a `header_image` entry in the front matter and nothing would be displayed without one.

I think having it use the first one seems good, but happy to switch if it becomes problematic.

PR #1492, opened 2021-07-09
Start building/experimenting with welcome guide

peteryates:
### Trello card

https://trello.com/c/wYBPu5Fw/1587-prepare-dev-work-for-welcome-guide-once-approved

### Context

Havas have designed a welcome guide that's intended to be accessible via a link once people have signed up to the mailing list. It will include some personalised content and have a distinct look to the rest of the site.

Screenshot from 2021-07-09 16-44-06


peteryates:
Here's a preview. We don't (yet) really know how we're going to accommodate screen sizes other than those provided in the wireframes.

Screenshot from 2021-07-09 16-27-22

peteryates:
A glimpse of today's progress (sans squiggly arrows):

| Device | Welcome to Teaching | Maths is rewarding |
| --------- | --------------------------- | ---------------------------|
| Desktop | Screenshot from 2021-07-12 16-32-15 |Screenshot from 2021-07-12 16-31-56 |
| Mobile | Screenshot from 2021-07-12 16-35-01 | Screenshot from 2021-07-12 16-31-46 |

## Some notes

* there's an **incredibly high** CSS:content ratio
* it's quite inflexible, content changes are likely to require CSS tweaks and vice versa
* the CSS is complex in parts

peteryates:
Above I mentioned that this is a high effort, complex approach. I'm not sure there's a better way of doing it, but for completeness here's how it's built (so far!)

* there's a new welcome page layout, it's stripped back and only contains the header and footer (sans 'talk to us')
* added a `welcome.md` page that's empty and am using our existing `content` frontmatter partial loaded to pull in each section
* sections are their own separate HTML partials
* sections have a corresponding set of rules in `welcome.scss` that govern their layout, colours, look, feel etc
  For the layout we need to use [grid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout) as it's the only sensible way of making content overlap. Each grid is different because the layouts vary. I plan each one before I start like this:

| One | Two |
| ----- | ----- |
| Screenshot from 2021-07-12 15-18-47 | Screenshot from 2021-07-13 15-21-38 |

ethax-ross (dismissed):
Looks good code-wise; the "did you know" section looks a bit off on Mac OS Chrome and even more off on Safari, there's a lack of left-padding on the question and the answer buttons are oval rather than circular:

| Chrome      | Safari |
| ----------- | ----------- |
| Screenshot 2021-07-21 at 08 22 56    | Screenshot 2021-07-21 at 08 23 34   |

Happy to approve as a first iteration though ✅

peteryates:
Ah thanks @ethax-ross - looks like the focus state is a funny on Chrome too:

Screenshot from 2021-07-21 10-21-55

I'll adjust that and push up a fix.

peteryates:
After the latest push, bigger title, better spacing and rounder radios 🔘 

Screenshot from 2021-07-21 10-36-12

app/views/layouts/welcome.html.erb
```diff
@@ -5,6 +5,9 @@
 
     <%= render HeaderComponent.new %>
 
+    <%# FIXME: remove this to enable the page in prod %>
```
MylesJarvis, line 8: @ethax-ross can you confirm this is all good form a dev perspective to stop it rendering in prod? I'll review the rest of the copy and design in the wider PR.

ethax-ross: I would have stuck in a route to catch it but that'll get the job done! 🙈

ethax-ross (approved):

app/views/layouts/welcome.html.erb
peteryates: Yeah, route would be much cleaner but went for the keeping it all in one place approach 😅

MylesJarvis (approved):

PR #1510, opened 2021-07-16
Hard-code the mailing list URL in the header and footer

peteryates:
### Context and changes

We're currently relying on Rails' routing for generating the mailing list paths in these two places but using the hard-coded path everywhere else in the app.

Due to the way the sign-up wizard is written the individual pages don't have their own paths; instead they rely on a redirect in WizardSteps#index to send users to the right place.

A _correct_ navigation to `/mailinglist/signup` results in a 302 and the user is routed to `/mailinglist/signup/name` (the first step in that particular wizard). We believe this redirection has had an impact on paid search results.

We have two options here:

* Supply the wizard class to the `step_path` method so we can use Rails'
  built-in functionality:

  ```ruby
    redirect_to step_path(wizard_class.first_key, query_params)
  ```

* or, hard-code it

I'm going for the latter here as it's lighter touch, we don't need to introduce any new routing rules or helpers to ease the process and we can test it out. It also makes links to the mailing list easier to find via ag/grep

The disadvantage is that we can't change the routing on a whim and have the new rules reflected throughout automatically.

### DIFF
--- a/app/components/footer_component.html.erb
+++ b/app/components/footer_component.html.erb
@@ -43,7 +43,7 @@
           <a href="<%= resource.path %>"><%= resource.title %></a>
         <% end %>
         <%= link_to "Find an event near you", events_path %>
-        <%= link_to "Sign up for updates", mailing_list_steps_path %>
+        <%= link_to "Sign up for updates", "/mailinglist/signup/name" %>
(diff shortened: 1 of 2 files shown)


MylesJarvis (dismissed):
LGTM

peteryates:
Just improved the commit message a bit @MylesJarvis, mind re-checking?

MylesJarvis (approved):

PR #1512, opened 2021-07-16
Revamp the first page of the signup form

leoapost:
### Trello card
https://trello.com/c/8XeG8nER

### Context
We want to see if we can increase the mailing list signups by making the first page of the signup form more attractive.

Create a new layout to be rendered only in the first page of the signup page.

### Guidance to review
|before|after|
|-|-|
|||
|||


app/views/mailing_list/steps/_name.html.erb
```diff
@@ -13,7 +13,7 @@
 
 <%= f.govuk_error_summary %>
 
-<%= f.govuk_text_field :first_name %>
-<%= f.govuk_text_field :last_name %>
-<%= f.govuk_email_field :email %>
+<%= f.govuk_text_field :first_name, class: 'govuk-!-width-two-thirds' %>
+<%= f.govuk_text_field :last_name, class: 'govuk-!-width-two-thirds' %>
+<%= f.govuk_email_field :email, class: 'govuk-!-width-two-thirds' %>
```
peteryates, line 18: You can add the classes using built-in formbuilder functionality, https://govuk-form-builder.netlify.app/form-elements/text-field/#generating-text-inputs-with-custom-widths

leoapost: That's a nice one! I've updated it to use the `width:`

app/views/layouts/registration_attractive.html.erb
```diff
@@ -0,0 +1,27 @@
...
+
+  <%= render "sections/head" %>
+  <%= analytics_body_tag class: "govuk-template__body govuk-body", data: { controller: "video link scroll", "link-target": "content" }, id: "body" do %>
+
+    <%= render HeaderComponent.new %>
+
+    <main role="main" id="main-content" class="attractive-layout">
+      <section class="signup-container">
+
+        <%= image_pack_tag 'content/hero-images/0025.jpg', class: 'left-img hide-on-mobile hide-on-tablet' %>
```
MylesJarvis, line 15: Can we add alt tags to these?

MylesJarvis: @leoapost

peteryates: I'd create empty alt tags for these ones, `alt: ""`, as they're purely for decoration and we wouldn't want the screen reader to distract the user from the form.

Annoyingly, if there's no alt tag at all the screen reader will just read out the filename.

leoapost: I just pushed a commit with the empty `alt` attributes.

MylesJarvis (approved):
LGTM

PR #1515, opened 2021-07-19
Support Open Graph images

ethax-ross:
### Trello card

[Trello-1700](https://trello.com/c/yDPUgqJo/1700-ensure-that-twitter-posts-have-a-thumbnail-that-is-pulled-through-to-complement-the-accompanying-snippet)

### Context

When a web page is shared on social media the platform will render the Open Graph image contained within the `og:image` meta tag in the post.

Currently we do not specify an `og:image` meta tag and get the default no image/grey box icon for most platforms.

We can also provide `alt` text; as we have opted to use the hero images I've copied the same `alt` text used there.

- Support Open Graph images

### DIFF
--- a/app/helpers/metadata_helper.rb
+++ b/app/helpers/metadata_helper.rb
@@ -1,8 +1,21 @@
 module MetadataHelper
+  include Webpacker::Helper
+
   def meta_tag(key:, value:, opengraph: false)
     return if value.blank?
 
     tag.meta(name: prepend_opengraph(key, opengraph), content: value)
+  end
+
+  def image_meta_tags(base_url:, image_path:, alt:, opengraph: true)
+    return if image_path.blank?
+
+    image_url = "#{base_url}#{asset_pack_path(image_path)}"
+
+    safe_join([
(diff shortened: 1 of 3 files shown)


app/helpers/metadata_helper.rb
```diff
@@ -1,10 +1,23 @@
...
   end
 
+  def image_meta_tags(base_url:, value:, opengraph: true)
+    return if value.blank?
+
+    image_url = "#{base_url}#{asset_pack_path(value)}"
+
+    safe_join([
+      meta_tag(key: "image", value: image_url, opengraph: opengraph),
+      meta_tag(key: "image:alt", value: "Photograph of teaching taking place in a classroom", opengraph: opengraph),
```
leoapost, line 17: does the `image:alt` value needs to be dynamic too?

ethax-ross: Yeah that would be sensible 🤦

ethax-ross: I've fixed up the change ^

leoapost (approved):
Looks good

PR #1518, opened 2021-07-20
Use default Prometheus client data store

ethax-ross:
### Trello card

[Trello-1713](https://trello.com/c/24EwocwF/1713-switch-prometheus-client-to-use-redis-as-the-data-store)

### Context

We are seeing high CPU usage over time with the `DataFileStore`; it turns out that we don't run Puma in
a multi-process environment and so we can use the default store, which is in-memory.

- Use default Prometheus data store

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -16,25 +16,9 @@ require "action_view/railtie"
 require "rails/test_unit/railtie"
 require "view_component/engine"
 
-require "prometheus/client/data_stores/direct_file_store"
-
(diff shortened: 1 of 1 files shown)


sfawcett123 (dismissed):

peteryates (approved):
Looks great.

```ruby
# This horror is explained on `synchronize`
```
😂

peteryates (approved):
👍🏽

PR #1531, opened 2021-07-22
Updates to book a callback feature

ethax-ross:
### Trello card

[Trello-1714](https://trello.com/c/duIFg3TN/1714-updates-to-the-book-a-callback-feature)

### Context

The book a callback feature is mostly done already and only enabled in the test environment. This PR adds two additional steps to the wizard:

- We want to gather 'talking points' for the call so that the adviser is aware of what the candidate wants to discuss ahead of calling them back.
- We only accept existing candidates signing up to the callback service; we need an additional step after they attempt to 'login' that handles if we were unable to match their candidate details.
  - On the first 2 failed attempts we will encourage them to check their details and try again
  - On subsequent attempts we will inform them that we are unable to process their request

- Add talking points step to callback wizard

When signing up for a callback we want to let the user select the reason for their callback request.

Show a pre-defined list of talking points for the call to select from.

- Track number of failed matchback attempts

We want to track the number of times a user fails to matchback as part of the `IssueVerificationCode` concern, so that we can have custom copy on the book a callback feature for candidates that are failing matchback.

We are also tracking the last failure code from the CRM so that we can tailor the copy depending on if the user wasn't found vs the CRM is unavailable.

- Add matchback failed step to callback wizard

We can only support existing candidates booking a callback (not new candidates). As we only run a CTA for the book a callback form on the mailing list completion page it should always be the case that the candidate exists in the system. That being said, the sign in may fail for a couple of reasons:

```
- The candidate has entered a different first/last/email to what the signed up
  with.
- The candidate got to the form without going through the mailing list sign up
  (someone sent them a direct link, perhaps)
- The CRM is offline
```

To deal with this, if the user has attempted to log in (unsuccessfully) < 3 times we will show a message encouraging them to try again. On their third attempt we will display a message informing them that we can't accept their sign up at the moment.

If the CRM is unavailable (responded with a non-404 HTTP status code) then we inform the user we are having technical difficulties and they should try at a later time.

- Include new steps in wizard and update specs

Add the `MatchbackFailed` and `TalkingPoints` steps to the wizard flow and update the feature specs accordingly.

- Update copy on personal details/callback steps

### Guidance to review

https://review-get-into-teaching-app-1531.london.cloudapps.digital/callbacks/book/personal_details

After I finished putting in the change to determine if the CRM is unavailable I realised that users probably won't hit this screen as the API will fail silently (return 404 to signify the candidate couldn't be folund). It will work if the API is broken for any reason though, so I'm going to leave it in for that edge case.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,12 +1,12 @@
 GIT
   remote: https://github.com/DFE-Digital/get-into-teaching-api-ruby-client.git
-  revision: 85622bdfeb43e5764ac7c7a1b49f284429cf5004
+  revision: 87844c0e514093f23e951795efcb1719ddffbae4
   specs:
     get_into_teaching_api_client (1.1.17)
       addressable (~> 2.3, >= 2.3.0)
       json (~> 2.1, >= 2.1.0)
       typhoeus (~> 1.0, >= 1.0.1)
-    get_into_teaching_api_client_faraday (0.1.38)
+    get_into_teaching_api_client_faraday (0.1.39)
       activesupport
       faraday
       faraday-encoding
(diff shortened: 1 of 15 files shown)


leoapost (approved):
Looks good to me

PR #1535, opened 2021-07-22
Add draft status to posts

peteryates:
There are a few pages that are either in development or old and ready to be dropped that we don't want to appear in Google searches or generally be accessed by visitors.

Hiding them from the sitemap should help on that front.

We may at some point want to also withhold draft pages from the navigation, that can be done in a similar manner.

### Context and changes

- Rename sitemap helper to navigation helper
- Stop draft pages from appearing in the sitemap
- Set the welcome guide and guidance page as draft

### DIFF
--- a/app/views/content/guidance/financial-support-for-international-applicants.md
+++ b/app/views/content/guidance/financial-support-for-international-applicants.md
@@ -1,5 +1,6 @@
 ---
 title: Financial support for international applicants
+draft: true
 ---
 
 This page has moved.
--- a/app/views/content/welcome.md
+++ b/app/views/content/welcome.md
@@ -1,5 +1,6 @@
 ---
 title: Welcome to teaching
+draft: true
 layout: "layouts/welcome"
 content:
   - "content/welcome/landing/welcome-to-teaching"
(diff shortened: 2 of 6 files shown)


app/views/content/guidance/financial-support-for-international-applicants.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Financial support for international applicants
+draft: true
```
MylesJarvis, line 3: Does this need to be draft still? I see we've published it now @peteryates

app/views/content/welcome.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Welcome to teaching
+draft: true
```
MylesJarvis, line 3: Does this mean we should build in removing/reverting the env variables, which you added as a separate PR yesterday, to this PR? @peteryates

peteryates: No, this will just stop the pages from appearing in the sitemap.

The if statement in the welcome guide will prevent the page from rendering if someone manually navigates to `/welcome`

Thinking about it, it might be worth preventing `draft` pages from being rendered at all (or being searchable) in production, but that's a bigger change and would be a separate PR.

app/views/content/guidance/financial-support-for-international-applicants.md
peteryates: I marked it as draft because we don't really want it to appear in the sitemap, it only exists for people who've bookmarked the page (or have the URL in an email or something)

MylesJarvis (approved):

PR #1536, opened 2021-07-22
Hide footer distractions from registation layout

peteryates:
This brings the mailing list steps two onwards in line with the first and will make the event signup screens a little more minimal.


MylesJarvis (approved):
LGTM

PR #1537, opened 2021-07-23
Allow searching by all months

ethax-ross:
We default to the current month in the search; this has been fine but now we have TTT events that start in a few months time and we find users click the search with the default month values most often. So that TTT events are in the results we are changing the month selector to default to 'all months'.

Its a tricky one to test as the review apps point to dev and there are no TTT events on dev yet. I've tested it locally against the test API instance though and it seems to work fine.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -90,7 +90,7 @@ private
   end
 
   def event_search_params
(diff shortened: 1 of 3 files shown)


peteryates (approved):
Looks good

PR #1539, opened 2021-07-26
Improve font size assignment

peteryates:
#### Changing the font sizing convention

Our current font sizes implementation ties them to actual font sizes (based on 16px), which means if we adjust them the naming convention makes no sense. Now they're just an array of sizes, from `0.9rem` to `5.3333rem`, and we can modify them at will.

### DIFF
--- a/app/webpacker/styles/font-sizes.scss
+++ b/app/webpacker/styles/font-sizes.scss
@@ -4,45 +4,45 @@ html {
   @include mq($from: tablet) { font-size: 100%; }
 }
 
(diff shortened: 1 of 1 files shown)


jkempster34 (commented):
LGTM 👁️ .

Wondering if we could make some of this stuff private (I didn't know you could do that is Sass)?

image

peteryates:
Thanks for the tip @jkempster34, didn't realise that was a thing in Sass!

jkempster34 (approved):
🥇

PR #1548, opened 2021-07-28
Attach external link icons in Ruby

jkempster34:
### Trello card
https://trello.com/c/QSHLpZlP

### Context
External link icons are currently set in CSS. However, they need an alt attribute, and while this can be done in CSS, it is not supported by all browsers.

- Apply external link icons in Ruby instead of CSS

### Guidance to review
In CSS, the external link icon was being suppressed in several places. I have only included these in the `markdown` class, supressing the `button` class. Flicking through I can't see any difference.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -57,12 +57,17 @@ private
     doc = Nokogiri::HTML(body)
     markdown_links = doc.css(".markdown a")
     classes_to_suppress = %w[button]
+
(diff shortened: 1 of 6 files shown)


peteryates (approved):
This works really well.

Initially I thought it would make more sense for this functionality to live in the `TemplateHandlers::Markdown` but perhaps here makes more sense - I suspect some events have them etc.

Nice 👍🏽

peteryates (approved):

PR #1550, opened 2021-07-28
Fix and run ESLint

jkempster34:
### Context
Tidy up - Fix and run JS linter

- Update ESLint parser name. `babel-eslint` has been deprecated in favour of `@babel/eslint-parser`
- Run ESLint

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -27,4 +27,7 @@ module.exports = {
       },
     },
   ],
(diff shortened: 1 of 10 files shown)


ethax-ross (approved):
Looks good, can we run the linter as part of CI to avoid having to do this again in the future?

Those semi colons though 🙈

PR #1551, opened 2021-07-28
Native support for Question Time events

ethax-ross:
[Trello-1718](https://trello.com/c/pHVDpbZ2/1718-implement-front-end-changes-for-question-time-events)

Currently the GiT API remaps Question Time events with a Train to Teach event type. This is because the events team want the QT and TTT events both displayed as if they are TTT events on the GiT website.

This, however, is a bit of confusing and potentially hazardous code in the GiT API and we would prefer to remove this and support QT events natively in the GiT website.

This commit achieves that by:

- Updating the `Events::GroupPresenter` to group TTT and QT events for   displaying in the event search results.

- As a byproduct of this ^ we need the presenter to now know about how many   events there should be per-type (as we're combining so could end up with more than expected).

- Updating the `Events::Search` model to remove QT events as a searchable   option and query both TTT and QT events if searching for TTT events.

- Mis-representing QT events as TTT event types in the event box by   providing translations.

The API was previously updated to support querying by multiple `type_ids` so that when searching we can perform a single search for both QT and TTT events.

As the QT events are being mapped to TTT events in the API, once this PR has shipped we will need to undo that bit of logic and give it a thorough test in dev to ensure its working as expected.

### DIFF
--- a/app/models/events/search.rb
+++ b/app/models/events/search.rb
@@ -33,9 +33,13 @@ module Events
 
     class << self
       def available_event_types
-        @available_event_types ||= GetIntoTeachingApiClient::Constants::EVENT_TYPES.map do |key, value|
-          GetIntoTeachingApiClient::PickListItem.new(id: value, value: key)
-        end
+        # We can't search for Question Time events explicitly. Instead, they
+        # are returned as Train to Teach events.
+        @available_event_types ||= GetIntoTeachingApiClient::Constants::EVENT_TYPES
+          .except("Question Time")
+          .map do |key, value|
+            GetIntoTeachingApiClient::PickListItem.new(id: value, value: key)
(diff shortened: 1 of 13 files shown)


app/models/events/search.rb
```diff
@@ -89,7 +99,7 @@ def today_start_of_month?
 
     def query_events_api(limit)
       GetIntoTeachingApiClient::TeachingEventsApi.new.search_teaching_events_grouped_by_type(
-        type_id: type,
+        type_ids: type_ids,
```
ethax-ross, line 102: This isn't currently supported by the API, but we could do this to support querying multiple event types

jkempster34 (approved):
LGTM 💯

PR #1560, opened 2021-07-30
Improve the nav link display

peteryates:
The nav link focus styles were a little bit broken, mainly due to them being `inline` instead of `block`, so when they wrapped the background colour from one line overlapped the other.

| Before | After |
| ----- | ------ |
| Screenshot from 2021-07-30 13-30-36 | Screenshot from 2021-07-30 13-30-08 |
| Screenshot from 2021-07-30 11-09-48 | Screenshot from 2021-07-30 11-12-17 |

### DIFF
--- a/app/webpacker/styles/header/navigation.scss
+++ b/app/webpacker/styles/header/navigation.scss
@@ -6,6 +6,7 @@
   a {
     color: $black;
     text-decoration: none;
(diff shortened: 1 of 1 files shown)


MylesJarvis (approved):
LGTM

PR #1582, opened 2021-08-02
Make Scribble ID optional when creating online events

jkempster34:
### Context
Not all Online events are Scribble events. Non-Scribble events are rare, but this change will enable these to be created via the form instead of having to use the CRM. 
The only difference between Scribble and non-Scribble events is that non-Scribble don't have a Scribble ID.

Make Scribble ID optional when creating online events


peteryates (approved):
Makes sense to me

PR #1587, opened 2021-08-03
Update ML Completion Page Copy

MylesJarvis:
Stakeholder amends to ML completion page.


ethax-ross (approved):
Looks good, though I think a test may fail 🤞

PR #1588, opened 2021-08-03
Display confirmation in correct time zone

ethax-ross:
### Trello card

[Trello-1775](https://trello.com/c/0mJTnskW/1775-fix-booking-confirmation-page-being-an-hour-out-callback-feature)

### Context

When a user books a callback we display the slots available in London/Europe timezone, however when we display the selected time on the confirmation page we were outputting the UTC time.

Convert the chosen time from UTC to London/Europe prior to displaying on the confirmation page.

- Display confirmation in correct time zone

### DIFF
--- a/app/controllers/callbacks/steps_controller.rb
+++ b/app/controllers/callbacks/steps_controller.rb
@@ -16,7 +16,7 @@ module Callbacks
     end
 
     def completed_step_path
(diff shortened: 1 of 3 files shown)


jkempster34 (approved):
LGTM 👍

PR #1589, opened 2021-08-03
Hide the cookie popup for Silktide tests

peteryates:
Our score is artificially low because the overlay rendered with our cookie popup affects the bot's ability to focus and click on certain elements.


ethax-ross (commented):
Looks like you have a spec failure, though I'm not sure if its related... that being said could have a wee spec for this functionality? 😅

ethax-ross (approved):

PR #1593, opened 2021-08-03
Edit published events

jkempster34:
### Trello card
https://trello.com/c/on2SgkXj/1774-further-events-form-updates

### Context
Sometimes the details of `open` events can change. Editing them via the website is currently not possible and must be done in the CRM.

- Add guidance box to tell elevated users (publisher role) how to edit a published event
- Add a `withdraw` action - We have agreed that it will be fine for events to go offline while they are being edited (this also adds a level of safety because events can be checked again before being made live). Add a withdraw action so that the event can be returned to pending status, and will appear back in the internal events section where it can be edited.
- Add a page of `open` events - Each event has a link the event on the main website, plus a link to withdraw. When an event is withdrawn, we show a banner linking the user to the now `pending` event.

Some refactoring: 
  - Tidy up most of the factories
    - Move some common attributes into traits
    - Stop using FactoryBot to create expected request bodies. Remove these and use modify other existing values instead
  - The `success` param is now being used to inform the `index` banner of what event has been created (so that the user can be directed there). Rename to `status` which makes more sense now.
  - Move event status banner text to locales - Previously, we have been passing instance variables to inform the user of the status of the event. Now that the number of statuses has grown, shift it into locales instead.

### Guidance to review
I am filtering the events by type in the `open_events` action. Its probably best to open a new ticket to update the `search_teaching_events_grouped_by_type` API method to accept an array of types like it does for the `status_id`s

### DIFF
--- a/app/controllers/internal/events_controller.rb
+++ b/app/controllers/internal/events_controller.rb
@@ -62,6 +62,20 @@ module Internal
       )
     end
 
+    def live_events
+      events = GetIntoTeachingApiClient::TeachingEventsApi
+                 .new
+                 .search_teaching_events_grouped_by_type(quantity_per_type: 1_000,
+                                                         start_after: DateTime.now.utc.beginning_of_day,
+                                                         status_ids: [open_event_status_id])
+      events = events.select do |key|
+        key.type_id == event_types[:provider] || key.type_id == event_types[:online]
+      end
+      if events[0] && events[1]
+        @live_events = events[0].teaching_events.concat(events[1].teaching_events)
+      end
+    end
+
     def create
       @event = Event.new(event_params)
       @event.assign_building(building_params) unless @event.online_event?
@@ -131,6 +145,10 @@ module Internal
 
     def pending_event_status_id
       GetIntoTeachingApiClient::Constants::EVENT_STATUS["Pending"]
+    end
+
+    def open_event_status_id
+      GetIntoTeachingApiClient::Constants::EVENT_STATUS["Open"]
     end
 
     def event_types
--- a/app/controllers/internal/events_controller.rb
+++ b/app/controllers/internal/events_controller.rb
@@ -13,10 +13,7 @@ module Internal
       load_pending_events(@event_type)
       @no_results = @events.blank?
 
-      @successful = params[:status]
-      @withdrawn = params[:status] == "withdrawn"
-      @published = params[:status] == "published"
-      @pending = params[:status] == "pending"
+      @status = params[:status]
       @readable_id = params[:readable_id]
     end
 
--- a/app/controllers/internal/events_controller.rb
+++ b/app/controllers/internal/events_controller.rb
@@ -62,18 +62,17 @@ module Internal
       )
     end
 
-    def live_events
+    def open_events
       events = GetIntoTeachingApiClient::TeachingEventsApi
                  .new
                  .search_teaching_events_grouped_by_type(quantity_per_type: 1_000,
                                                          start_after: DateTime.now.utc.beginning_of_day,
                                                          status_ids: [open_event_status_id])
+
       events = events.select do |key|
         key.type_id == event_types[:provider] || key.type_id == event_types[:online]
       end
-      if events[0] && events[1]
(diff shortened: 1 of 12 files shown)


sfawcett123 (dismissed):

app/controllers/internal/events_controller.rb
```diff
@@ -14,6 +14,8 @@ def index
       @no_results = @events.blank?
 
       @success = params[:success]
+      @withdrawn = params[:success] == "withdrawn"
```
ethax-ross, line 17: I would maybe rename this param to be `status` or something, `success` doesn't really make sense with respect to the value

app/controllers/internal/events_controller.rb
```diff
@@ -42,12 +44,38 @@ def approve
...
+    end
+
+    def live_events
+      events = GetIntoTeachingApiClient::TeachingEventsApi
+                 .new
+                 .search_teaching_events_grouped_by_type(quantity_per_type: 1_000,
+                                                         start_after: DateTime.now.utc.beginning_of_day,
+                                                         status_ids: [open_event_status_id])
+      events = events.select do |key|
+        key.type_id == event_types[:provider] || key.type_id == event_types[:online]
```
ethax-ross, line 72: I agree this would be nicer to bundle into the search query.

app/controllers/internal/events_controller.rb
```diff
@@ -42,12 +44,38 @@ def approve
...
+      end
+      if events[0] && events[1]
```
ethax-ross, line 74: What if there are only live events of one type? Could you do something like:

```
@live_events = events.map(&:teaching_events).flatten
```

Does it make sense to use the CRM notion of 'open' instead of 'live' for consistency?

spec/features/internal_spec.rb
```diff
@@ -3,24 +3,24 @@
 
 RSpec.feature "Internal section", type: :feature do
   let(:types) { Events::Search.available_event_type_ids }
-  let(:provider_event) do
+  let(:pending_provider_event) do
     build(:event_api,
```
ethax-ross, line 7: Worth moving some of this setup as a trait on the factory?

spec/requests/internal/events_controller_spec.rb
```diff
@@ -531,4 +559,158 @@
...
+      end
+
+      context "when provider event" do
+        before do
+          allow_any_instance_of(GetIntoTeachingApiClient::TeachingEventBuildingsApi)
+            .to receive(:get_teaching_event_buildings) { [] }
+        end
+        let(:event) { pending_provider_event }
+        let(:expected_request_body) do
+          build(:event_api,
```
ethax-ross, line 578: I wonder if we could move this to a factory as well, something like:

```
build(:event_api, with_event: event)
```

ethax-ross (commented):
Looking good, just a few suggestions

app/views/internal/events/index.html.erb
```diff
@@ -1,23 +1,26 @@
...
-                          @pending ? internal_event_path(@readable_id) : event_path(@readable_id) %>
-            </strong>
-          </div>
-        <% end %>
-      </div>
+<% if @successful %>
+  <div class="confirmation-panel-container">
+    <div class="govuk-panel govuk-panel--confirmation">
+      <h1 class="govuk-panel__title">
+        <%= "Event submitted for review" if @pending %>
```
ethax-ross, line 5: Can be picked up later, but I think the text dependent on status could be cleaner if we moved the text to the translations file and had something like:

```
<h1 class="govuk-panel__title">
<%= t("some.key.#{@status}") %>
</h1>
```

spec/features/internal_spec.rb
```diff
@@ -163,6 +163,40 @@
     end
   end
 
+  context "when there are live events" do
+    start_at = Time.zone.today.at_beginning_of_month + 1.day
```
ethax-ross, line 167: let(:start_at) {...}

ethax-ross (approved):
LGTM

PR #1602, opened 2021-08-04
Add https://googleads.g.doubleclick.net to CSP

ethax-ross:
We are seeing this violation in the reports and as there is no subdomain the current wildcard directive may not be matching.

### DIFF
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -9,7 +9,7 @@ SecureHeaders::Configuration.default do |config|
 
   tta_service_hosts = []
   tta_service_hosts << URI.parse(ENV["TTA_SERVICE_URL"]).host if ENV["TTA_SERVICE_URL"].present?
-  google_analytics = %w[*.google-analytics.com *.googletagmanager.com *.googleusercontent.com *.gstatic.com s.ytimg.com *.googleadservices.com *.googleads.g.doubleclick.net]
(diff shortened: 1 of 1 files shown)


MylesJarvis (approved):
LGTM

config/initializers/secure_headers.rb
```diff
@@ -9,7 +9,7 @@
 
   tta_service_hosts = []
   tta_service_hosts << URI.parse(ENV["TTA_SERVICE_URL"]).host if ENV["TTA_SERVICE_URL"].present?
-  google_analytics = %w[*.google-analytics.com *.googletagmanager.com *.googleusercontent.com *.gstatic.com s.ytimg.com *.googleadservices.com *.googleads.g.doubleclick.net]
+  google_analytics = %w[*.google-analytics.com *.googletagmanager.com *.googleusercontent.com *.gstatic.com s.ytimg.com *.googleadservices.com *.googleads.g.doubleclick.net https://googleads.g.doubleclick.net]
```
peteryates, line 12: should it have the `https://` here or should it just be `googleads.g.doubleclick.net`?

ethax-ross: I think it was in with the `https://` previously, though it probably works with both 🤷‍♂️

peteryates (approved):
Looks good to me

PR #1604, opened 2021-08-04
Minor CSP tweaks for Facebook, Snapchat and LID

ethax-ross:
Some minor tweaks to the directives for Facebook, Snapchat and LID tracking to make the matching a bit looser and cover extra subdomains.


MylesJarvis (approved):
LGTM

PR #1607, opened 2021-08-05
Update frame-src field to add GA properties 

MylesJarvis:
Noticed we're getting some errors from `googlesyndication` on frame-src for ML completion page, so adding it in.

Can you check this is okay from a security perspective @ethax-ross ?


ethax-ross (dismissed):
We may want to try and pair some of this back at some point (as this will add all the GA domains) but should be fine for now 👍

ethax-ross (approved):

PR #1608, opened 2021-08-05
Disable Turbolink to see analytics impact

peteryates:
This is a temporary change intended to be reverted after a short time!

### DIFF
--- a/app/webpacker/controllers/link_controller.js
+++ b/app/webpacker/controllers/link_controller.js
@@ -4,20 +4,20 @@ export default class extends Controller {
   static targets = ['content'];
 
   connect() {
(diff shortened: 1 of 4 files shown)


ethax-ross (approved):
Looks fine to me and clicking about all seems to work still. I wonder if the analytics will be effected as something is triggered on `turbolinks:load` but I _think_ its fine as that wouldn't cover the initial page load anyway, so having Turbolinks disabled just means we use the same mechanism as the first page load.

peteryates:
Shall we give this a try first thing tomorrow morning and keep an eye on it for a couple of hours?

cc @MylesJarvis

PR #1610, opened 2021-08-05
Tweak CSP for Google services

ethax-ross:
- Add extra CSP entries for GTM preview
- Extra CSP entries for Universal Analytics
- Extra entry for Google Ads remarketing


peteryates (approved):
Looks fine to me. `data:` rings alarm bells but as we're limiting it to `font-src` it should be ok.

PR #1615, opened 2021-08-09
Move Anchore scan to new file with cron schedule

peteryates:
### Trello card

N/A

### Context

The Anchore security scan is currently run for every PR. This doesn't really make sense because in most cases it's not going to be the PR that will have caused failures, but general security flaws being uncovered in our images, libraries and gems. These are unlikely to be fixed within the same PR.

As a result, it _probably_ makes more sense to run the scan nightly against `master` instead. This is the first step, it's simply duplicating the relevant part of `build.yml` in `anchore.yml`. When we're happy this is running correctly the step from `build.yml` can be dropped.

### DIFF
--- a/.github/workflows/anchore.yml
+++ b/.github/workflows/anchore.yml
@@ -1,0 +1,44 @@
+name: Nightly Anchore Security Scan
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


leoapost (approved):
Seems sensible

PR #1619, opened 2021-08-10
Make Gemma the owner of GIT content and images

peteryates:
### Trello card

N/A

### Context

Myles has moved on and Gemma has taken his place. He probably doesn't want all these notifications 😅


MylesJarvis (approved):
LGTM - although v weird being asked to sign my own power away

PR #1621, opened 2021-08-10
Update the pictures in the signup page

leoapost:
### Trello card
https://trello.com/c/3yw5wapm

### Context
We want to update the pictures in the sign up page to improve the looks.

Replace the right picture because it's out of focus and looks like low quality. Move the left picture to the right so it doesn't look that distant.

### Guidance to review

|before|after|
|-|-|
|before|after|


peteryates (approved):
Looks good!

PR #1633, opened 2021-08-11
Ensure footer is at bottom of viewport on short pages

peteryates:
### Trello card

https://trello.com/c/FVynHujX/1768-footer-is-too-high-on-mailing-list-signup

### Context

We currently have a bug where the `<main>` element being short means that the footer isn't at the bottom of the screen. This looks bad but is admittedly a minor problem.

The simple fix is to make `<body>` [flex](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow) and allow `<main>` to grow. This will push the footer to the bottom of the viewport. I've also made the content area a tiny bit wider on bigger screens.

- Make body element display as a flex column
- Make content container a bit wider

### Guidance to review

Probably worth giving this a closer look in Safari to make sure nothing's horribly broken - it shouldn't be!

### Preview

| Before | After |
| ------ | ------- |
| Screenshot from 2021-08-11 16-02-48 | Screenshot from 2021-08-11 16-02-50|

### DIFF
--- a/app/components/footer/talk_to_us_component.html.erb
+++ b/app/components/footer/talk_to_us_component.html.erb
@@ -1,35 +1,37 @@
-<section class="talk-to-us" data-controller="talk-to-us">
-  <section class="container">
-    <div>
-      <h2 class="purple" id="talk-to-us">Talk to us</h2>
-      <div class="content">
-        <section class="contact">
-          <p>Monday to Friday between <strong>8:30am and 5:30pm</strong></p>
-          <div class="options">
-            <div>
-              <p>Live chat</p>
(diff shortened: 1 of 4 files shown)


gemmadallmandfe:
I've looked at this in Edge and Chrome on desktop and Safari on an iPhone. I can't actually replicate the original issue without zooming out to 50% but the fix looks fine to me.

Given your comment re Safari, I haven't set to Approve as I don't have access to a Mac yet, but happy for someone else to!

ethax-ross (approved):
Looks good, checked on Safari Mac and seems fine 👍

PR #1638, opened 2021-08-12
Fix the blog main heading

peteryates:
### Trello card

N/A

### Context

I inadvertantly broke this in #1635, we have some fragility in our front matter handling when it comes to symbols vs strings.

- Make blog heading not care about frontmatter format
- Add spec that ensures there's one heading per page

### Guidance to review

Is the blog heading now present?


app/controllers/blog_controller.rb
```diff
@@ -5,7 +5,7 @@ class BlogController < ApplicationController
   layout "layouts/blog/index"
 
   def index
-    @front_matter = { "title" => "Blog" }
+    @front_matter = { "title" => "Get Into Teaching Blog" }
```
leoapost, line 8: Haven't tried it, but shouldn't it work if we just symbolise this key and keep the rest as they are? :thinking:

peteryates: I think we use strings everywhere except here. It's unintuitive though, and easy to miss as `@front_matter[:something]` will return nil.

leoapost: Makes sense to change the symbols to strings then.

leoapost (approved):
Looks good

PR #1644, opened 2021-08-16
Use event list anchor for pagination links

peteryates:
### Trello card

https://trello.com/c/C4rF9UWs/1858-fix-scroll-position-after-navigating-to-a-different-numbered-events-page

### Context and changes

- Anchor pagination links to the top of current list
- Remove the now-unused scroll controller

Previously we used a JavaScript plugin to _remember_ how far we've scrolled, so the pagination isn't jarring (#969), but this can be unpredictable - if one page has fewer events on it, the user might end up looking at a different category or at another part of the page (usually the footer).

This is an attempt to improve that behaviour by adding anchors to the events list(s) and adding an `anchor` param to the paginator. Now, on every page change, the browser will anchor itself to the top of the current list.

### Guidance to review

Is the new behaviour an improvement? It's not exactly as described in the Trello ticket because that would be even more jarring for users when using pagination on the search results screen. Keep in mind, each of the three events lists can be separately-paginated **on the same page**, so moving them to the top of the page would mean they'd need to scroll (a long way) to find the list they're interested in.


gemmadallmandfe:
Thanks @peteryates - I had missed that there were three sets of pagination on /events but your fix to go to the top of the relevant section makes sense for now, until we review the whole Events page design more thoroughly.

**/events**
Fix looks good for the School and Uni events section, both on desktop and mobile. I can't test the other sections as there are not enough test events but presume they will work in the same way.

**/event-categories/train-to-teach-events
/event-categories/online-q-as
/event-categories/school-and-university-events**
Fix looks good for the School and Uni page as far as I can tell - I can't test the other sections as there are not enough test events but presume they will work in the same way.

ethax-ross (approved):
I think this works better overall 👍 unrelated to this, but when did the pagination controls get so enormous?

Screenshot 2021-08-19 at 09 23 11

peteryates:
@ethax-ross in #1595 🙈 

They were failing the WCAG contrast check at the current size, the option was bolder or bigger and we opted for bigger.

gemmadallmandfe:
@ethax-ross fair point re the pagination control size but something we can look at separately with the wider design changes to the events hub pages and the brand more generally (at least we are more WCAG compliant in the interim!)

PR #1645, opened 2021-08-17
Fix typos in the SCITT acronym

peteryates:
### Context and changes

These were copied over from the old get-into-teaching-content repo


gemmadallmandfe (approved):
Looks good to me!

PR #1647, opened 2021-08-17
Allow page headings and titles to be set independently

peteryates:
### Trello card

https://trello.com/c/rgeIda1X/1861-separate-title-and-h1-to-give-greater-flexibility-for-seo

### Context

Sometimes it might be beneficial to be able to set the page title and heading separately.

#### How it fits together

cc @gemmadallmandfe 

Now there are three heading/title related settings available in the front matter:

* `title` - the minimum that's required, when only it is set this value will form the page title, the main heading (`<h1>`) on the page and the text inside the navigation menu
* `navigation_title` - when set, it'll override the `title` in the navigation menu. Used on the homepage where `title: Inspire the next generation` and `navigation_title: Home`
* `heading` - when set it'll override the `title` in the main page header (usually in the hero)

- Allow front matter heading to override title
- Update the salaries and benefits heading

### DIFF
--- a/app/components/header/hero_component.rb
+++ b/app/components/header/hero_component.rb
@@ -6,7 +6,7 @@ module Header
       return if front_matter.blank?
 
       front_matter.with_indifferent_access.tap do |fm|
-        @title           = fm["title"]
+        @title           = fm["heading"] || fm["title"]
         @subtitle        = fm["subtitle"]
         @subtitle_link   = fm["subtitle_link"]
         @subtitle_button = fm["subtitle_button"]
--- a/app/models/pages/navigation.rb
+++ b/app/models/pages/navigation.rb
@@ -34,7 +34,7 @@ module Pages
         @path = front_matter[:navigation_path] || path
 
         front_matter.tap do |fm|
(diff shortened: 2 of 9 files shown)


gemmadallmandfe:
Thanks @peteryates - I can see that <title> and <h1> are separate on the salaries page

Would this always change the menu option in the nav bar as well to be title or is the nav option set independently (eg Teaching events seems to be different again!)? Given that we are in the middle of reviewing the whole nav, I don't want to be too gung-ho about any changes; I need to understand how it all fits together!

peteryates:
Ah no, currently the nav will match the title. It'd make more sense to match the header, I'll change that tomorrow.

gemmadallmandfe (commented):
Looks good @peteryates - I can see that <title> is different to <h1> in the source for the Salaries and Benefits page.

Before merging, please can you:
- change back the <title> on the Salaries and benefits page (we need to do a proper SEO review!)
- confirm that when this is deployed, <title> will be the same as <h1> as default (presume so)

peteryates:
I've reset them to how they were before:


Screenshot-2021-09-07-10:04:42

gemmadallmandfe (approved):
Looks good to me

PR #1650, opened 2021-08-18
Tidy up CSP headers

peteryates:
### Trello card

https://trello.com/c/KAhgKi98/1792-clean-up-secureheaders-module

### Context and changes

This file had grown a bit unwieldy, especially with recent chops and changes. We're not changing too much in the way of content here (yet), but grouping by domain should make the next stage of removing some unwanted ones easier.

- Split up CSP wildcards into groups and reorganise
- Rubocop prefers to join filenames by string
- Remove the commented out original lines

### Guidance to review

See b41fc42668c578a8fff for a mostly-complete version with the original lines left in (but commented out) for easy comparison.


config/csp/google_supported_domains.csv
```diff
@@ -0,0 +1,190 @@
+.google.com
```
leoapost, line 1: I believe #1650 added these domains as an attempt to solve the analytics issue we had few weeks ago? If I recall correctly, the issue was fixed and it was unrelated to the introduced changes, so maybe we should clean them up?

ethax-ross: It may not have caused that particular issue but we still had CSP violations against some of these domains; I think given we can't control which domains Google are using (and it probably changes now and then) we should leave them in to be safe

ethax-ross (dismissed):
Vast improvement 💯 looks like you have a spec failure but seems unrelated so probably flakey

config/csp/google_supported_domains.csv
leoapost: ah ok then :)

peteryates:
Ah, the bug's legit - looks like a it's complaining about some policies. Will fix 🤦🏽‍♂️

ethax-ross (approved):

PR #1652, opened 2021-08-19
Some more CSP tweaks

peteryates:
- Add google_analytics to img_src domains list
- Add another pinimg wildcard (matching old syntax)
- Add a wider facebook.net wildcard
- Add google_doubleclick to img_src list

### DIFF
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -61,7 +61,7 @@ SecureHeaders::Configuration.default do |config|
     form_action: ["'self'"].concat(snapchat, facebook, vwo, govuk),
     frame_src: ["'self'"].concat(scribble, snapchat, facebook, youtube, hotjar, google_doubleclick, vwo, google_analytics),
     frame_ancestors: ["'self'"],
-    img_src: ["'self'"].concat(govuk, pinterest, facebook, vwo, youtube, twitter, google_supported, google_adservice, google_apis, google_analytics, data, lid_pixels, %w[cx.atdmt.com linkbam.uk]),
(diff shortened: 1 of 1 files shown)


config/initializers/secure_headers.rb
```diff
@@ -32,11 +32,11 @@
   google_doubleclick = %w[*.doubleclick.net *.googleads.g.doubleclick.net *.ad.doubleclick.net *.fls.doubleclick.net stats.g.doubleclick.net]
   google_apis        = %w[*.googleapis.com https://fonts.googleapis.com]
 
-  facebook  = %w[*.facebook.com *.connect.facebook.net]
+  facebook  = %w[*.facebook.com *.facebook.net *.connect.facebook.net]
   govuk     = %w[*.gov.uk www.gov.uk]
   hotjar    = %w[*.hotjar.com vc.hotjar.io wss://*.hotjar.com]
   jquery    = %w[code.jquery.com]
-  pinterest = %w[*.pinterest.com ct.pinterest.com *.s.pinimg.com]
+  pinterest = %w[*.pinterest.com ct.pinterest.com *.pinimg.com *.s.pinimg.com]
```
ethax-ross, line 39: I would have expected `*.pinimg.com` to match `*.s.pinimg.com` - sometimes I think it takes a while for CSP changes to propagate as well 🤷‍♂️

ethax-ross (approved):

config/initializers/secure_headers.rb
peteryates: Yeah, I'd have thought so too. I compared it to the list before yesterday's changes and this is how it was before.

peteryates: Agree, will remove the `s` prefix one

ethax-ross (approved):

PR #1654, opened 2021-08-19
Add an extra assertion to redirects spec

peteryates:
### Trello card

N/A

### Context and changes

The new assertion does a `GET` to ensure the target pages is definitely there. Skip any pages where we contact the CRM (dates and privacy policy).

This highlighted an incorrect rule that's a result of an old duplicate story that was removed several months ago - it has been removed.


gemmadallmandfe:
Thanks for flagging this @peteryates - makes sense to me. Not sure how I can test it to approve though!

peteryates:
@gemmadallmandfe - ah don't worry about that, just cc'd you on it as we were discussing it yesterday.

This just checks everything's in order before we merge/release anything.

ethax-ross (approved):

PR #1656, opened 2021-08-20
Make the event section layout columnar on mobile

peteryates:
### Trello card

https://trello.com/c/3QNsSmwD/1877-events-section-heading-doesnt-resize-gracefully

### Context

This prevents the logo from squishing the text into an awkward shape.  We're using column-reverse here to make the logo appear above the heading text.

| Before | After |
| ------ | ------ |
| Screenshot-2021-08-20-15:54:22 | Screenshot-2021-08-20-15:54:07 |

### Review guidance

My mouse must've gone over the Birmingham event when screenshotting on live, the black outline is from the hover state!

### DIFF
--- a/app/webpacker/styles/events.scss
+++ b/app/webpacker/styles/events.scss
@@ -216,6 +216,10 @@
         display: flex;
         justify-content: space-between;
         align-items: center;
+
+        @include mq($until: tablet) {
+          flex-direction: column-reverse;
+        }
       }
 
       &__logo {
@@ -224,6 +228,10 @@
         left: auto;
         transform: scale(.7);
         margin-right: -58px;
+
+        @include mq($until: tablet) {
+          align-self: end;
+        }
       }
     }
   }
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
Looks much more readable @peteryates! Only comment is that the image is now floating centrally, which looks a little bit odd?

Screenshot from iPhone 11
Events

peteryates:
@gemmadallmandfe ah - it looks like `align-self: end` works differently in Firefox and Chrome/Safari. I changed to `align-self: flex-end`.

gemmadallmandfe:
Looks good on iphone 11 Safari/Chrome/Edge and desktop Edge @peteryates 

Emulator looks a bit odd on tablet but can't test properly and don't think this is related to this change?

peteryates:
@gemmadallmandfe - I believe the 'oddness' is caused by the stretching of the page fixed in #1657. I've rebased this branch with that fix so when the next review app pops up it _should_ look better 🤞🏽

ethax-ross (approved):
Looks good to me 👍

gemmadallmandfe:
iPad/iPad Pro looks much better in the emulator now @peteryates  👍

PR #1657, opened 2021-08-20
Make breadcrumbs bar 100% width up to max page

peteryates:
Using max instead of min will prevent the screen from stretching on mobile.


gemmadallmandfe:
Tested on iphone 11/Safari and in emulator - resolves the issue with responsiveness on My stories into teaching child pages.

ethax-ross (approved):

PR #1663, opened 2021-08-23
Updated image

emilytraff:
Update of image and associated alt text


peteryates (approved):
Looks good and everything works.

PR #1664, opened 2021-08-24
Display YouTube previews in the correct aspect ratio

ethax-ross:
### Trello card

[Trello-1874](https://trello.com/c/74cEZHxR/1874-youtube-clips-in-blog-posts-are-awkwardly-cropped-on-mobile)

### Context

We need to be able to set the aspect ratio for YouTube videos in order to maintain a full preview thumbnail (without it being cropped).

We can do this with some CSS by wrapping the `iframe` and applying an appropriate `padding-bottom` to the wrapper.

It would have been possible to do this just in CSS by manually working out the aspect ratios for the videos and setting the appropriate padding, however I figured it would be more user friendly to be able to take the width/height and have a Stimulus controller do it for you (initially I thought the videos would be a mixture of aspect ratios but it turned out they're all the  same...). The downside to this is the few with JS disabled will still get cropped previews.

- Add AspectRatioController

- Update YouTube videos with correct aspect ratio

Apply the new `AspectRatioController` to the YouTube videos so that they are displayed with the correct aspect ratio.

### Guidance to review

| Before      | After |
| ----------- | ----------- |
| Screenshot 2021-08-24 at 10 59 22      | Screenshot 2021-08-24 at 10 59 07      |
| Screenshot 2021-08-24 at 11 01 13   | Screenshot 2021-08-24 at 11 01 00        |


gemmadallmandfe:
Looks much better @ethax-ross 

My only comments would be:

- in the emulator, the thumbnail looks pretty low quality at small sizes (eg iPhone 5)? Likely to be the case IRL?

- on my phone (iPhone 11/Safari), the thumbnail is still slightly cropped (see screenshot below)

Neither are a showstopper - not sure if there is anything you can do?

image

ethax-ross:
@gemmadallmandfe I see what you mean; it looks like Google have some responsive code that triggers at 375px (width of your iPhone screen) causing it to crop slightly. I can get around it by tweaking the aspect ratio to be slightly wider than what YouTube suggests... I've pushed up the change so the review app should rebuild shortly

gemmadallmandfe (approved):
Looks good to me - have tested on iPhone 11 (Safari / Chrome / Edge) and in emulator

(I'll leave you to merge in case you need any tech eyes on it!)

PR #1665, opened 2021-08-24
Refresh the first page of the event signup wizard

peteryates:
### Trello card

https://trello.com/c/lbTnxf5u/1769-design-improve-visual-appearance-of-train-to-teach-event-transaction-flow-startpage

### Context

The first page of the mailing list wizard recently had a refresh that had a slight uptick in performance. This makes similar changes to the mailing list one but with a single big photo at the top of the form and the 'event info' box slightly overhanging it.

- Rename attractive layout to 'with side images'
- Add layout for first page of event wizard
- Add custom styles for first page of event wizard
- Add styles for the first page of the event signup
- Add dates and times to mobile signup component

### Things that might need addressing

* fix the heading style so we don't have `<h1>` and `<h2>` alongside each other?
* improve the wizard step headings so they're more descriptive about what we're asking for on the page?

### Preview

| Mobile | Desktop |
| ------- | -------- |
| Screenshot-2021-08-24-17:10:54 | Screenshot-2021-08-24-17:10:45 |


gemmadallmandfe:
Looks good @peteryates!

Some comments/suggestions:

- Is 'Registering for' overkill? Having said that, I think it actually works well on mobile in the reduced summary box, so maybe leave as is

- I'm not sure we need 'Your attendance will be added to a register and an email will be sent.'? Sounds a bit too formal and bureaucratic - I'll check with Gary as to the rationale behind this

- I actually quite like the summary box overhanging the photo but not the form (on desktop) rather than the form as well - the latter feels a bit boxy to me

- I agree that we need to review the headings next - the current h2 is superfluous here

- I know you haven't looked at markup yet but just to note that we have a skipped heading level at the moment (h3) which will get picked up as an accessibility issue

- I'd like to run it past the Events team before we deploy, to increase their involvement

**Additional note:**
If we are changing headings etc., could we also change **button labels** to be sentence case, as per GDS guidelines?
(https://design-system.service.gov.uk/components/button/)? Or should I create a new PR for this as we have other instances eg mailing list signup (is it a common component in the form wizard?)?

peteryates:
> * Is 'Registering for' overkill? Having said that, I think it actually works well on mobile in the reduced summary box, so maybe leave as is
>
> * I'm not sure we need 'Your attendance will be added to a register and an email will be sent.'? Sounds a bit too formal and bureaucratic - I'll check with Gary as to the rationale behind this

I think it's worth getting @Tomcowan1 to cast his eye over these. I copied them word-for-word form the design but don't think either bit of text really provides the user with any useful information at this point.

> I agree that we need to review the headings next - the current h2 is superfluous here
>
> I know you haven't looked at markup yet but just to note that we have a skipped heading level at the moment (h3) which will get picked up as an accessibility issue

Agree with both of these, will address them in a separate PR as it'll affect the other pages of the wizard too.

peteryates:
I've reduced some of the duplication and split out the first page form so we can adjust it without affecting the others.

Now it looks like this:

Screenshot from 2021-09-02 15-23-55

I put the event name as the title for the time being, I don't really know what we should be using for it here. We're asking for personal details below so it feels like it should be something descriptive.

Tomcowan1:
Hi @peteryates and @gemmadallmandfe - just taking a look at this (apols for the delay), I would:

- Change the heading to an action. Either _Sign up for Train to teach question time Bristol_, or just _Sign up for this event_.
- I'd change '_Registering for_' to '_Event details_'.
- I think the grey box should just have the name, the date, the time and whether it's happening in person or online. Maybe we could just remove "Train to teach event" and the "Your attendance will be..." from below the first purple line and show either the laptop/online icon and "Online event" or the people icon and the address?

peteryates:
Took another crack at this. Followed Tom's suggestions but made the 'Event details' look more like the [GOV.UK caption pattern](https://design-system.service.gov.uk/styles/typography/#headings-with-captions). The 'Event details' is still inside the `<h4>` so a screen reader should read it out in one sentence.

Screenshot-2021-09-08-10:32:44

Not sure we can change the text next to the checkboxes without doing it across the other event pages/components too.

We could make this better if the title wasn't three bits of information concatenated.

gemmadallmandfe:
This looks better - thanks @Tomcowan1 and @peteryates!

Only comments are:

- In the latest version, the grey box changes from 'Event details' on desktop to 'Sign up for' on mobile - is this necessary, particularly if we have 'Sign up for...' in the h1 as well?

- I still think the heading levels are an issue @peteryates - we are skipping levels as we only have an h1 and an h4 on the page (and the h4 come first in the source?)

- Do we now need to review the headings for the other steps?

peteryates:
Agree that doubling up on "Sign up for" is superfluous, here's how it looks with the text above removed:

Screenshot-2021-09-08-12:00:25

On mobile, there is no heading above the `<h1>`, that's just strong text. Not really sure what's best when it comes to content above the main heading, never done it before.

I think the other pages of the wizard should be reviewed. If this was a regular GOV.UK service the page headings would be descriptive of the content and we'd be telling the user why we're asking for certain bits of information.

For example.

| GIT | School Experience |
| ---- |------|
| Screenshot-2021-09-08-12:08:11 | Screenshot-2021-09-08-12:07:20 |

Tomcowan1:
Agree it looks a lot better, and agree to get rid of "sign up for" on the mobile banner. I don't think it looks bad in the grey box (h level heading conversation not withstanding). 

Happy to take a look at the other headings. I think they're always the same "Sign up for this event: Train to teach London", but that should be in line with the Gov.uk style guides. Is that something to look at in this piece of work, or separately?

leoapost (approved):
Looks good to me

config/locales/en.yml
```diff
@@ -523,9 +523,8 @@ en:
 
     hint:
       events_steps_contact_details:
-        telephone: |-
-          This helps us contact you with updates to this event and other
-          important information.
+        address_telephone: |-
+           We'll only use this to let you know about any last-minute cancellations.
```
gemmadallmandfe, line 527:
```suggestion
           We'll use this to send you a reminder and let you know about any last-minute cancellations.
```

gemmadallmandfe:
Demoed to the Events team - one minor update needed to text as per suggestion

FROM: We'll only use this to let you know about any last-minute cancellations.

TO: We'll use this to send you a reminder and let you know about any last-minute cancellations.

Once this is done, we can put live.

app/controllers/mailing_list/steps_controller.rb
```diff
@@ -42,7 +42,7 @@ def crm_store
     def resolve_layout
       is_first_page = @current_step.instance_of? MailingList::Steps::Name
 
-      return "registration_attractive" if is_first_page
+      return "registration_with_side_images" if is_first_page
```
ethax-ross, line 45: I've removed `registration_attractive` as part of #1925 so I think this will conflict

ethax-ross (approved):
Looks good to me! 👍

gemmadallmandfe:
Looks good to me - happy for this to be merged!

app/controllers/mailing_list/steps_controller.rb
peteryates: I think it's ok. The 'attractive' one [is the one with photos at each side](https://user-images.githubusercontent.com/128088/139671307-46a15ed7-5986-4a39-a3de-f27c954ed3b4.png) - is #1925 intending to remove the images from the first page of the regular mailing list signup too?

PR #1667, opened 2021-08-25
Favour 'Next steps' over 'Next Steps'

peteryates:
The GOV.UK Design System strongly suggests that sentence rather than title case is used for headings, buttons and text in general:

> Write all headings in sentence case.

> Write button text in sentence case.


gemmadallmandfe (approved):
Looks good to me - thanks!

PR #1668, opened 2021-08-25
Add arrows to the welcome guide

ethax-ross:
### Trello card

[Trello-1741](https://trello.com/c/tkvCM43w/1741-welcome-guide-arrows)

### Context

The welcome guide has a bunch of colour-coded arrows in the design that direct the user from one section to the next. Adding in explicit SVGs for each arrow would be difficult to maintain and change in the future. Instead, we want to be able to dynamically generate arrows in Ruby and place/colour them in CSS.

- Add SVG/gemoetry gems

These gems will simplify drawing SVGs and calculating points along a bezier path.

- Add ArrowComponent

We need to be able to display arrows on the welcome guide between different bits of content.

Instead of having multiple SVGs for the different arrows pointing in various directions and with various amount of curving, this adds an `ArrowComponent` that - given points on a bezier path and an arrow head size - will render the arrow dynamically.

We can then color and position the arrow in CSS, which should make it easier to maintain and update in the future.

Includes previews for various arrow orientations.

- Remove unused welcome guide partial

This was added by mistake in an earlier PR; all the welcome guide CSS is in a single file.

- Add arrows to mobile view

Add arrows between sections to the mobile view of the page.

- Add arrows to desktop view

As the arrows are quite different in desktop we need to hide some of the mobile variants and add new desktop variants to get it to match the design.

### Guidance to review

There are a bunch of component previews to see how the arrows render that are available [here](/rails/view_components).

Drawing the arrow dynamically is pretty tricky - I've tried to explain the method as best I can in the comments of the view component. The only limitation with this approach is that the arrow head size is not consistent, this is because the bezier curve interpolation to get a point along the curve appears to be weighted by the control point (so a straight arrow will have a different a slightly different arrow head size to a very curvy path). To get around this for now I've made the arrow head size adjustable, which means you need to eyeball it for each arrow to get them consistent on a single page. Some of the designs also appear to use more of an arc path, which isn't replaceable as a bezier curve (I think we get close enough though). Diagram of how the drawing happens:

arrow drawning

The mobile design was fairly straight forward to implement; the desktop version isn't quite right yet though -- I haven't used CSS grids much in the past and the grid system in the welcome guide is beyond me at the minute, I'm pretty sure I've broken it a bit 🤯 

| Mobile      | Desktop |
| ----------- | ----------- |
| screencapture-0-0-0-0-3000-welcome-2021-08-25-14_51_23 | screencapture-0-0-0-0-3000-welcome-2021-08-25-14_50_30 |


app/webpacker/styles/welcome-guide.scss
```diff
@@ -213,10 +227,40 @@ $z-indices: ("bottom": 5, "low": 10, "medium": 15, "high": 20, "top": 25);
           object-fit: cover;
         }
       }
+
+      .arrow-container {
+        z-index: 14;
```
peteryates, line 232:
```suggestion
        z-index: map-get($z-indices, "low");
```

app/webpacker/styles/welcome-guide.scss
```diff
@@ -570,6 +651,22 @@ $z-indices: ("bottom": 5, "low": 10, "medium": 15, "high": 20, "top": 25);
...
+        }
+      }
+
+      .arrow-container.bottom {
+        text-align: right;
+
+        &.desktop {
+          grid-area: 6 / 4 / span 1 / span 1;
+          margin-bottom: -150px;
+          z-index: 1;
```
peteryates, line 666:
```suggestion
          z-index: map-get($z-indices, "bottom");
```

peteryates (commented):
Amazing! Love the solution to building the arrows. I think this is great, happy to approve once you're ready

peteryates (approved):
This is amazing! There's a few little tweaks we can do with the grid to bring the lines up to the screen edge, but they can be added later

PR #1637, opened 2021-08-12
Rewriting content to be easier to understand

Tomcowan1:
Part of the 'Basics' work - 

## not to be merged

Trello card

https://trello.com/c/f4W4RKHP/1650-the-basics-a-b-testing-a-new-version-of-the-salary-and-benefits-page

Context

Make content more understandable for people with little (or no) knowledge of the teaching world. 
Understand the financial life of a teacher.

PR #1718, opened 2021-09-02
Add a second variant of sals/benefits for testing

peteryates:
This is intended to be used for A/B testing so is removed from the navigation
and won't show up in search results or have SEO keywords. They should be added
back if this page 'wins' and supersedes the existing one.

Refs #1637
Fixes #1637


gemmadallmandfe (approved):
Looks good to me

PR #1720, opened 2021-09-02
Only show future pending events

jkempster34:
### Trello card
[Trello](https://trello.com/c/6N9robX0)

### Context
We have just had a discussion with the CRM team about whether we should be able delete events via the internal events form, deciding generally that we don't want to delete CRM entities via the API/Website.

Instead, to prevent the internal events index page getting cluttered with old withdrawn/incorrect events, we are going to only show future pending events. If users want events to be deleted properly they will still need contact the CRM team, but we may revisit this later if it becomes a bigger problem.

- Show only future pending events

### DIFF
--- a/app/controllers/internal/events_controller.rb
+++ b/app/controllers/internal/events_controller.rb
@@ -135,6 +135,7 @@ module Internal
       {
         type_id: event_type,
         status_ids: [pending_event_status_id],
+        start_after: DateTime.now.utc.beginning_of_day,
         quantity_per_type: 1_000,
       }
     end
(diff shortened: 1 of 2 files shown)


app/controllers/internal/events_controller.rb
```diff
@@ -135,6 +135,7 @@ def events_search_params(event_type)
       {
         type_id: event_type,
         status_ids: [pending_event_status_id],
+        start_after: DateTime.now.utc.beginning_of_day,
```
ethax-ross, line 138: Can we capture this in a spec? should just need to update this mock with a `.with({ type_id:....})`:

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/spec/requests/internal/events_controller_spec.rb#L41-L42

ethax-ross (approved):

PR #1728, opened 2021-09-07
Improve tab ordering in the cookie popup

peteryates:
### Trello card

https://trello.com/c/CCIJUUyS/1923-accessibility-fix-keyboard-focus-on-cookie-pop-up

### Context

The previous implementation had a couple of slight usability problems that this (hopefully) addresses.

One was that the 'cookies' link (to our cookies page) wasn't accessible via keyboard, tabbing only went between the acccpt/reject buttons.

The initial attempt to restrict tabbing to just one part of the page (the modal) failed, even when using the new Dialog API - as did adding `inert` in various places to prevent tabbing fron continuing to the obstructed background.

Additionally, any click to the modal would move focus onto the next element. This could have unintentional consequences for users who misclick.

### Changes

The approach I took was to just add a third target to the component and set up the tab order manually. It works for both <kbd>tab</kbd> and <kbd>shift+tab</kbd>

The event was changed from `blur` to `keydown` to stop clicking and other interactions moving the focus along unintentionally.

### Guidance to review

Testing tabbing also [appears to be painful](https://stackoverflow.com/questions/1601593/fire-tab-keypress-event-in-javascript) without bringing something that actually drives the browser along to the party.

### Preview

https://user-images.githubusercontent.com/128088/132350489-97e9f87d-cd90-4f55-8e64-28549f95ca97.mp4

### DIFF
--- a/app/components/footer/cookie_acceptance_component.html.erb
+++ b/app/components/footer/cookie_acceptance_component.html.erb
@@ -13,7 +13,7 @@
       <div class="dialog__text">
 
         <p>
-          We use <a href="/cookies" class="dark-link">cookies</a> to collect information about how you use this website.
+          We use <a id="cookies-info" href="/cookies" class="dark-link" data-cookie-acceptance-target="info">cookies</a> to collect information about how you use this website.
           We use this information to make the website work as well as possible, and improve this website.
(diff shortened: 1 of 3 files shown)


ethax-ross (approved):
Looks good and works well when I try it 👍

PR #1729, opened 2021-09-08
Put each structured data section behind feature flag

ethax-ross:
### Trello card

[Trello-1926](https://trello.com/c/MIJstS3b/1926-split-feature-flags-for-structured-data)

### Context

Some of the structured data is fine to be shipped just now, but other sections (such as events and blogs) need more
discussion out-with our team to get a green light.

Wrap each section behind a feature flag and enable the breadcrumb list in production; we're going to use this one
to ensure its working and determine how long it would take for Google to realise the change (so we get an idea of how
long a 'revert' would take if we make a mistake elsewhere, as there's no way to preview what the search results will
look like ahead of making them live).

- Put each structured data section behind feature flag

### DIFF
--- a/app/helpers/structured_data_helper.rb
+++ b/app/helpers/structured_data_helper.rb
@@ -6,7 +6,7 @@ module StructuredDataHelper
   EVENT_SCHEDULED = "https://schema.org/EventScheduled".freeze
 
   def structured_data(type, data)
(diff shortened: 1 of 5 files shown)


jkempster34 (approved):
👍 LGTM

PR #1741, opened 2021-09-10
Improving 'ways to train' badge style

peteryates:
### Trello card

https://trello.com/b/lRXllOHK/get-into-teaching-public-beta

### Context

The current badges aren't really 'on brand' or accessible. @gazcarr suggested some updated styles in the above Trello ticket and this is a quick look at a potential implementation.

| Current | New without bold text | New with bold text |
| ----- | ------ | ------ |
| Screenshot-2021-09-10-12:18:56 | Screenshot-2021-09-10-11:59:05 | Screenshot-2021-09-10-12:20:10 |

### DIFF
--- a/app/views/content/ways-to-train.md
+++ b/app/views/content/ways-to-train.md
@@ -91,10 +91,20 @@ Fees are usually around £9,250 but you can:
 <div class="training-attributes">
   <h4>University-led training</h4>
   <div class="badges">
-    <span class="badge pink">Costs £9,250</span>
-    <span class="badge dark-cyan">Funding available</span>
-    <span class="badge purple">Takes 1 year</span>
-    <span class="badge dark-blue">PGCE or PGDE</span>
+    <div class="badge pink">
+      <div class="badge-text">Costs £9,250</div>
+    </div>
+    <div class="badge dark-cyan">
+      <div class="badge-text">Funding available</div>
+    </div>
+
+    <div class="badge purple">
+      <div class="badge-text">Takes 1 year</div>
+    </div>
+
+    <div class="badge dark-blue">
+      <div class="badge-text">PGCE or PGDE</div>
+    </div>
   </div>
 </div>
 
--- a/app/views/content/ways-to-train.md
+++ b/app/views/content/ways-to-train.md
@@ -124,10 +124,26 @@ Fees are usually around £9,250 but you can:
 <div class="training-attributes">
   <h4>School direct<br/>(fee-funded)</h4>
   <div class="badges">
-    <span class="badge pink">Costs £9,250</span>
-    <span class="badge dark-cyan">Funding available</span>
-    <span class="badge purple">Takes 1 year</span>
-    <span class="badge dark-blue">Often a PGCE or PGDE</span>
+    <div class="badge pink">
+      <div class="badge-text">
+        Costs £9,250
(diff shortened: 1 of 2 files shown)


gazcarr:
Can we make the font size the same as the body copy (19px) and not bold. Also on the preview, I noticed some of them are italics.

Accessibility, I am not sure how much the dotted lines would impact readability, they seem a little close to the bottom of the text

gazcarr:
I'm getting a bit of corner cut off when expanded also

image

peteryates:
Made the fonts the regular size and weight and fixed the overhang problem.

The other badges look italic because of the skew, ignore them for the time being and I'll add the extra markup when we're fully happy with the top set.

peteryates:
I've applied the style changes to the rest of the badges on the page now. CSS is all in `app/webpacker/styles/ways-to-train.scss` for any final tweaks.

peteryates:
The solution to the pink/cyan badges now missing out on WCAG AA ratings is to:


* make the font slightly bigger (`@include font-size(small, $adjust: .32rem);`)
  Screenshot-2021-09-13-14:23:37

* make the text bold (`font-weight: bold;`)

  Screenshot-2021-09-13-14:25:27


* make the colours darker (not tried this yet 🎨)

gemmadallmandfe:
Happy to take your steer on this from a design perspective @gazcarr 

(I think the big font size is probably too big so I'd lean towards the bold, but happy for you to disagree!)

gazcarr:
@gemmadallmandfe I'd agree, bold is the better of the 2 options for this

gemmadallmandfe:
@peteryates can we see the bold solution in the review app? Happy to then approve.

gemmadallmandfe:
Bold tags look good - thanks @peteryates - happy with the design @gazcarr !

I've tested on desktop, in emulator and on my iPhone 11 (Chrome/Safari)

My only concern is width on smaller screen sizes

eg Emulator / iPhone 5
image

This page in Review also shifts horizontally a lot on my iPhone (whereas the one in prod doesn't, and the others in Review don't) - when I try and pinch it back to the original place, the width decreases permanently:

image

Is this to do with the new badges?

peteryates:
I think it's more-likely to do with this problem having been fixed in `master` already. I'll rebase and see if it fixes the problem @gemmadallmandfe

peteryates:
This now begs the question, should we also collapse the display (from `row` to `column`) on mobile?

| Row | Column |
| ---- | ----- |
| Screenshot-2021-09-15-14:21:31 | Screenshot-2021-09-15-14:21:54 |

@gemmadallmandfe @gazcarr

gemmadallmandfe:
Thanks @peteryates - width issues now resolved in emulator/iPhone 5 and on my iPhone 11/Safari and Chrome.

Good shout re vertical stacking which would resolve horrible wrapping on smaller screens - and would also be consistent with how we are displaying these badges in the Earn a salary section (not sure why these are different?):

image

peteryates:
They're different because they're marked up as a [definition list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) and we have some generic styles that wrap them on mobile. I'll make the others do the same too.

gemmadallmandfe (approved):
Looks good to me

PR #1749, opened 2021-09-14
Fix Train to Teach events links on Ways to Train

gemmadallmandfe:


peteryates (approved):
Looks good, new link works

PR #1751, opened 2021-09-15
2 wizard 2 gem

jkempster34:
### Trello card
[Trello](https://trello.com/c/dzHEOYnI)

### Context
The core Wizard code for multi-step forms is duplicated in the GiT app and TTA app. Move this code into a shared [Gem](https://github.com/DFE-Digital/dfe_wizard).

- Move Wizard code into a Gem
- Work with the changes to the Wizard introduced in the TTA app
  - The Wizard code has diverged in some small ways between the two apps. Update the GiT app to work with the changes that have been introduced in the TTA app


app/controllers/concerns/wizard_steps.rb
```diff
@@ -1,38 +1,13 @@
 module WizardSteps
   extend ActiveSupport::Concern
 
+  include DFEWizard::Controller
```
ethax-ross, line 4: I would have expected this whole concern to be in the gem... did you run into a reason why it can't be?

ethax-ross (commented):
Looks good and all seems to work well, just one question around why the concern couldn't all be extracted to the gem?

app/controllers/concerns/wizard_steps.rb
peteryates: How generic do we think the gem should be? Think it'll be of use to other teams?

ethax-ross: I was talking to Joe about this earlier, I think our wizard is so coupled to GiT/the API with matchback, magic links, pre-filling etc that trying to pull out a just a basic wizard would mean we don’t get very DRY between the two services (unless we added yet another gem into the mix). I think this wizard should be focused to GiT/API/CRM services, but open to other opinions

PR #1752, opened 2021-09-15
Extend nightly link checker to ensure internal links work too

peteryates:
### Trello card

https://trello.com/c/VNjS4sNU/1943-expand-broken-link-checker-to-cover-all-links-in-prod

### Context and changes

Currently we run a nightly check of our external links to ensure they resolve. Our internal link checker only runs as part of the CI process where we don't talk to the API, so links to API-requiring pages like events and the privacy policy are skipped.

This lead to a recent problem (#1749) where we had a broken link that wasn't picked up.

Here we're extending the link checker to do an internal run too. Internal links are identified by starting with a `/`. Instead of a single `LinkChecker` class we now have `LinkChecker::External` and `LinkChecker::Internal`. They work in mostly the same way but have slightly different `IGNORE` mechanisms and the internal checker also requires an extra step for 'translating' pack paths. As there's only one substitution required currently this is done by passing an array of args into `gsub` but this can be extended if needed.

### DIFF
--- a/spec/features/external_links_spec.rb
+++ b/spec/features/external_links_spec.rb
@@ -1,16 +1,40 @@
 require "rails_helper"
 
-RSpec.feature "external links check", :onschedule do
-  describe "all pages" do
-    it "will have no broken links" do
+RSpec.feature "link checks", :onschedule do
+  describe "all pages (external)" do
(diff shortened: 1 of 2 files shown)


ethax-ross (approved):
Looks great 👍

PR #1759, opened 2021-09-16
Swap order of page title so the name of the individual page comes first

jkempster34:
### Trello card
[Trello](https://trello.com/c/xjNidiD4)

### Context
The page title is shown in the browser's tab or as the window heading, and will be truncated if longer than the space available. 

Put the name of the individual page before `Get into Teaching` (instead of after) so that users can more easily see which page they are on when only the first part of the title is visible.

Before on the left, after on the right:
image

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -38,7 +38,7 @@ module ApplicationHelper
 
   def prefix_title(title)
     if title
-      "Get Into Teaching: #{title}"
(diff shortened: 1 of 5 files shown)


peteryates (approved):
This looks fine to me, neater with the page titles defined via `let` 👍🏽

ethax-ross (approved):
Looks much nicer 👍

gemmadallmandfe:
Looks good to me!

PR #1760, opened 2021-09-16
Temporarily unpublish the eligibility checker

peteryates:
This page shouldn't really be 'live' yet and we're no longer actively testing it, so unpublish for the time being.


gemmadallmandfe (approved):
Thanks @peteryates - I've tested that eligibility checker is not available in Review app so happy for this to be merged

PR #1761, opened 2021-09-17
Reduce Content Layout Shift in header on desktop

ethax-ross:
### Trello card

[Trello-1938](https://trello.com/c/3TWYT44i/1938-seo-fix-cumulative-layout-shift-cls-issue-core-web-vitals)

### Context

We are seeing a small amount of Content Layout Shift caused by the header section and Google is flagging this in Search Console. We want to remove the CLS if possible to improve our scores/rankings.

- Set navigation height to avoid layout shift

When the autocomplete loads it injects an input into the navigation bar, which changes its height. Instead, we can fix the height so that it doesn't shift on load.

- Hide covid banner by default

The covid banner is in a dismissed state for the Google Page Speed checker and because we show it by default and hide it with Javascript we get a content layout shift reported. To avoid this we can hide it by default and show it only when it has not been dismissed.

Minor refactoring to the `BannerController` to use the new `*-value` syntax.

If Javascript is disabled we show the banner (which maintains the current behaviour).

### Guidance to review

There is still a very small amount of CLS (~0.001) from the autocomplete loading in via Javascript; I've tried to minimise this by setting the bounds of the auto-complete but it doesn't appear to eradicate it completely.

### DIFF
--- a/app/components/header_component.html.erb
+++ b/app/components/header_component.html.erb
@@ -5,7 +5,7 @@
 </div>
 
 <div aria-label="main">
-  <section class="covid" data-controller="banner" data-banner-name="Covid">
+  <section class="covid" data-controller="banner" data-banner-name-value="Covid">
     <div class="limit-content-width">
       <span class="covid__header"><a href="/covid-19">Find out how coronavirus is affecting teacher training</a></span>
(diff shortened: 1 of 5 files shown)


jkempster34 (approved):
LGTM 💯

PR #1763, opened 2021-09-17
Fix map image size on desktop

ethax-ross:
### Trello card

[Trello-1934](https://trello.com/c/TFY33143/1934-seo-fix-largest-contentful-paint-lcp-issue-core-web-vitals)

### Context

Page insights is flagging the unknown width/height of the image as an issue; we can resolve this on desktop as the width of the content is a static value, but we're stuck with a responsive image on tablet/mobile.

- Fix map image size on desktop

### DIFF
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -80,7 +80,7 @@ module EventsHelper
     ajax_map(
       address,
       zoom: 10,
-      mapsize: [628, 420],
+      mapsize: [732, 490],
       title: event.name,
       description: address,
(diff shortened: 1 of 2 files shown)


jkempster34 (approved):
Looks good to me! 🗺️ 

Is it worth storing the CSS values in a variable or adding a comment to say that they match what is being sent to Google Maps API?

ethax-ross:
Ah sorry @jkempster34 I had auto-merge on 😬 I'll raise a PR with a comment in a bit

PR #1767, opened 2021-09-17
Add Zendesk settings

ethax-ross:
Route101 have requested these settings to fix an issue they have encountered.


peteryates (approved):
Looks fine to me

PR #1768, opened 2021-09-17
Remove 'Read more' and make heading links clearer

peteryates:
### Trello card

https://trello.com/c/rAAcA3xI

### Context and changes

There were a couple of usability fails here, firstly that the 'Read more' link was in isolation so it wouldn't be clear to screenreader users _what_ they'd be reading.

Secondly, because the headings were links too we were repeating links to the same page in quick succession.

| Mobile | Desktop |
| ------ | -------|
| Screenshot-2021-09-17-13:45:54 | Screenshot-2021-09-17-13:46:02 |

### DIFF
--- a/app/views/blog/_post_preview.html.erb
+++ b/app/views/blog/_post_preview.html.erb
@@ -11,6 +11,4 @@
     <p><%= first_image_from_post(fm[:images]) %></p>
   <% end %>
   <p><%= fm[:description] %></p>
-
-  <%= link_to "Read more", path %>
 </article>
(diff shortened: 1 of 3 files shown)


gemmadallmandfe:
This looks good to me - but I haven't approved yet in case you want to investigate further/change given the comments on the cross-gov Slack we saw yesterday @peteryates.

I'll approve if you are still happy with this approach

gemmadallmandfe (approved):
Happy to approve this

PR #1769, opened 2021-09-20
Prevent cookie banner from being the LCP element

ethax-ross:
### Trello card

[Trello-1934](https://trello.com/c/TFY33143/1934-seo-fix-largest-contentful-paint-lcp-issue-core-web-vitals)

### Context

Currently, if a user has not accepted our cookie policy the banner text gets classified as the Largest Contentful Paint
element on pages without a hero image or a lot of text above the fold (i.e. most event pages).

The only way I've found to get around this is to load the banner visible but off-screen and then transition the banner to the
main viewport with a CSS animation (of 0s duration) by adding the class in JS if the user hasn't accepted cookies yet.

- Prevent cookie banner from being the LCP element

### DIFF
--- a/app/webpacker/controllers/cookie-acceptance_controller.js
+++ b/app/webpacker/controllers/cookie-acceptance_controller.js
@@ -19,14 +19,14 @@ export default class extends Controller {
 
   accept(event) {
     event.preventDefault();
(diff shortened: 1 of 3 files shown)


peteryates (approved):
Approach seems sensible and works well for me

PR #1771, opened 2021-09-20
Fix position search box on mobile

jkempster34:
Apply height to the search box on desktop only. The `extra-navigation` component takes up a lot more space on mobile and was spilling over onto the hero.

image


ethax-ross (approved):
This is definitely less-broken, but I think it may have been 100% width on mobile previously; I'll check out an old version of master and see but worth merging this down in the interim 👍

PR #1773, opened 2021-09-20
Remove CSP whitelist of Google domains

jkempster34:
### Trello card
[Trello](https://trello.com/c/jwN0DVrb)

### Context
A number of Google domains (~200) have been whitelisted for Google Analytics across 4 directives (connect-src, script-src, img-src, style-src). 

However, it is recommended that headers are kept as small as possible.

Remove the list, separately adding a Prometheus alert for any CSP violations that include `Google` so that the list can be built back up.


config/initializers/secure_headers.rb
```diff
@@ -25,9 +25,7 @@
     https://www.googletagmanager.com/gtm.js
   ]
 
-  # curl https://www.google.com/supported_domains > config/csp/google_supported_domains.csv
```
peteryates, line 28: This is fine but I have no idea why of all of Google's domains `.com.sa` is special 🤷🏽‍♂️

peteryates (approved):
Looks fine, puzzled by `.sa`

config/initializers/secure_headers.rb
jkempster34: That was there before the supported domains list was pulled in so I figured it was one that was commonly popping up 🤔 

Just removed it!

ethax-ross (approved):
Looks good, but we'll need to keep a close eye on the CSP reports when this goes out 🕵️

config/initializers/secure_headers.rb
peteryates: Yeah - I remember it being there before, I just never understood why it was. I think reverting back to what you've left in is sensible and we can just keep an eye on it and add some back if needs be 👍🏽

PR #1774, opened 2021-09-20
Updating content following feedback about accuracy

Tomcowan1:
### Trello card

https://trello.com/c/0REY4GDx/1944-review-wording-on-ways-to-train-re-time-spent-on-theory-of-teaching

### Context

Have edited wording. Hopefully speaking to Policy to agree it.

### DIFF
--- a/app/views/content/ways-to-train.md
+++ b/app/views/content/ways-to-train.md
@@ -112,7 +112,7 @@ Fees are usually around £9,250 but you can:
 
 On a 'school direct (fee-funded)' programme, your learning will be centred on a particular school, but you will also be placed in different schools during your course.
 
-Many school direct programmes give you a PGCE or PGDE (an academic qualification) in addition to QTS.
+Many 'school direct (fee-funded)' programmes give you a PGCE or PGDE (an academic qualification) in addition to QTS.
 
 Courses typically take one year but you may be able to study part-time.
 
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
Changes look good to me @Tomcowan1 - have they been approved by policy?

The only wording I would maybe question is 'These courses are **usually suitable**....' under Earn a salary. I'm presuming it is worded like this as we need to keep it open - ie it is not _mandatory_ to have teaching experience to do this type of training, but it is usual?

Tomcowan1:
> Changes look good to me @Tomcowan1 - have they been approved by policy?
> 
> The only wording I would maybe question is 'These courses are **usually suitable**....' under Earn a salary. I'm presuming it is worded like this as we need to keep it open - ie it is not _mandatory_ to have teaching experience to do this type of training, but it is usual?

@gemmadallmandfe Yes that's right. It's not mandatory, and UCAS currently describe it as "transferable experience", but a TTA has told me before that this way of training is really only open to people who are doing the course as a way to qualify in a job they're already pretty much doing.

gemmadallmandfe:
Thanks @Tomcowan1 - looks good to me but will not approve until you've met with policy :-)

Tomcowan1:
HI @gemmadallmandfe have talked Rick Rahman from the ITT compliance team (who worked with Emma) through the changes and he is happy. Think we're good to merge.

gemmadallmandfe (approved):
Looks good to me

PR #1778, opened 2021-09-21
Add Google Tag Manager nonce

jkempster34:
We are getting alerts that our CSP policy is blocking https://www.googletagmanager.com/gtm.js. Add nonce as suggested in https://developers.google.com/tag-manager/web/csp


jkempster34:
Already in place

PR #1772, opened 2021-09-20
Updating promotion space with the mailing list

Tomcowan1:
### Trello card

https://trello.com/c/00caDBym/2450-replace-events-promo-on-home-page-to-highlight-mailing-list-tie-in-with-new-tv-campaign

PR #1779, opened 2021-09-21
Fix right column cta indentation on funding page

peteryates:
Small fix, just make the CTA show on the funding page. It's currently skipped due to it not being nested inside `right_column`. Refs #1772


gemmadallmandfe (approved):
Looks good to me - happy for this to be merged

Will Tom's PR need rebasing (1772)?

PR #1781, opened 2021-09-21
Improve back link text on cookie prefs page

ethax-ross:
### Trello card

[Trello-1868](https://trello.com/c/SNAhkS5L/1868-back-to-page-button-is-misleading-on-cookie-preferences-page)

### Context

When someone comes from an external referrer/unknown source or directly onto the cookie preferences page we show 'Back to page' even though we take them to the homepage and not where they came from.

Instead, we can change this text to be 'Go to home page' for better clarity and only show 'Back to page' if we actually know
the page they came from. 

- Improve back link text on cookie prefs page

### Guidance to review

| When we know the page they came from      | When we don't know the page they came from |
| ----------- | ----------- |
| Screenshot 2021-09-21 at 13 46 34      | Screenshot 2021-09-21 at 13 46 44      |

### DIFF
--- a/app/views/cookie_preferences/show.html.erb
+++ b/app/views/cookie_preferences/show.html.erb
@@ -183,8 +183,7 @@
       </p>
 
       <p>
-        <%= link_to "Find out more", cookies_path %>
-        about cookies on this service
+        <%= link_to "Find out more about cookies on this service", cookies_path %>
       </p>
     </div>
   </form>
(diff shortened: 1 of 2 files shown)


peteryates (dismissed):
👍🏽

gemmadallmandfe:
As an additional improvement, could you possibly change the link from "Find out more" to "Find out more about cookies on this service" (ie the whole sentence) so that we provide more context from an accessibility point of view?

Otherwise looks good to me - happy for this to be merged @ethax-ross 

Thanks!

peteryates (approved):
Looks good

PR #1782, opened 2021-09-21
Remove closing </source> tag from within <picture> elements

ethax-ross:
### Trello card

### Context

When we perform image processing we manually modify the HTML response with Nokogiri to inject `<source>` tags. By default Nokogiri will render closing tags for empty elements, resulting in `<source></source>` which is not valid HTML 5 (tags with no content should be self-closing).

It is possible to configure Nokogiri to self-close empty tags, however in doing so it completely trashes other parts of the HTML document (we probably have invalid HTML/an unclosed tag somewhere that's causing this). For now we can manually remove the closing `</source>` tags.

- Remove closing `</source>` tags

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -45,6 +45,11 @@ private
     response.body = NextGenImages.new(response.body).html
     response.body = ResponsiveImages.new(response.body).html
     response.body = LazyLoadImages.new(response.body).html
+
+    # <source> has no content so should be self-closing; configuring Nokogiri
+    # to remove the closing tags results in breakages elsewhere in the document,
+    # so we have to remove them manually post image processing.
+    response.body = response.body.gsub("</source>", "")
   end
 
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -45,6 +45,11 @@ def process_images
     response.body = NextGenImages.new(response.body).html
     response.body = ResponsiveImages.new(response.body).html
     response.body = LazyLoadImages.new(response.body).html
+
+    # <source> has no content so should be self-closing; configuring Nokogiri
+    # to remove the closing tags results in breakages elsewhere in the document,
+    # so we have to remove them manually post image processing.
+    response.body = response.body.gsub("</source>", "")
```
jkempster34, line 52: Looked up `gsub`, could you use `response.body.delete("</source>")` instead?

jkempster34 (approved):
👍

PR #1783, opened 2021-09-21
Revert "Attempt to stop Google CSP violations"

jkempster34:
Reverts DFE-Digital/get-into-teaching-app#1780

The CSP reports have continued, so this didn't work.


ethax-ross (approved):
I would maybe give it until the morning as I've found it takes a while for the CSP updates to propagate, unless you're seeing the updates in the CSP report already

PR #1786, opened 2021-09-22
Fix possible error on search param

ethax-ross:
If a user manually enters `/search?search` then the `params.fetch` returns `nil` and if they enter `/search?search=` then `params.fetch` will be an empty string. We need to ensure that we don't call `permit` on either of these values as it will result in an exception. Instead, we check the type is `ActionController::Parameters` and, if not, we return `nil` to give an empty
result.

### DIFF
--- a/app/controllers/searches_controller.rb
+++ b/app/controllers/searches_controller.rb
@@ -17,7 +17,13 @@ private
   end
 
   def search_params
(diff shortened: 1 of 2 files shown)


jkempster34 (approved):
💯 LGTM

PR #1789, opened 2021-09-22
Various HTML validation fixes

ethax-ross:
### Trello card

### Context

Ahead of turning on HTML linting I'm working through the majority of the validation issues; these are a bunch of quick fixes - there are more to follow that are a little harder to unpick.

- Fix duplicate meta/charset

It looks like our current `<meta>` tag specifying the `charset` is not recognised by Nokogiri so our image processing adds a second, which causes invalid HTML (we have both). Updating our `<meta>` to be more complete seems to prevent Nokogiri from adding a second one.

- Fix <nav> role attribute value

Our `<nav>` tag has a `role` attribute of `nav` which is not valid; it should be `navigation`, but that is also no longer necessary (and provokes a warning) so it can just be omitted.

- Remove misuse of aria-label on Covid banner

Flagged in the HTML validator; the content of the banner is screen-reader friendly so this isn't necessary.

- Remove unnecessary role attribute from main element

The HTML validator flags this as a warning; its no longer necessary to specify it.

- Prevent empty src/srcset attributes

When we lazy load images we move the `src` to `data-src`; this means the `src` is initially empty, which is not valid HTML. To get around this we can load a tiny base-64 GIF that will then be replaced when the lazy image is loaded.

- Move <noscript> outside of <picture> tag

We currently have a `<noscript>` inside a `<picture>` element, which is not valid HTML 5. Instead, we can duplicate the `<picture>` tag and add it to a `<noscript>` sibling to get the same result.

- Set src of video iframe

An iframe without a `src` attribute is not valid; instead, we can set it to be a tiny GIF. Switch other iframes to using the native `loading` attribute to support lazy loading. Also incides some minor tweaks to the markdown to make the video player more semantically correct.

### DIFF
--- a/app/components/footer/video_player_component.html.erb
+++ b/app/components/footer/video_player_component.html.erb
@@ -3,19 +3,18 @@
   </div>
   <div class='video-overlay__video-container'>
     <div class="video-overlay__video-container__wrapper">
+      <!-- 'about:blank' as a placeholder src to avoid HTML 5 validation error -->
       <iframe data-video-target="iframe"
               title="Get Into Teaching Video Player"
               width="800"
               height="450"
-              src=""
-              alt="Watch a video for more information about becoming a teacher"
+              src="about:blank"
               allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"
-              allowfullscreen
-              playsinline="false">
+              allowfullscreen>
       </iframe>
     </div>
     <button id="close-button" class="video-overlay__video-container__dismiss" data-video-target="close" data-action="click->video#close">
-      <div class="icon-video-close"><span class="visually-hidden">Close video</span></div>
+      <span class="icon-video-close"><span class="visually-hidden">Close video</span></span>
     </button>
     <a class="visually-hidden" href="#">
       Close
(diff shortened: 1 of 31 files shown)


jkempster34 (approved):
LGTM 🔥 . Just wondering why a tiny gif? Found that says use about:blank https://javascript.tutorialink.com/is-an-empty-iframe-src-valid/

ethax-ross:
Yeah I think you're right `about:blank` is a better fit for the iFrame src; I'll update it and check it passes the HTML validation still.

app/components/footer/video_player_component.html.erb
```diff
@@ -3,19 +3,18 @@
   </div>
   <div class='video-overlay__video-container'>
     <div class="video-overlay__video-container__wrapper">
+      <!-- Tiny GIF as a placeholder src to avoid HTML 5 validation error -->
```
jkempster34, line 6: Comment still here 👀

jkempster34 (changes requested):

jkempster34 (approved):

app/views/content/teacher-training-advisers.md
```diff
@@ -89,8 +89,8 @@ Your adviser can help you navigate your path into teaching, and make sure you ge
 
 <div data-controller="aspect-ratio" data-aspect-ratio-width-value="560" data-aspect-ratio-height-value="290">
   <iframe 
-    class="lazyload"
-    data-src="https://www.youtube-nocookie.com/embed/T9Bhcaa6LJ4" 
+    loading="lazy"
```
peteryates, line 92: 🤯

gemmadallmandfe (approved):
Happy to approve as requested!

PR #1790, opened 2021-09-22
Prevent users double submitting events search

jkempster34:
## Context
We have seen an [events search error](https://sentry.io/share/issue/acb9ab87765e443e8933b502f57ca499/) a few times.

When a user searches for a postcode that isn't in the cache, the API will look it up via Google Geocoding then persist it.

If the API gets multiple requests in quick succession, the requests will be queued, resulting in a [duplicate key exception](https://sentry.io/share/issue/f87f1c5cdec04085a009219955e5b30a/) as the API tries to persist the same postcode multiple times (the key for the `Locations` table is the postcode).

We can prevent users submitting the form multiple times with `data-disable-with`. The message `Searching...` will be shown in the button while the request is processing.

image

### Guidance to review
To replicate the error:
- Visit `events/search`
- Choose `Within X miles` for `Location` with a non-cached postcode (easier if you choose a partial postcode. I have done quite a few in the M1-M33 and ST1-ST9 range already).
- Click a few times fast on the search button.


peteryates (approved):
Nice

peteryates:
Approved - perhaps worth making sure the `data-disable-with` attribute is presence in a spec though?

PR #1794, opened 2021-09-23
Load Zendesk JS when user clicks chat button

ethax-ross:
### Trello card

[Trello-2479](https://trello.com/c/fUi5y1l3/2479-optimise-loading-of-zendesk-widget?filter=label:Development&menu=filter)

### Context

Currently we load the Zendesk code on every page; even if we defer the loading to the end it still has a significant impact on our page speed.

Instead, this commit doesn't load the Javascript until the user interacts with the chat online button; at which point it will download the JS and then open the chat window.

We could change the button state to be 'loading' while it downloads, but in my testing it's still reasonably responsive without it.

As user can get to the various privacy pages without accepting any cookies we also need to ensure they can't use the live chat from this page (it sets a functional cookie).

Lastly, the Zendesk chat sets a cookie to store the customer ID for authentication purposes and we should reflect this in our cookies page. It can also store a cookie relating to their cookie acceptance, however as we prompt for this outwith Zendesk we do not have it enabled in the chat widget separately.

- Load Zendesk JS when user clicks chat button
- Update cookie policy with Zendesk chat cookie
- Hide live chat on cookie preferences page

### Guidance to review

This brings the page speed score on mobile back up to around 98/100 from 55/100.

| Updated cookie policy table      | Hidden chat on privacy pages |
| ----------- | ----------- |
| Screenshot 2021-09-23 at 11 59 26 |  Screenshot 2021-09-23 at 12 00 50       |


app/views/pages/cookies.md
```diff
@@ -35,6 +35,7 @@ Some examples of these cookies include:
 | `git-cookie-preferences-v1`  | Record acceptance of beta site cookie policy       | 90 days                   |
 | `_vwo_uuid_v2`               | Redirection from original site to new Beta site    | 1 year                    |
 | `GiTBetaFeedbackPrompt`      | Record response to our request for feedback        | End of browser session    |
+| `__zlcmid`                   | Store customer ID for chat widget authentication   | 365 days                  |
```
gemmadallmandfe, line 38:
```suggestion
| `__zlcmid`                   | Store unique ID for chat widget authentication   | 365 days                  |
```

gemmadallmandfe:
Thanks @ethax-ross - this looks good to me.

The **delay in loading the chat widget** does not appear significant, so the change in text to 'Starting chat' or similar may not be necessary after all, as you say. Is there any risk that users with worse connection speeds will face longer delays?

**Hiding the chat on the cookies preference page** works well.

I've made a minor suggestion to the **wording on the cookies page** (replacing 'customer') - we need to do a proper review of this page anyway.

I can see that the **zlcmid cookie only drops if you interact with the Chat online button** - this works well (can't see any other cookies except the dfe_session and git-cookie-preference, but I presume that is just because they aren't on Review!).

Happy for this to be deployed but I think that we should just notify Route 101/TPUK as to when this will happen, just in case there are any issues/they get any feedback.

gemmadallmandfe:
I've tested the loading message - looks good to me.
image

Thanks for changing the cookies page too.

Happy to approve this when needed - I will contact Route 101/TPUK prior to merging.

spec/javascript/controllers/talk-to-us_controller_spec.js
```diff
@@ -61,22 +64,40 @@ describe('TalkToUsController', () => {
...
+    });
+
+    describe('when clicking the chat button twice', () => {
+      it('only appends the Zendesk snippet once and does not show the loading message for the second click', () => {
+        const button = document.querySelector('a');
+        button.click();
+        expect(button.textContent).toEqual("Starting chat...");
+        jest.runOnlyPendingTimers(); // Timer for script loading,
+        jest.runOnlyPendingTimers(); // Timer to wait for chat window to open.
+        expect(chatShowSpy).toBeCalledWith();
```
jkempster34, line 87: Should this `toBeCalledWith` have a value?

app/webpacker/controllers/talk-to-us_controller.js
```diff
@@ -46,7 +46,46 @@ export default class extends Controller {
...
+    }
+
+    const script = document.createElement('script');
+    script.setAttribute('id', 'ze-snippet');
+    script.src =
+      'https://static.zdassets.com/ekr/snippet.js?key=34a8599c-cfec-4014-99bd-404a91839e37';
+    document.body.appendChild(script);
+  }
+
+  waitForZendesk(callback) {
```
jkempster34, line 82: There is async/await in JS now if we want to avoid nesting callbacks. Might be neater?
```
 async waitForZendesk() {
    const interval = setInterval(() => {
      if (window.$zopim && window.$zopim.livechat) {
        clearInterval(interval);
      }
    }, 100);
  }

async showZendeskChat() {
  await this.waitForZendesk()
  ...
}
```

jkempster34 (dismissed):
Left some small comments, not sure if they are important. Other than that, LGTM 💯

app/webpacker/controllers/talk-to-us_controller.js
jkempster34: Actually, it seems `setInterval` doesn't work nicely with Promises/async stuff because it can trigger multiple times so nevermind 🤷

spec/javascript/controllers/talk-to-us_controller_spec.js
ethax-ross: It works without, but yeah I'll change it to `toHaveBeenCalled` as that reads much nicer 👍

jkempster34 (approved):

jkempster34 (approved):

PR #1796, opened 2021-09-24
Update assessment only contact details for GUTP

gemmadallmandfe:
Request from The Grand Union Trading Partnership to update their contact details (name and email) on Assessment Only page and add their details to the South East section as well as East Midlands.


peteryates (approved):
Looks good

PR #1798, opened 2021-09-24
Enable Markdown linting

ethax-ross:
### Trello card

### Context

We want to lint our Markdown files to keep them consistent and avoid potential errors.

- Add Markdown linter and configure

Adds a Markdown linter and configures it so that it by-en-large passes on our set of Markdown files.

The linter can be ran with:

```
mdl app/views/content/**/*.md
```

It is set to run during CI and fail if there are any linting errors present. This may be too heavy-handed and blocking for content editors so we may end up switching it off and running it nightly instead.

- Various Markdown linting fixes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -194,7 +194,14 @@ jobs:
 
       - name: Lint Markdown
         run: |-
-          docker run -t --rm ${{needs.build.outputs.DOCKER_IMAGE}} "bundle exec mdl app/views/**/*.md"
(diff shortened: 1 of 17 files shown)


sfawcett123 (changes requested):
Its great we have a system to check these, but if there is an output we do not see it, except in the logs. can I suggest that you add a step like we do in rubocop where we keep the output on failure as an artefact

sfawcett123 (approved):

gemmadallmandfe (approved):
Happy to approve following Steve's review

PR #1799, opened 2021-09-27
Fix broken train to teach events link

ethax-ross:
The events link in the funding your training CTA was incorrect.


gemmadallmandfe (approved):
Looks good to me

PR #1801, opened 2021-09-27
Add a new mailing list completed page

peteryates:
### Trello card

https://trello.com/c/BbakOzej/1921-welcome-guide-design-and-develop-mailing-list-completion-page

### Context

This change replaces the old mailing list completion page with a new personalised one. To do this, rather than purging the mailing list data from the session at the end of the wizard we prune it, leaving in the few bits we'll need for personalised pages we want to display (here, the completion page and welcome guide).

For now, because the welcome guide's not ready, I've left the purple boxes in when the environment is `production`. This will allow the new completion page to be deployed before the welcome guide is finalised and not leave users with a dead end.

#### Potential data risks

The `#prune!` method is a little like purge except it allows some keys (none by default) to be left in the session's `app_data` store.

For now, if a user restarts the mailing list signup, the first page will have the `first_name` and `last_name` they used on the previous run pre-filled.

Data is all stored in the session so assumed to be safe, although using a shared machine could allow people to see previous users' data. This would be the same had a user partially filled in the wizard then stopped though, so we're not really introducing a new attack vector here, but it's something to be aware of.

### Changes

- Replace mailing list completion page
- Add `#prune!` to Wizard::Store

### Preview

|  Desktop | Mobile |
| ----- | ----- |
| Screenshot-2021-09-27-14:12:11 | Screenshot-2021-09-27-14:11:58 |

### Guidance to review

* is leaving things in the session a new security/PII risk?
* are the purple boxes on the completion page suitable for users who finish the wizard? Should we add a button that takes them to the homepage too?

### DIFF
--- a/app/models/mailing_list/wizard.rb
+++ b/app/models/mailing_list/wizard.rb
@@ -20,7 +20,11 @@ module MailingList
         break unless result
 
         add_member_to_mailing_list
-        @store.purge!
+
+        # we're taking the last name too so if people restart the wizard
+        # both are filled rather than just their first name, which looks
+        # a bit odd
+        @store.prune!(leave: %w[first_name last_name])
       end
     end
 
--- a/app/components/registration/mailing_list_signup_confirmation_component.html.erb
+++ b/app/components/registration/mailing_list_signup_confirmation_component.html.erb
@@ -1,0 +1,14 @@
+<section class="text-content mailing-list__completed">
+  <%= tag.h2(heading_text, class: 'green') %>
+
+  <p>
+    We'll send your first email shortly. From teacher stories to practical
+    information, we've got all the info you need to become a teacher.
+  </p>
+
+  <p>
+    You can unsubscribe at any time.
+  </p>
+
+  <%= link_to "Continue your journey", "/welcome", class: "button" %>
+</section>
--- a/app/components/registration/mailing_list_signup_confirmation_component.rb
+++ b/app/components/registration/mailing_list_signup_confirmation_component.rb
@@ -1,0 +1,15 @@
+class Registration::MailingListSignupConfirmationComponent < ViewComponent::Base
+  attr_reader :mailing_list_session
+
+  def initialize(mailing_list_session)
+    @first_name = mailing_list_session["first_name"]
+  end
+
+  def heading_text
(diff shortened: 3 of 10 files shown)


app/models/mailing_list/wizard.rb
```diff
@@ -20,7 +20,7 @@ def complete!
         break unless result
 
         add_member_to_mailing_list
-        @store.purge!
+        @store.prune!(leave: %w[first_name last_name])
```
ethax-ross, line 23: Looks like `last_name` isn't referenced in the view?

ethax-ross (approved):
Looks good, one minor comment. Is the intention to enable the welcome guide in production here as well, or am I misreading that change?

gemmadallmandfe:
@peteryates This personalised version of the mailing list completion page with a link to the welcome experience should only be served to certain segments of users (combination of subject/stage/commitment - we are awaiting the final mapping from Havas).

We've deliberately stripped out the other CTAs from the page to encourage them to clickthrough to the welcome experience content.

However, other users should still get the original mailing list completion page with additional signposting.

Sorry if this wasn't clear (it's in the emails but not on the Trello card) - very happy to discuss.

Re Ross's comment - I agree that we only need first name for the personalisation. It would be good to understand more about purge vs prune though!

peteryates:
@gemmadallmandfe ah I'd never have gleaned that from what's in the Trello ticket 😩 

Will sort it tomorrow.

app/models/mailing_list/wizard.rb
peteryates: Yeah - I added that because it looked weird if you restart the wizard and your first name is pre-filled and your last name isn't!

ethax-ross: Ah that makes sense; worth adding a quick comment to that effect so someone doesn't accidentally refactor it out later?

peteryates:
Update: now the new green button layout is only visible in non-test/non-prod environments. 

It'll look as before in prod and test (and there are now a tests to confirm it)

cc @ethax-ross, @gemmadallmandfe

ethax-ross (approved):
Looks good 👍

PR #1803, opened 2021-09-28
Replace salaries and benefits with simplified one (UR)

peteryates:
This is to test the page side-by-side in the UR environment against the regular page in production.


peteryates:
I believe this UR is done so closing this. Can resurrect if needed.

PR #1809, opened 2021-09-29
Remove and validate extra characters when creating event partial URL

jkempster34:
### Trello card
[Trello](https://trello.com/c/G8l32WiI)

### Context
Users can generate a partial URL based on the event name and start date of an event. Currently there is only minimal validation, so we have seen event partial URLs with non-alphanumeric characters such as brackets, colons, and question marks.

- Update the JavaScript that generates the partial URL to cut out any characters that are not alphanumeric or whitespace. 
- Update the partial URL validation so that only alphanumeric, hyphens, and underscores are allowed.

### Guidance to review
Generating the partial URL:
image
image

Can the JS regex be improved? is there a way to do `[^\w\s]` without underscores? [here](https://regexr.com/66hm2)

### DIFF
--- a/app/models/internal/event.rb
+++ b/app/models/internal/event.rb
@@ -29,7 +29,7 @@ module Internal
     attribute :venue_type, default: VENUE_TYPES[:none]
 
     validates :name, presence: true, allow_blank: false, length: { maximum: 300 }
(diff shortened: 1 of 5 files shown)


peteryates (approved):
Looks good to me

PR #1810, opened 2021-09-30
Display email fallback in place of chat when JS is disabled

ethax-ross:
### Trello card

[Trello-1879](https://trello.com/c/HvPB0Jrh/1879-test-adding-age-range-to-the-mailing-sign-up-process)

### Context

When JS is disabled we need to show a fallback email address in the talk to us section and change the title to 'Email us' instead of 'Chat to us'. When out of support hours we want to hide the 'Email us' section entirely. This differs to how we handle it for JS-enabled users (we retain the heading and replace the button with 'Chat closed' text).

- Add feature specs for chat online buttons

Add feature spec coverage for all the chat online buttons/variations. This was challenging because we need to be able to mock the current browser time from Rails for Capybara, which is in a different process. The method I used involves setting the time you want to use in the query string and then loading in a new pack that mocks out the current Javascript `Date` object. I used `sinon` for this instead of `jest` because you can't include `jest` outside of the test environment. The downside is we now have a test dependency in our non-dev JS dependencies, but its only loaded in on the test environment so it won't have a performance hit.

- Support fallback email for when JS is disabled

### Guidance to review

This is a temporary solution ahead of a wider redesign of the chat component/CTAs; the code is complex/hard to follow and not particularly elegant due to all the different ways we show a chat CTA and the variations for if chat is online/offline and JS is enabled/disabled. If the plan was to keep it as-is longer term I would have refactored the `chat_link` helper into something more maintainable... that being said, it should be robust enough as-is for now given the feature specs that surround the different uses of the chat online button/component.

Screenshot 2021-10-21 at 07 52 30

### DIFF
--- a/app/components/footer/talk_to_us_component.html.erb
+++ b/app/components/footer/talk_to_us_component.html.erb
@@ -11,7 +11,7 @@
               <div class="live-chat">
                 <p class="js-enabled-title">Live chat</p>
                 <p class="js-disabled-title">Email us</p>
-                <%= helpers.chat_link("Chat online", classes: "button", fallback_text: nil, fallback_email: "getintoteaching@education.gov.uk", offline_text: "Chat closed") %>
+                <%= helpers.chat_link("Chat online", classes: "button", fallback_text: nil, fallback_email: "getintoteaching.getintoteaching@education.gov.uk", offline_text: "Chat closed") %>
               </div>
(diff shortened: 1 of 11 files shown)


peteryates (approved):
Looks good.

peteryates (approved):
🚢

PR #1813, opened 2021-09-30
Remove automated event closed message

ethax-ross:
### Trello card

[Trello-2493](https://trello.com/c/uZ2jJ38y/2493-events-remove-automated-this-event-is-closed-banner-message)

### Context

We currently display an automated message on events that are 'closed'. Instead, the events team want to use the existing `message` attribute to display a message when the event is closed. We already render this when its filled on the event page, so we simply need to remove the automated message.

- Remove automated event closed message

### DIFF
--- a/app/views/events/show.html.erb
+++ b/app/views/events/show.html.erb
@@ -7,12 +7,6 @@
 
   <p class="event-info__date"><%= format_event_date(@event, stacked: false) %></p>
 
-  <% unless event_status_open?(@event) || event_status_pending?(@event) %>
-    <div class="content-alert content-alert--fullwidth">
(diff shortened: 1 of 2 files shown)


jkempster34 (approved):
LGTM 💯 . So is this message going to be updated automatically in the CRM when the events closed date happens?

image

gemmadallmandfe:
@jkempster34 The message you highlighted will be updated **manually** in the CRM by the Events team. The message varies by event/time - for example, if an event is at capacity or if they are encouraging walk-ins.

We did ask if there were rules that could be applied to enable automation but their preference was to manage this message manually for now.

PR #1816, opened 2021-09-30
Add welcome guide content mappings

peteryates:
### Trello card

https://trello.com/c/cc9T5A0l/2497-welcome-guide-personalisation-mappings

### Context

The welcome guide will have several 'modes' that target candidates with particular attributes. This helper implements a `#welcome_guide_mode` method that uses the `WelcomeGuideModeSelector` to determine the right mode based on the contents of the `mailing_list_session`

To ensure the mode selector has everything it needs to make the decision, some additional fields (`:preferred_teaching_subject_id`, `:consideration_journey_stage_id` and `:degree_status_id`) have been added to `#prune`'s `leave` array.

The supported combinations are:

* Maths, final year student
* Maths, graduate
* Science, final year student
* Science, graduate
* MFL, final year student
* MFL, graduate
* English, final year student
* English, graduate
* Generic, final year student
* Generic, graduate

- Ensure extra fields remain in mailing list store
- Suppress some incorrect Rubocop rules
- Add some extra languages we reference to constants
- Add WelcomeGuideHelper to identify modes

### Guidance to review

Is this the right approach?


app/helpers/welcome_guide_helper.rb
```diff
@@ -0,0 +1,97 @@
+module WelcomeGuideHelper
+  # Contains logic that we'll used to decide what 'mode' the Welcome Guide will
+  # be shown in. The intention is for it to contain custom content for
+  # candidates who meet certain critera - e.g., a Maths final year student
+  class WelcomeGuideModeSelector
+    attr_reader :consideration_journey_stage_id, :preferred_teaching_subject, :committment_level, :degree_status_id
+
+    def initialize(mailing_list_session = {})
+      @committment_level              = mailing_list_session["committment_level"]
```
gemmadallmandfe, line 9: Typo - commitment not committment?

Is this predefined so not changeable, or something new?

peteryates: Ah, well spotted. Fixed and force pushed - thanks

gemmadallmandfe:
The way you've done degree status/subject make sense to me

What's the difference between `commitment_level` and `consideration_journey_stage_id`?

Now that we have the Havas content, let's discuss whether we can simplify this or whether it makes sense to build in flexibility for future use.

peteryates:
I think this PR is no longer necessary - the content options provided mean we only need either the subject or degree stage to switch content - not both.

Going to close it, we can resurrect it if we do think it'll be useful but at this stage it probably just confuses things more.

PR #1828, opened 2021-10-01
Dependabot updates for Ocotober 2021

peteryates:
### Trello card

N/A

### Context

A roundup of the batch of update PRs created by Dependabot this morning.

- Update secure_headers to 6.3.3
- Update webpacker_to 5.4.3
- Update sentry-ruby and sentry-rails to 4.7.3
- Update bootsnap to 1.9.1
- Update @babel/eslint-parser to 7.15.7
- Update js-cookie to 3.0.1
- Update @rails/webpacker to 5.4.3
- Quieten warnings when running yarn spec

### Guidance to review

I've had a good look around and everything seems fine. I held back the one **big** update (Stimulus 3.0.0) and these all appear to be minor version bumps.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,7 @@ gem "webpacker", ">= 5.4.3"
 # gem 'mini_magick', '~> 4.8'
 
 # Reduces boot times through caching; required in config/boot.rb
-gem "bootsnap", ">= 1.8.1", require: false
+gem "bootsnap", ">= 1.9.1", require: false
 
 # Manage multiple processes i.e. web server and webpack
 gem "foreman"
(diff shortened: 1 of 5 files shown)


ethax-ross (approved):
Gave it a quick test in the review app and it looks good. I wonder if we should update to Turbo when we upgrade Stimulus; there's a few useful fixes in Turbo that don't appear to have made it to Turbolinks, specifically it handles anchor links on the same page properly so we could drop this workaround:

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/webpacker/controllers/link_controller.js#L12-L20

peteryates:
Yeah - well up for making the jump. I suspect it'll be quite an involved change though 😓

ethax-ross:
Its not bad actually, I did it recently in a personal project - basically remove the JS/gem and install `hotwired/turbo-rails` then just rename all occurances of `turbolinks` to `turbo` and I think the initial import doesn't need the `.start()` but that's pretty much it 🤷

PR #1836, opened 2021-10-04
More blue links

jkempster34:
### Trello card
[Trello](https://trello.com/c/mn8w3HG5)

### Context
Silktide is reporting some links that do not pass their accessibility checks (blue link on grey background). It doesn't give a full list of occurrences because we don't pay, but I have been through each area with a $grey background and think I have got them all

Found some more failing links 👇 
dark-links

### DIFF
--- a/app/views/event_categories/show.html.erb
+++ b/app/views/event_categories/show.html.erb
@@ -37,8 +37,7 @@
       You can see <%= past_category_name(@type.id).downcase %>, including
       all questions and answers,
       <%= link_to "on this page",
-                  archive_event_category_path(pluralised_category_name(@type.id).parameterize),
-                  class: "dark-link" %>
+                  archive_event_category_path(pluralised_category_name(@type.id).parameterize) %>
     </p>
   </section>
 <% end %>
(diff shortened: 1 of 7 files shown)


ethax-ross (approved):
Looks good to me; I think if I was doing it I may have extracted the grey background to a mixin that does both - something like this, but its probably six and two threes:

```
@mixin gray-background {
  background-color: $grey;

  a {
    color: $dark-link;
  }
}
```

jkempster34:
> Looks good to me; I think if I was doing it I may have extracted the grey background to a mixin that does both - something like this, but its probably six and two threes:
> 
> ```
> @mixin grey-background {
>   background-color: $grey;
> 
>   a {
>     color: $dark-link;
>   }
> }
> ```

Would that stop the link colour cascading down? For example to these white boxes? 
image

jkempster34:
> > Looks good to me; I think if I was doing it I may have extracted the grey background to a mixin that does both - something like this, but its probably six and two threes:
> > ```
> > @mixin grey-background {
> >   background-color: $grey;
> > 
> >   a {
> >     color: $dark-link;
> >   }
> > }
> > ```
> 
> Would that stop the link colour cascading down? For example to these white boxes? image

Actually nevermind, that's a stupid question

PR #1838, opened 2021-10-05
Split test for a new age step in the mailing list wizard

ethax-ross:
### Trello card

[Trello-1879](https://trello.com/c/HvPB0Jrh/1879-test-adding-age-range-to-the-mailing-sign-up-process)

### Context

We want to experiment with asking users to provide their age when signing up to the mailing list. Before we actually start gathering this information we want to try a few different methods for asking users their age in order to observe the impact this has in drop-off/form completion rates. The data collected will not be retained to begin with and we want to have the following variations presented to users randomly:

- No age step (current state of form)
- Ask for date of birth (input for day, month, year)
- Ask for year of birth (dropdown)
- Ask for age range (radio button of different age brackets)

We want to track when users are presented with one of the different age input variations so that we can evaluate the data after running the experiment for a period of time.

If a user is presented with one variation they should continue to see that variation if they navigate forward/backward in the wizard flow or restart the sign up.

- Add timeliness gem
- Add new age step to mailing list wizard
- Serve random (but consistent) age question
- Track age step as virtual page view
- Rename mailing list feature specs (single page)
- Clean up the mailing list feature specs

### Guidance to review

I did some additional refactoring in the test suite so this PR contains a fair number of changes; best reviewed by-commit.

I haven't updated the single-page mailing list sign up form as we are not currently using this and it may not be used in the future (to be checked).

We still need content/copy for the age step variations; it should be fine to pick this up in a follow-up PR though as this is all behind a feature switch/not enabled in production.

| Date of Birth      | Year of Birth | Age Range |
| ----------- | ----------- | -------------- |
| Screenshot 2021-10-06 at 09 17 42      | Screenshot 2021-10-06 at 09 19 20       |  Screenshot 2021-10-06 at 09 19 53       |


gemmadallmandfe:
Thanks for this @ethax-ross 

I have asked Joe Hill for the age ranges that are useful for them, so we have realistic data

Re 'year of birth' - rather than having a dropdown, can we just have a standard Year field, like in the DOB variant? Scrolling down a list of years is not much fun as you get older :-)

peteryates:
Going to close this for now, resurrect if needed.

PR #1854, opened 2021-10-11
Swap phone number links

jkempster34:
### Trello card
[Trello](https://trello.com/c/FJ6nGXWx)

### Context
Telephone number links have been flagged as inaccessible due to how they are read by screen readers.

[The GOV.UK Design System suggests that phone numbers should not be links at all. Most devices automatically detect phone numbers if they support phone calls.](https://design-system.service.gov.uk/patterns/telephone-numbers/#do-not-display-telephone-numbers-as-links-on-devices-that-cannot-make-calls)

- Replace telephone number links with text only

### Guidance to review
See replacements below (some look the same):

| Before      | After |
| ----------- | ----------- |
|   image  |  image  |
|  image | image  |  
image  | image  |  
image | image  |  
image | image  |

### DIFF
--- a/app/components/footer/talk_to_us_component.html.erb
+++ b/app/components/footer/talk_to_us_component.html.erb
@@ -15,7 +15,7 @@
               <% end %>
               <div>
                 <p>Call us</p>
-                <a class="telephone" href="tel:08003892500" aria-label="Telephone">0800 389 2500</a>
+                <p class="telephone">0800 389 2500</p>
               </div>
             </div>
           </div>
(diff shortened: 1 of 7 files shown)


peteryates:
Ok, so this is working pretty well for me overall. On Windows Narrator everything is read out fine. On Linux (Orca) it's mostly there except the numbers are read out like "two thousand five hundred".

peteryates:
Having had a quick read, I think we should drop the link here altogether. The [guidance from the design system docs](https://design-system.service.gov.uk/patterns/telephone-numbers/#do-not-display-telephone-numbers-as-links-on-devices-that-cannot-make-calls) makes a good case for doing so:

> Do not display telephone numbers as links on devices that cannot make calls
>
> It’s possible to mark up telephone numbers as links, like this:
>
> `<a href="tel:+442079476330">020 7947 6330</a>`
>
> However, doing this will style telephone numbers as links, which is confusing on devices that do not support telephone calls, like most desktop machines.
>
> It’s also not necessary - most modern mobile browsers automatically detect telephone numbers and display them as links anyway.
>
> If you do need to mark up your telephone number as links, for example, to support a device that cannot automatically detect them, make sure they do not display as links on devices that cannot make calls.

PR #1856, opened 2021-10-12
Combine MFL into single languages entry

peteryates:
### Trello card

N/A

### Context and changes

We've combined all the languages, which all shared the same bursary amount, into a single Languages entry.

It also encompasses 'ancient languages' like Latin and Greek, and is labelled as such.

As a result, the bursary has been removed from classics.

### Guidance to review

Is "English as a second language" a _language_, should it have a bursary attached?


christinehitch:
@peteryates 
Happy with this Pete and have confirmed with FI team that 'English as a second language' counts as English to no bursary.

PR #1903, opened 2021-10-26
Upgrade to Turbo (take 2)

ethax-ross:
Reverts DFE-Digital/get-into-teaching-app#1899 - we had to revert this due to an issue with the current GTM container tag triggers.


peteryates (approved):

peteryates:
Just a thought on this, you might want to change the formbuilder over to `turbo` mode too, see [here](https://www.rubydoc.info/gems/govuk_design_system_formbuilder/GOVUKDesignSystemFormBuilder)

PR #1926, opened 2021-11-01
Pull out non-cache code from StaticPages module

ethax-ross:
### Trello card

[Trello-2592](https://trello.com/c/NarUxiQl/2592-review-caching-method)

### Context

The `StaticPages` module seems to mostly deal with caching except for when its pulled into the `PagesController`, where it has some handling code for rendering 404s if a template is not found.

As the `PagesController` is the only place this is used I think it makes sense to extract it to that controller and re-purpose the `StaticPages` module as a `StaticPageCache` module.

- Pull out non-cache code from StaticPages module

### Guidance to review

This is the first step towards updating the caching behaviour of the site so that it expires etags when the content changes rather than on every deploy.


peteryates (approved):
Much clearer

PR #1930, opened 2021-11-01
Improve caching behaviour of the website

ethax-ross:
### Trello card

[Trello-2592](https://trello.com/c/NarUxiQl/2592-review-caching-method)

### Context

The website is currently setup to serve ETags that change on each deploy (though for some reason we never seem to serve a 304 not modified -- my thoughts on why that is are below).

We have a couple of options for caching; ETags, Cache-Control headers and a static web page cache. I think a combination of Cache-Control headers and static web page caching is going to give us our best performance. The ETag method currently in use will hit Rails every time and basically just saves on not having to send a response when nothing has changed - we still construct the response, which takes time. 

If we add in a static web page cache it means that the Rails stack is hit only on the first request to each route, thereafter the generated HTML is cached and served from the public directory (we could improve this further by using nginx to prevent hitting Rails at all). We can also add in a Cache-Control header with a short TTL of 5 minutes or so, which will allow pages to be cached for subsequent views however changes may not be reflected after a deploy for up to 5 minutes. Using static web page caching also removes the ability to use ETags, but I think the performance we get will outweigh the 304 not modified response benefits anyway.

- Remove ETag caching

Our ETag caching doesn't appear to currently work; the ETags don't match between requests for some reason and even if I force them to match we never serve a 304 Not Modified. I _think_ its because we have an explicit `render` call in our `PagesController#show` action and according to the Rails docs the not modified behaviour is invoked if we don't call `render` explicitly.

I'm removing the implementation for now (which means we will default to using Rails ETags that compute on the response body rather than the last deployment SHA - which means we should be no worse off anyway).

The next step is adding a static page cache, which should actually give us a decent performance boost as we won't need to hit Rails after the first request for a static page. It means ETags won't work but I expect this to be faster regardless.

- Add actionpack-page_caching gem

The page cache was removed from Rails core, so bringing it back in with the gem.

- Resolve conflicts in Pages namespace

We use the `Pages` namespace which conflicts with the `actionpack-page_caching` gem, so we need to update our references to be absolute (leading `::`).

Reworks the `crsf_meta` to only include it in templates that submit a form as we can't include it in cached pages.

- Specify actions to be cached

These actions serve static HTML and are safe to cache.

- Add small TTL on Cache-Control headers

Setup the static pages to serve from `/public/cached_pages` via a new `ActionDispatch::Static` middleware so that we can specify a 5 minute cache TTL (otherwise the global 1 year TTL is applied from `/public`).

### Guidance to review

I need to ensure we're fine with a short TTL on the Cache-Control headers; I can't think of a problem in doing this, but it will mean changes take a little longer to reflect in production.

Due to the way `caches_page` adds an `after_action` we have no way of enabling cache on a per-test basis in RSpec in order to test the caching behaviour.


peteryates (approved):
All looks good to me, agree it'd be a pain to test.

PR #1931, opened 2021-11-02
Use event summary as description

ethax-ross:
The event description contains HTML however event summaries should not; they are also much shorter so are better suited to the small space available in the event search result boxes. I've left the code to strip any HTML in place as it won't hurt in case someone does add some into that field in the future.


peteryates (approved):
Looks good, seen some dodgy bits of markup in the descriptions! 😅

PR #1932, opened 2021-11-02
Remove Bradford SCITT from AO providers

gemmadallmandfe:
### Context
Request via helpdesk to remove Bradford Birth to 19 SCITT from AO providers page.


peteryates (approved):
Looks good, let's go!

PR #1936, opened 2021-11-03
Update images for event structured data

ethax-ross:
### Trello card

[Trello-2590](https://trello.com/c/iZMhdQZY/2590-events-add-default-images-for-structured-data-event-listings)

### Context

Instead of using the building image we want to show one image for online events and a different image for in-person events. The images are in different aspect ratios as per Google's guidance.

- Update images for event structured data

### Guidance to review

I've ran it through the [schema markup testing tool](https://search.google.com/test/rich-results) and it looks good (the other warnings are unrelated and are optional).


peteryates (approved):
Makes sense

PR #1947, opened 2021-11-03
Rework how page cache is called

ethax-ross:
### Trello card

[Trello-2624](https://trello.com/c/ACqZyeCE/2624-fix-next-gen-image-formats-not-serving)

### Context

As the gem calls its own `after_action` internally we can't get it to call in the correct sequence due to our image post-processing happening in our `ApplicationController` and the `caches_page` call happening in our sub-class controllers

Instead of calling the provided `caches_page` method that sets an `after_action` we instead call the underlying `cache_page` method in our own `after_action`, this ensures the ordering remains correct.

- Rework how page cache is called


peteryates (approved):
Nice 👍🏽

Definitely easier to follow

peteryates (approved):

PR #1949, opened 2021-11-04
Fix page caching on funding your training page

ethax-ross:
### Trello card

[Trello-2629](https://trello.com/c/2TaMA1rF/2629-revisit-new-caching-method)

### Context

When the new caching behaviour was turned on in production we noticed the funding your training page was not responding correctly. This is because we use query string parameters to determine the state of the form on load, so we can't use the static page cache here.

- Re-enable static page cache

This was disabled due to a bug with the funding your training page/widget - re-enabling as we're fixing the issue.

- Exclude /funding-your-training from static page cache

Route `/funding-your-training` to its own controller action so that it doesn't get cached. This page doesn't work with the static page cache due to its use of query string parameters to construct the form in a given state on page load.

### Guidance to review

Have I missed any other pages that have forms/use query strings/may not work with the static page cache?

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -33,10 +33,9 @@ private
     process_images
     process_links
 
-    # TODO: turn back on once the funding widget is fixed.
-    # if perform_caching? && action_name.to_sym.in?(static_page_actions)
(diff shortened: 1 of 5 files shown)


peteryates (approved):
Looks good, makes sense

PR #1952, opened 2021-11-04
Fix image post-processing for hosted assets

ethax-ross:
### Trello card

[Trello-2624](https://trello.com/c/ACqZyeCE/2624-fix-next-gen-image-formats-not-serving)

### Context

When the app is running in the preprod/prod environment the assets are hosted on a different domain, so the various image `src` attributes contain absolute URLs instead of paths.

Update the `NextGenImages` and `ResponsiveImages` to account for `src` attributes containing either a domain or path.

Even though the assets are hosted elsewhere they remain in the docker image so the various `File` checks still work.

- Fix image post-processing for hosted assets

- Correctly serve webp images

`Rack::Mime` doesn't support webp images out of the box; to get around this we have to tell it to use the correct content type for the extension.

Oddly, its been fine serving webp images as `text/plain` up until we added in a `Access-Control-Allow-Origin` header when we started serving assets on a different domain. A side-effect of this is that the content-type has to now be correct or the request is blocked by cross-origin read blocking (CORB). More info here:

https://www.chromium.org/Home/chromium-security/corb-for-developers

Specifically:

> For example, a response served with a “X-Content-Type-Options: nosniff” response header and an incorrect “Content-Type” response header may be blocked. This could, for example, block an actual image which is mislabeled as “Content-Type: text/html” and “nosniff.” If this occurs and interferes with a page’s behavior, we recommend informing the website and requesting that they correct the “Content-Type” header for the response.

- Flush asset cache

Updates the path prefix with a version in order to flush the assets as we now have webp images cached in the staging CloudFront with the wrong content-type (from my testing).

I can't find a better way of doing this; Webpack seems to let you define filenames explcitly but I'd have to do it for every
extension type which is awkward. Sprockets lets you specify the asset version which weighs into the digest value, however I can't find a corresponding option for Webpack and instead the easiest thing seems to be to change the pack path to include a version.

### Guidance to review

I pushed this build to our staging environment (which has hosts assets on a different domain) in order to test it as it appears in prod:

Screenshot 2021-11-05 at 10 01 36

### DIFF
--- a/spec/lib/next_gen_images_spec.rb
+++ b/spec/lib/next_gen_images_spec.rb
@@ -5,24 +5,26 @@ describe NextGenImages do
   describe "#html" do
     subject { instance.html }
 
-    let(:original_ext) { File.extname(original_src) }
-    let(:body) { "<img src=\"#{original_src}\">" }
+    let(:original_ext) { File.extname(original_src_path) }
+    let(:body) { %(<img src="#{original_src_url}">) }
     let(:instance) { described_class.new(body) }
+    let(:asset_host) { "" }
+    let(:original_src_url) { "#{asset_host}#{original_src_path}" }
 
     before do
       allow(File).to receive(:exist?).and_return(false)
-      allow(File).to receive(:exist?).with("#{Rails.public_path}/#{original_src}").and_return(true)
+      allow(File).to receive(:exist?).with("#{Rails.public_path}#{original_src_path}").and_return(true)
     end
 
     context "when the original image is a jpg" do
-      let(:original_src) { "path/to/image.jpg" }
+      let(:original_src_path) { "/path/to/image.jpg" }
 
       context "when there are no next-gen images" do
-        it { is_expected.to include("<picture>#{source(original_src, 'image/jpeg')}<img src=\"#{original_src}\"></picture>") }
+        it { is_expected.to include(%(picture>#{source(original_src_url, 'image/jpeg')}<img src="#{original_src_url}"></picture>)) }
       end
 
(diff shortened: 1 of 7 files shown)


spec/lib/next_gen_images_spec.rb
```diff
@@ -5,64 +5,75 @@
...
+      let(:original_src_path) { "/path/to/image.svg" }
 
-      it { is_expected.to include("<picture>#{source(original_src, 'image/svg+xml')}<img src=\"#{original_src}\"></picture>") }
+      it { is_expected.to include("<picture>#{source(original_src_path, 'image/svg+xml')}<img src=\"#{original_src_url}\"></picture>") }
     end
   end
 
   def source(src, mime_type)
-    "<source srcset=\"#{src}\" type=\"#{mime_type}\"></source>"
+    "<source srcset=\"#{asset_host}#{src}\" type=\"#{mime_type}\"></source>"
```
peteryates, line 77: worth using an alternative delimiter to avoid escaping?

```suggestion
    %(<source srcset="#{asset_host}#{src}" type="#{mime_type}"></source>)
```

peteryates (dismissed):
Looks fine to me, everything in the review app seems to be working as expected too

spec/lib/next_gen_images_spec.rb
ethax-ross: I'll never get used to that as a string delimiter, but yeah agreed I'll push an update 😅

peteryates (approved):

PR #1955, opened 2021-11-08
adding blog - being myself

emilytraff:
Adding a blog from Medium that always performed well there.


peteryates (approved):
Looks good to me

PR #1956, opened 2021-11-09
Update README with blog post instructions

ethax-ross:
Starts a content README that contains instructions on creating blog posts, specifically around creating suitable thumbnails images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,3 @@ If a user does not require a role (site-wide authentication), the role credentia
 ```yaml
 username|password
 ```
-
-
-
--- a/doc/content.md
+++ b/doc/content.md
@@ -1,0 +1,39 @@
+# Content Documentation/Guide
+
+## Creating a Blog Post
+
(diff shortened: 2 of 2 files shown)


gemmadallmandfe:
Thanks for this @ethax-ross !

I wonder if it is worth starting off a separate readme that relates to content only - ie for editors rather than devs?

We can then build this out with other info as and when

Any thoughts @peteryates ?

peteryates:
This looks sensible to me. Worth cc'ing @emilytraff and @christinehitch in on this - I believe the main change is to add a thumbnail in addition to a regular-sized image.

The instructions definitely make it clear what needs to be in place though.

ethax-ross:
I've moved this into its own markdown document `/doc/content.md` so that we can keep it separate from the main README and maybe expand it with more guides going forward.

gemmadallmandfe:
@ethax-ross Fab - thanks for doing this. Think this makes a lot of sense and we can add more to it over time.

sfawcett123 (approved):

PR #1959, opened 2021-11-09
Improve resizing behaviour in the footer

peteryates:
Now it will be a column on mobile and a row on tablet/desktop. This stops it from pushing the page out wider than it needs to be.

| Before | After |
| ------ | ------- |
| Screenshot from 2021-11-09 16-17-12 | Screenshot from 2021-11-09 16-15-05 |


gemmadallmandfe (approved):
Looks good to me!

PR #1961, opened 2021-11-10
Don't show external icon for links to assets

ethax-ross:
### Trello card

[Trello-2643](https://trello.com/c/lN13BuX4/2643-blue-external-icon-shouldnt-be-on-pdf-download-button)

### Context

As our assets are now hosted on a different domain (in test/prod) they are being picked up as external links and an icon is being added incorrectly.

Update `LinkController` so it knows about the asset URL and treats it as internal.

- Don't show external icon for links to assets

### Guidance to review

| Before      | After |
| ----------- | ----------- |
| Screenshot 2021-11-10 at 10 03 35      | Screenshot 2021-11-10 at 10 03 44       |

We could probably DRY up the body tag but I will wait until the new GTM container PRs have been deployed as they simplify the existing body tag a lot and it will be easier to do then.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,7 +1,7 @@
 <!doctype html>
 <html lang="en" class="govuk-template">
   <%= render "sections/head" %>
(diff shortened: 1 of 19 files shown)


gemmadallmandfe:
Looks good to me!

peteryates (approved):
Looks good.

PR #1962, opened 2021-11-10
Allow pages to be easily excluded static page cache

ethax-ross:
### Trello card

[Trello-2642](https://trello.com/c/mIknuOTy/2642-add-a-simple-way-to-omit-pages-from-caching)

### Context

Some pages are not suitable for caching (usually due to the use of query parameters/forms). We need a way to easily exclude these from the default static page caching behaviour on the `show` action.

- Allow pages to be easily excluded from static page cache

Add `DYNAMIC_PAGE_PATHS` which will be excluded from the static page cache.

### Guidance to review

There's no way we can test this unfortunately due to caching being disabled in the test env by default 😞 I have manually tested however by adding a route to `DYNAMIC_PAGES` and checking it it doesn't get cached in the preview environment.


peteryates (approved):
Nice!

PR #1964, opened 2021-11-10
Add guidance on safeguarding question

Tomcowan1:
### Trello card

https://trello.com/c/b9QYDp7S/2647-review-reference-advice-re-keeping-children-safe

### DIFF
--- a/app/views/content/tips-on-applying-for-teacher-training.md
+++ b/app/views/content/tips-on-applying-for-teacher-training.md
@@ -117,6 +117,8 @@ Choose referees who could comment on things like your:
 * transferable skills
 * academic skills
 
+Your referee will aslo be asked if they know of any reason why you should not work with children.
+
 Ideal referees could include:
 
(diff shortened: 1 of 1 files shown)


Tomcowan1:
Hi @gemmadallmandfe have added a line into the current "Tips on applying" page to bring it in line with Apply. I've used exactly they same wording.

app/views/content/tips-on-applying-for-teacher-training.md
```diff
@@ -117,6 +117,8 @@ Choose referees who could comment on things like your:
 * transferable skills
 * academic skills
 
+Your referee will aslo be asked if they know of any reason why you should not work with children.
```
gemmadallmandfe, line 120:
```suggestion
Your referee will also be asked if they know of any reason why you should not work with children.
```

gemmadallmandfe (approved):
Looks good to me!

PR #1967, opened 2021-11-11
Re-introduce CSRF meta tag on internal layout

ethax-ross:
I did a refactoring on this when I brought in static page caching but missed adding it to the internal events layout.


jkempster34 (approved):
🙌

PR #1968, opened 2021-11-11
Add Google Optimize sync script

ethax-ross:
### Trello card

### Context

We want to run A/B tests on certain pages using Google Optimize. Including it via GTM results in a page flicker when the variant is served; instead we need to include it in the `<head>` tag synchronously, but only on pages that are running an active experiment as there will be a performance hit to doing this.

I've not prevented the script from loading until the user accepts cookies as I _believe_ that it won't drop a cookie unless Google Analytics has fully loaded, which requires consent. If this turns out not to be the case we can wrap it in a Stimulus controller so it only drops on accepting cookies.

- Add Google Optimize sync script


peteryates (approved):
Looks good

PR #1969, opened 2021-11-11
Perform beta redirect before static page cache

ethax-ross:
### Trello card

[Trello-2654](https://trello.com/c/OMCl1OIG/2654-fix-beta-redirect-not-working)

### Context

The beta redirect currently happens as a Rails route, which used to work fine but now that the pages are statically cached the routes aren't evaluated on every request. If the page matches a static page it will be served by Rack; to re-instate the redirect we need to do so in Rack before the `ActionDispatch::Static` middleware.

- Perform beta redirect before static page cache

### Guidance to review

Added to the test environment as well so that the existing rspec test still passes (it was passing despite not working due to the caching being disabled in the test env; not a lot we can do about that.)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -83,6 +83,8 @@ gem "better_html"
 
 gem "dfe_wizard", github: "DFE-Digital/dfe_wizard"
 
+gem "rack-host-redirect"
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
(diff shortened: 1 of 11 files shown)


peteryates (approved):
Looks good. We can't be getting too much traffic to `beta.` now though surely?

ethax-ross:
Nah we don't seem to get many/any; we should probably look to decomission it, but for now this will stop Google having a potential fit.

PR #1978, opened 2021-11-12
Simplify the mailing list logic on completion

peteryates:
Now we're going ahead with the WG update the specs so they expect the WG content where the conditions are matched. Also flip the `if` statement so it reads more easily; now we're showing it everywhere except prod.


ethax-ross (approved):
Much easier to follow 💯

PR #1979, opened 2021-11-15
Ensure Google Optimize respects cookie consent

ethax-ross:
> ⚠️ WIP

### Trello card

### Context

Initially I thought Google Optimize wouldn't drop a cookie until GA was pulled in, however that doesn't appear to be the case so we need to prevent it firing before the user has opted-in to non-functional cookies.

- Switch to loading Google Optimize on cookie acceptance

This creates a new `GoogleOptimizeController` that does that for the legacy tracking pixel method; a follow-up commit will ensure it also works for the new tracking pixel setup.

- Load Google Optimize using new tracking setup

We want to load Google Optimize in the app for the new GTM-based tracking setup as well, so it continues to work when we make the switch to using the new container.


ethax-ross:
Closing this off as we've decided we want to tackle it in a different way (blur the acceptance background until the user accepts and loading the optimize script earlier to avoid a flicker).

PR #1981, opened 2021-11-15
Fix the scribble signup display

peteryates:
### Context

The page is 404ing in the test env due to a typo in the partial name

Fix the typo and add some better styling to the scribble event box.

### Guidance to review

There's now an online event accessible from the test so it should be visible and look fine when viewed. Thanks @jkempster34 for helping get this set up!

Screenshot from 2021-11-15 16-28-04


jkempster34 (approved):
💯

PR #1982, opened 2021-11-16
Send welcome guide variant on ML sign up

ethax-ross:
### Trello card

[Trello-2661](https://trello.com/c/catUOcuk/2661-update-api-to-include-welcome-guide-query-parameters)

### Context

When a candidate signs up to the Mailing List we want to populate the `welcome_guide_variant` attribute with the query string that can be used to serve the correct version of the welcome guide in an email that will later be sent to the candidate.

We only want to populate this field if the candidate qualifies for the welcome guide, leaving it as `nil` if they don't.

- Send welcome guide variant on ML sign up

### Guidance to review

The API/CRM only support this new field in dev at the moment, but it _shouldn't_ cause issues if its set in test/prod payloads; it'll just be ignored

### DIFF
--- a/app/models/mailing_list/wizard.rb
+++ b/app/models/mailing_list/wizard.rb
@@ -59,10 +59,28 @@ module MailingList
   private
 
     def add_member_to_mailing_list
-      request = GetIntoTeachingApiClient::MailingListAddMember.new(export_camelized_hash)
+      request = GetIntoTeachingApiClient::MailingListAddMember.new(construct_export)
       api = GetIntoTeachingApiClient::MailingListApi.new
       Rails.logger.info("MailingList::Wizard#add_mailing_list_member: #{AttributeFilter.filtered_json(request)}")
       api.add_mailing_list_member(request)
+    end
+
+    def construct_export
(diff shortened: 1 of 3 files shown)


app/models/mailing_list/wizard.rb
```diff
@@ -59,10 +59,28 @@ def exchange_magic_link_token(token)
...
+      request = GetIntoTeachingApiClient::MailingListAddMember.new(construct_export)
       api = GetIntoTeachingApiClient::MailingListApi.new
       Rails.logger.info("MailingList::Wizard#add_mailing_list_member: #{AttributeFilter.filtered_json(request)}")
       api.add_mailing_list_member(request)
     end
+
+    def construct_export
+      export = export_camelized_hash
+
+      show_welcome_guide = ApplicationController.helpers.show_welcome_guide?(
```
ethax-ross, line 71: The `show_welcome_guide?` method should perhaps be pulled out into its own module rather than pulling in a view helper here, but I'm not sure if its worth it 🤷‍♂️

peteryates (approved):
Looks good, makes sense

PR #1983, opened 2021-11-16
Adjust the link checker so it works page-by-page

peteryates:
### Trello card

https://trello.com/c/D8aK0Vw9/2566-investigate-listing-urls-when-the-missing-link-tool-outputs-slack-messages

### Context and changes

This moves the loop outside of the scenario so tests are run at a more-granular level - previously all errors in the run would be reported in a single failure.


ethax-ross (approved):

peteryates:
The underlying behaviour's changed, will re-open if we still want this but it should report fewer fails now.

PR #1985, opened 2021-11-16
Refactor global hidden style

ethax-ross:
### Trello card

[Trello-2644](https://trello.com/c/eiXVZiU3/2644-clean-up-hidden-classes)

### Context

There was a global `.hidden` class that was only applied when JS is enabled; this is confusing and a pain to override, as you often want the element to be hidden regardless of JS status when applying a `hidden` class.

Rename `.hidden` to `.hidden-when-js-enabled` and create a new `.hidden` class that will always hide an element when applied. It's decorated with `important!` so that custom components that have, for example, `display: flex` will still be
hidden when the class is applied.

Fix typo in style attribute that should be `margin-bottom`, not `bottom-margin`.

- Refactor global hidden style

### DIFF
--- a/app/webpacker/styles/layout.scss
+++ b/app/webpacker/styles/layout.scss
@@ -37,8 +37,16 @@ main,
   width: 1px;
 }
 
-.js-enabled .hidden {
-  display: none;
+.js-enabled {
+  .hidden-when-js-enabled {
+    display: none;
+  }
+}
+
+.hidden {
+  // We use important! to ensure any component-level
+  // display values (such as flex) are still hidden.
+  display: none !important;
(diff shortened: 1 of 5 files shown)


app/webpacker/styles/layout.scss
```diff
@@ -55,7 +63,7 @@ main {
   }
 
   &--no-bottom-margin {
-    bottom-margin: 0;
+    margin-bottom: 0;
```
peteryates, line 66: 🤦🏽‍♂️

peteryates (approved):
Definite improvement, much easier to follow!

PR #1994, opened 2021-11-18
Prevent external link icon appearing on links to internal assets

ethax-ross:
### Trello card

[Trello-2643](https://trello.com/c/lN13BuX4/2643-blue-external-icon-shouldnt-be-on-pdf-download-button)

### Context

In the test/prod environments our internal assets are hosted on an different domain, which is then treated as external by our code that adds the external icon to links.

Ensure externally hosted 'internal' assets are not treated as external by the link checker.

- Prevent external link icon appearing on links to internal assets

### Guidance to review

I thought I had fixed this a while ago, but it turns out we are doing this logic in multiple places; the `LinkController` (JS) and `ApplicationController` (Ruby). I've created another ticket to consolidate these two approaches: https://trello.com/c/MUYUEYgC/2671-consolidate-how-we-add-external-link-icons

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -75,7 +75,10 @@ private
 
       next if classes && (classes & classes_to_suppress).any?
 
-      if anchor[:href].include?("//")
+      external_link = anchor[:href].include?("//")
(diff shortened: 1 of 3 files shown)


peteryates (approved):
Makes sense

PR #1997, opened 2021-11-18
Fix broken link checker failing on assets

ethax-ross:
### Trello card

[Trello-2672](https://trello.com/c/Jn29oHA3/2672-fix-broken-link-checker)

### Context

When the broken link checker finds a link that goes to an asset it replaces `packs-test` with `packs`, however the production assets have since been versioned to `packs/v1`, so we need to replicate this in the test environment in order to have the asset paths ultimately match.

- Fix broken link checker failing on assets

### DIFF
--- a/config/webpacker.yml
+++ b/config/webpacker.yml
@@ -80,8 +80,10 @@ test:
   <<: *default
   compile: false
 
-  # Compile test packs to a separate directory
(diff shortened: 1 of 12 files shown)


peteryates (dismissed):
Ah good catch

peteryates (approved):

PR #2002, opened 2021-11-18
Disable GZip compression for certain assets

ethax-ross:
### Trello card

[Trello-2635](https://trello.com/c/kObYlPsJ/2635-disable-gzip-compression-for-images)

### Context

We currently GZip compress all assets, however its recommended that you don't GZip images and other binary files as they are usually already compressed and running them through GZip leads to slightly larger files (and wasted processing).

Subclass `Rack::Deflater` so that we can exclude certain requests.

- Disable GZip compression for certain assets

### Guidance to review

Tested locally and it seems to correctly prevent gzipping binary assets whilst still compressing HTML pages, etc:

```
❯ curl -Is -H "Accept-Encoding: gzip" "http://0.0.0.0:3000/packs/v1/media/images/content/hero-images/0012-cb6435a02b879e8df922882afba620a8.jpg" | grep "Content-Encoding"

❯ curl -Is -H "Accept-Encoding: gzip" "http://0.0.0.0:3000/" | grep "Content-Encoding"
Content-Encoding: gzip
```

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -4,8 +4,13 @@ require_relative 'config/environment'
 
 require 'prometheus/middleware/collector'
 require 'prometheus/middleware/exporter'
+require_relative 'lib/rack/deflater_with_exclusions'
 
-use Rack::Deflater
+EXTENSIONS_TO_EXCLUDE = %w(.jpg .jpeg .png .gif .pdf).freeze
+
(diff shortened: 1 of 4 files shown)


peteryates (approved):
Makes sense

PR #2006, opened 2021-11-18
Add ability to easily exclude a content page from sitemap/index

ethax-ross:
### Trello card

[Trello-2617](https://trello.com/c/ERCqXqVM/2617-enable-pages-to-be-excluded-from-the-sitemap)

### Context

When we are running A/B tests we will need to de-index any variants that we don't want to let Google know about.

- Add ability to easily exclude a content page from sitemap/index

Provide simple mechanism to exclude a page from the sitemap and serve a noindex header; include `noindex: true` in the frontmatter.

Refactor our existing test A/B pages to use the new mechanism.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -44,19 +44,6 @@ class PagesController < ApplicationController
     render_page(params[:page])
   end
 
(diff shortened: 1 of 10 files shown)


peteryates (approved):
Looks good

PR #2011, opened 2021-11-19
Return 404 for past teaching events

ethax-ross:
### Trello card

[Trello-2675](https://trello.com/c/d6dk89B4/2675-address-why-so-many-historic-events-are-still-on-site-and-appear-in-serps)

### Context

If a teaching event is in the past we no longer want to serve it to the user, instead returning a 404 with the custom event not found page.

The 404 response should prompt Google to deindex the page and the not found events page directs the user to search for another event.

We define "in the past" as being on the previous day, rather than a time-granular comparison; this makes it consistent with the search results so if we return an event the starts at 9am and its now 10am it will still be accessible for the remainder of the day.

The exception to this is with Online Q&A events, which have an archive so that users can still view transcripts/comments of the event after it has happened. We want to still allow users to access these events and Google to index them.

- Return 404 for past teaching events

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -31,7 +31,7 @@ class EventsController < ApplicationController
     breadcrumb "events.search", :events_path
 
     @event = GetIntoTeachingApiClient::TeachingEventsApi.new.get_teaching_event(params[:id])
-    render_not_found && return if @event.status_id == pending_event_status_id
+    render_not_found && return unless viewable_event?(@event)
 
     @page_title = @event.name
     @front_matter = { "description" => @event.summary }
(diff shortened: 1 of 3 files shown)


app/controllers/events_controller.rb
```diff
@@ -56,9 +56,17 @@ def not_available_path
 
 private
 
+  def viewable_event?(event)
+    pending = event.status_id == pending_event_status_id
+    in_past = event.start_at.to_date < DateTime.now.utc.to_date
+    online_q_a = event.type_id == GetIntoTeachingApiClient::Constants::EVENT_TYPES["Online event"]
+
+    !pending && (!in_past || online_q_a)
```
peteryates, line 64: I think the `!`s make the final line difficult to parse, maybe it could be flipped a bit `!pending && (future || online_qa)`. I'd go further but can't work out what the opposite of pending is!

peteryates (approved):
Looks great, left one minor suggestion

app/controllers/events_controller.rb
ethax-ross: Yeah I think you're right - the opposite of pending in this case is either `open` or `closed`, which is a bit yuk... I guess 'live' kind of fits but doesn't really encapsulate the closed state so much, so I'll maybe go with `not_pending && (in_future || online_q_a)` for clarity

peteryates (approved):
Looks great!

PR #2015, opened 2021-11-22
Add descriptions for event types in a details box

peteryates:
### Trello card

https://trello.com/c/bb8rAGQr/2663-events-build-event-type-explanation-popup-and-ttt-page

### Context and changes

This is an alternative to the proposed modal popup. It has some usability advantages:

* no accessibility concerns (ie, how do we announce to screen readers that an information modal has occupied the screen?)
* no need for a non-JS fallback
* no additional code/tests to maintain

Disadvantages:

* isn't as fancy looking

https://user-images.githubusercontent.com/128088/142866806-584468ff-97bd-4bd4-b467-6c7721757722.mp4

### DIFF
--- a/app/views/teaching_events/index.html.erb
+++ b/app/views/teaching_events/index.html.erb
@@ -5,6 +5,24 @@
     <p>
       All events listed here are designed to help you get into teacher training. They are all free to attend.
     </p>
+
+    <details>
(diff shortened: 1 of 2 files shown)


gemmadallmandfe (approved):
Looks good to me!

(Content will probably need changing but that can be done later)

PR #2017, opened 2021-11-22
Rubocop update

ethax-ross:
### Trello card

[Trello-2515](https://trello.com/c/lIfTKYRH/2515-upgrade-to-rubocop-govuk-410)

### Context

We want to use the latest version of Rubocop/govuk so that our code is consistent and well linted.

- Update Rubocop/govuk gem
- Auto-fixes for udpated Rubocop
- Increase max for RSpec/MultipleMemoizedHelpers rule
- Fix Lint/MissingSuper rule
- Fix RSpec/StubbedMock rule
- Fix Style/OptionalBooleanParameter rule
- Fix Security/Open rule
- Fix Rubocop namespace warning
- Fix Rails/ContentTag rule
- Exclude Rails/SaveBang rule

### Guidance to review

We should probably aim to reduce the `Max` setting on the `RSpec/MultipleMemoizedHelpers` rule going forward, but this will at least prevent it getting any worse for now.


peteryates (approved):
all looks fine to me

PR #2018, opened 2021-11-22
Better align the atteneding providers content

peteryates:
Minor alignment change

Screenshot from 2021-11-22 16-40-08


gemmadallmandfe (approved):
Happy to approve so we can review properly in our test environment with production event data

PR #2024, opened 2021-11-23
Replace placeholder welcome guide photos

peteryates:
We've been waiting a while for the final images from Havas.

* [x] Food tech teacher Andrew
* [x] PE teacher Jack


gemmadallmandfe:
Jack's photo on the main page also needs replacing (along with the alt text) @peteryates 

image

It also looks like Abigail's photo is missing alt text on this page as well

If you can do this, let's get it merged down for final review

gemmadallmandfe (approved):
Happy for this to be merged so we can test the WG in full in Test before launch

PR #2025, opened 2021-11-24
Rework Google Optimize for consent

ethax-ross:
### Trello card

### Context

The current way we include Google Optimize works but is not GDPR compliant. We need to:

- Make sure we only drop the script when a user has consented to non-functional cookies
- Ensure it works when transitioning to an experiment page via Turbolinks and also on an initial page view
- Eliminate any page 'flicker' (where the control is shown before they are redirected to a variant)
- Only include the script on pages running an experiment (it has a negative performance impact we want to minimise)

To achieve this, the PR:

- Extracts the Google Optimize JS into its own pack that is loaded synchronously (so it executes as early as possible)
- Drops the Optimize script on initial page load if they have consented and are on an experiment page
- Eliminates flicker by hiding the page content until the Optimise script has initialised
  - Ensure a fallback shows the content in case the Optimize script fails to load/is not available for any reason
- Blurs the content if they land on an experiment page and haven't yet accepted cookies (avoids them seeing the control page before they get redirected to a variant)
  - When they accept cookies we can simply reload the page so the 'initial page view' scenario drops the script
- Intercepts Turbolinks when transitioning to an experiment page and forces a full page load, which will drop the Optimize script (assuming they are opted-in)

- Add Google Optimize ID to all environments

Google Optimize doesn't like to run against `localhost`, so in order to test it before going to prod we can enable it in other environments. There's no harm in doing this, as it will only execute if an experiment is active.

- Add new Google Optimize JS implementation

- Switch to the new Google Optimize implementation

Removes the previous partial that pulled in Google Optimize regardless of cookie consent. Pulls in the new Google Optimize pack and configures it accordingly.

### Guidance to review

This can be tested by visiting the [test page](https://review-get-into-teaching-app-2025.london.cloudapps.digital/test/a). You should get served a variant - its currently setup for an A/B test where the 'Test A' heading changes to blue on the variant.

It doesn't feel great that the google optimize script modifies the cookie dialog background to swap it out for a blurred one, but the. only other solution I could think of would be to have the cookie dialog controller become aware of experimentation paths with felt equally dirty...


peteryates (approved):
Looks good and everything seems to work fine. Definitely no tracking cookies once they're rejected.

Blur on the popup overlay looks good too!

gemmadallmandfe (approved):
Looks good to me - cookie only drops when consent given, and I've received both control and variant.

I agree that the blur is a bit odd but given that it is only on pages we are testing with, happy to go ahead with this

PR #2026, opened 2021-11-24
Correct location path reference

ethax-ross:
I was using `window.location.path` incorrectly; it should be `window.location.pathname` to get the current path for the page view event.


peteryates (approved):
Looks good

PR #2031, opened 2021-11-25
Fix linting errors

ethax-ross:
These must have snuck into master in between the updated Rubocop PR being raised and merged.


peteryates (approved):
Looks good

PR #2033, opened 2021-11-26
Switch noindex to meta tag method

ethax-ross:
### Trello card

[Trello-2683](https://trello.com/c/rS8Bpyfc/2683-check-noindex-behaviour)

### Context

We were preventing search engines from indexing a web page by returning a `X-Robots-Tag` header. This works for the first request that hits Rails, but if the page is then served by the static page cache middleware the header is no longer present.

Instead, we can add a `meta` tag to the page, which will then be part of the static page cache and return on every request.

- Switch noindex to meta tag method


peteryates (approved):
Looks good.

PR #2036, opened 2021-11-26
Make time format consistent with GOV.UK design system

ethax-ross:
### Trello card

[Trello-2646](https://trello.com/c/JB0MDxkZ/2646-accessibility-change-any-date-references-to-have-to-rather-than)

### Context

The GOV.UK design states that date/time ranges should use `to` instead of a hyphen. Update to use `to` and remove spacing before the meridian indicator to be consistent with the GOV.UK design system examples.

- Make time format consistent with GOV.UK design system


peteryates (approved):
Looks good

PR #2038, opened 2021-11-26
Rework the view teaching event page

peteryates:
Make some amendments based on feedback from the events team

### Context and changes

- Inline definition list and remove details wrapper
- Make training provider events grey in listing
- Move 'In person' above 'Online' in the filter
- Add the 'Do you have a teaching event?' CTA
- Add a presenter for event display
- Begin simplifying the event display

#### Inline the 'event types' section on desktop

https://user-images.githubusercontent.com/128088/143593678-a56557b4-a97f-4b72-8b5a-527a244b1b06.mp4

#### Make the training provider events grey

This leaves the DfE online events yellow so they'll stand out a bit more ✨ 

Screenshot from 2021-11-26 10-57-18

#### Try out using two columns on desktop

Putting the event venue stuff on the right keeps it together without making the page too long on desktop.

Screenshot from 2021-11-26 14-16-07

On mobile it goes beneath the event description

Screenshot from 2021-11-26 14-14-50


app/views/teaching_events/index.html.erb
```diff
@@ -6,23 +6,22 @@
...
-
+      <div>
         <dt>Training provider events</dt>
-
-        <dl>Find out about local courses and how to apply; hosted by teacher training providers.</dl>
-      </dl>
-    </details>
+        <dd>Find out about local courses and how to apply; hosted by teacher training providers.</dd>
+      </div>
+    </dl>
```
peteryates, line 24: @Tomcowan1, @EllieNodder - any thoughts on this text? They just provide an overview of the event types and what's involved:

Screenshot from 2021-11-29 11-21-40

gemmadallmandfe:
Feedback:

- issue with responsiveness in Safari as discussed - body expands beyond width of screen (tested using iPhone 11/iOS 15.1/Safari) even after changing the event type diagonal label

- embedded maps are not rendering - not sure if this is an issue specific to dev though (eg https://review-get-into-teaching-app-2038.london.cloudapps.digital/teaching-events/211202-teesside-university-get-into-teaching)

Otherwise happy for this to be merged to Test (some minor comments that I will add to a separate ticket to be addressed post go-live

app/views/teaching_events/show/_provider-info.html.erb
```diff
@@ -0,0 +1,24 @@
+<% if [
```
ethax-ross, line 1: We need to be careful here not to show provider details for train to teach/question time events, even if they are present (which they are now so that we can display them in the structured data). See #2034 - would it make sense to have this logic in the presenter?

app/views/teaching_events/show.html.erb
```diff
@@ -22,27 +22,19 @@
     </div>
   </header>
 
-  <section class="event-info">
-    <% if @event.message %>
-      <div class="event-alert">
-        <%= tag.p(@event.message) %>
-      </div>
-    <% end %>
+  <div class="container teaching-event__info">
```
ethax-ross, line 25: 💯 this is so much nicer than the current show view

ethax-ross (commented):
Looking really good - I don't think this is displaying correctly in Chrome on my Mac though?

Screenshot 2021-11-30 at 09 12 15

peteryates:
@gemmadallmandfe - I've rebased @ethax-ross's fix (#2050), hopefully when this builds it'll look right on Safari 🤞🏽

gemmadallmandfe:
> @gemmadallmandfe - I've rebased with @ethax-ross's fix (#2050), hopefully when this builds it'll look right on Safari 🤞🏽

Nice work @ethax-ross - no page width issues in Safari now 👍

app/views/teaching_events/show.html.erb
```diff
@@ -22,27 +22,19 @@
...
+    <article>
+      <%= render(partial: "teaching_events/show/event-summary") %>
+      <%= render(partial: "teaching_events/show/attending-training-providers") %>
+      <%= render(partial: "teaching_events/show/scribble") if @event.scribble_id.present? %>
+      <%= render(partial: "teaching_events/show/how-to-attend") if is_a_train_to_teach_event?(@event) %>
+      <%= render(partial: "teaching_events/show/provider-info") if @event.show_provider_information? %>
+    </article>
+
+    <aside>
+      <%= render(partial: "teaching_events/show/venue-information") if @event.building.present? %>
```
ethax-ross, line 35: I'm not sure if the behaviour change here is intentional or not, but we currently don't show venue information for 'virtual events' -- a virtual event is online and can (or always, not sure) have a building:

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/views/events/show.html.erb#L42

app/views/teaching_events/show.html.erb
```diff
@@ -22,27 +22,19 @@
...
-      <div class="event-alert">
-        <%= tag.p(@event.message) %>
-      </div>
-    <% end %>
+  <div class="container teaching-event__info">
+    <article>
+      <%= render(partial: "teaching_events/show/event-summary") %>
+      <%= render(partial: "teaching_events/show/attending-training-providers") %>
+      <%= render(partial: "teaching_events/show/scribble") if @event.scribble_id.present? %>
+      <%= render(partial: "teaching_events/show/how-to-attend") if is_a_train_to_teach_event?(@event) %>
```
ethax-ross, line 30: Are we replicating the other 'how to attend' variants somewhere for non TTT events?

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/views/events/show.html.erb#L84-L92

Also we only offer this currently if the event is open, has a web feed id and is TTT (the 'pending' conditional in there is so that the internal event previews work correctly -- speaking of which, we may need to make a ticket to update that to use the new event view):

https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/views/events/show.html.erb#L66
https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/helpers/events_helper.rb#L36-L38

ethax-ross (approved):

PR #2039, opened 2021-11-26
Trim white space before rendering Markdown placeholders

ethax-ross:
### Trello card

N/A I noticed it while investigating [Trello-2512](https://trello.com/c/S4jAhQGb/2512-table-column-scopes)

### Context

Some of the view components are wrapped in a conditional whereby the HTML content is then indented in the template file, for example:

```
<% if something %>
  <p>text</p>
<% end %>
```

Most of the time when this is injected into Markdown its absolutely fine, however if it happens to be injected immediately after a bullet list the leading spaces result in the component itself being included in the final `<li></li>` content. For example:

```
* one
* two
$component$
```

Results in:

```
<ul>
  <li>one</li>
  <li>two</li>
  <li>[component-html]</li>
</ul>
```

To prevent this, we need to trim the leading white space prior to rendering  components so that they are never treated as a list item.

- Trim white space before rendering Markdown placeholders

### Guidance to review

The reason this happens is because a leading space after a bullet list results in the content also being indented:

```
* one
* two

  three
```

Renders out as:

* one
* two

  three

Where as what we want is:

```
* one 
* two

three
```

Which becomes:

* one
* two

three

| Before      | After |
| ----------- | ----------- |
| Screenshot 2021-11-26 at 15 44 22      | Screenshot 2021-11-26 at 15 44 03       |

### DIFF
--- a/lib/template_handlers/markdown.rb
+++ b/lib/template_handlers/markdown.rb
@@ -61,9 +61,8 @@ module TemplateHandlers
       # use $1 rather than a block argument here because gsub assigns the
       # entire placeholder to the arg (including dollar symbols) but we only
       # want what's inside the capture group
-
       parsed.content.gsub(COMPONENT_PLACEHOLDER_REGEX) do
-        safe_join([cta_component($1), component("quote", $1), image($1)].compact)
+        safe_join([cta_component($1), component("quote", $1), image($1)].compact).strip
       end
     end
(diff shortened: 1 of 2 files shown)


peteryates (approved):
Ah good catch!

peteryates (approved):
Looks good

PR #2042, opened 2021-11-29
Add scope to table headings

ethax-ross:
### Trello card

[Trello-2512](https://trello.com/c/S4jAhQGb/2512-table-column-scopes)

### Context

Markdown table syntax does not support adding a `scope` attribute to table headings. Instead, we can add the `scope` attribute using a small Stimulus controller. If a table already has a scope we won't change it (which lets us write custom HTML tables that may have vertical headings and a `row` scope). If the heading has no text content we won't add a scope (which seems to be best practice).

- Add scope to Markdown table headings

### DIFF
--- a/spec/javascript/controllers/table_controller_spec.js
+++ b/spec/javascript/controllers/table_controller_spec.js
@@ -1,0 +1,58 @@
+import { Application } from 'stimulus';
+import TableController from 'table_controller.js';
+
+describe('TableController', () => {
+
+  beforeAll(() => registerController());
+
+  const setBody = () => {
+    document.body.innerHTML = `
+      <div data-controller="table">
+        <div class="markdown">
+          <table>
+            <thead>
+              <tr>
+                <th class="empty-col"><!-- empty --></th>
+                <th class="col">Col 1</th>
+                <th class="col">Col 2</th>
+                <th class="col-with-scope" scope="row">Col 3</th>
+              </tr>
+            </thead>
+            <tbody>
+              <tr>
+                <td>Cell 1</td>
+                <td>Cell 2</td>
+                <td>Cell 3</td>
+                <td>Cell 4</td>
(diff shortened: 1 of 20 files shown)


spec/javascript/controllers/table_controller_spec.js
```diff
@@ -0,0 +1,58 @@
...
+    application.register('table', TableController);
+  }
+
+  beforeEach(() => {
+    setBody();
+  });
+
+  it('adds a col scope to headings', () => {
+    const cols = [...document.querySelectorAll('.col')]
+    expect(cols.map(c => c.scope)).toEqual(['col', 'col']);
```
jkempster34, line 46: Wondering if this might read better

```js
cols.forEach(col => expect(col.scope).toEqual('col'))
```

spec/javascript/controllers/table_controller_spec.js
```diff
@@ -0,0 +1,58 @@
...
+  })
+
+  it('does not add a scope to empty headings', () => {
+    const emptyCol = document.querySelectorAll('.empty-col')
+    expect(emptyCol.scope).toBeUndefined();
```
jkempster34, line 51: I think this test will always fail? `querySelectorAll` returns a nodeList.

```js
    const emptyCol = document.querySelector('.empty-col')
    expect(emptyCol.scope).toBe(false) // or something, I think it will be an empty string which is false in JS
```

jkempster34 (changes requested):
LGTM, but I think there is a test that will always fail.

spec/javascript/controllers/table_controller_spec.js
ethax-ross: Whoops, yep well spotted

jkempster34 (approved):
💯

PR #2047, opened 2021-11-29
Updating Christmas contact hours on Steps page

EllieNodder:
### Trello card
https://trello.com/c/J6YOhH0r/2680-change-call-centre-opening-hours-for-christmas


peteryates (commented):
Looks fine to me.

Not sure if the `th`/`rd` on the day is required - GOV.UK suggests against using them but we do use them elsewhere.

PR #2052, opened 2021-11-30
Try out using an event type model

peteryates:
### Trello card

https://trello.com/c/jHOXJGrE/2703-clean-up-use-of-constants-file

### Context

There are lots of places where the same text is repeated through the codebase, especailly the event type names. We can remove some of this duplication by creating a model with some useful methods that allow:

* the lookup of an event type given the name/id
* retrieval of ids using a method
* instantiation with an event and the ability to interrogate it

- Add first draft of EventType model
- Switch over to using the new EventType model

### Guidance to review

Is this a good idea? Or is it just moving the pain upstream? 🙈

### DIFF
--- a/app/models/event_type.rb
+++ b/app/models/event_type.rb
@@ -7,7 +7,7 @@ class EventType
     :school_or_university_event_id,
     :train_to_teach_event_id,
     :question_time_event_id,
-    to: :class
+    to: :class,
   )
 
   class << self
@@ -50,7 +50,7 @@ class EventType
   end
 
   def type
-    self.lookup_by_id(type_id)
+    lookup_by_id(type_id)
   end
 
   # this name might be confusing...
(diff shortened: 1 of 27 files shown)


app/models/event_type.rb
```diff
@@ -0,0 +1,76 @@
...
+    @event   = event
+    @type_id = event.type_id
+  end
+
+  def type
+    lookup_by_id(type_id)
+  end
+
+  # this name might be confusing...
+  def online_event?
```
ethax-ross, line 57: I wonder if its just worth calling out that this is the type of the event, so `online_event_type?`... though you'd probably need to suffix the others to be consistent

ethax-ross (commented):
Huge improvement 💯

app/models/event_type.rb
peteryates: Yeah - it crossed my mind. Perhaps now we're referring to them as 'Online Q&A events' everywhere `#online_qa_event?` would work?

ethax-ross (approved):
Looks great

PR #2053, opened 2021-11-30
Ensure @front_matter is always declared

ethax-ross:
### Trello card

[Trello-2702](https://trello.com/c/Xw9I0rwi/2702-declare-frontmatter-everywhere)

### Context

Most but not all pages have associated frontmatter; to avoid `nil` checks we can ensure a `@front_matter` hash is declared in a `before_action` for all use cases.

- Ensure @front_matter is always declared


peteryates (approved):
Definitely an improvement

PR #2068, opened 2021-12-02
Remove legacy tracking/analytics

ethax-ross:
### Trello card

[Trello-2720](https://trello.com/c/gAYNJ8s1/2720-remove-legacy-gtm-tracking-code-from-the-app-tta)

### Context

All of our tracking/analytics is now served via Google Tag Manager, which means we can remove a bunch of now redundant code from the app - it is all currently disabled in production anyway.

Remove all references to:

- LID
- BAM
- Legacy GTM container
- Google Analytics
- Hotjar
- Twitter
- Facebook
- Pinterest

- Remove legacy tracking/analytics


peteryates (approved):
Massive improvement

🗑️

gemmadallmandfe (approved):
Happy for this to be merged given that this is already disabled in production and that it has had a tech review!

PR #2076, opened 2021-12-06
Retry failed external link checks

ethax-ross:
### Trello card

[Trello-2733](https://trello.com/c/Gda476AL/2733-investigate-internal-link-checker-failures)

### Context

We are seeing flakiness in the external link checker where it reports a broken link and when we check it manually straight after, or perform a re-run, the link check passes. To try and limit the impact of 'blips' with external websites we can retry the external links multiple times. The order is also shuffled which may help weed out flakiness.

- Retry failed external link checks

### Guidance to review

I thought about putting the retry logic within the `LinkChecker` class, but it would be difficult to work in a delay that way and the natural delay we get in doing it this way (given each run takes several minutes) will hopefully help make it more robust.

### DIFF
--- a/spec/features/external_links_spec.rb
+++ b/spec/features/external_links_spec.rb
@@ -3,24 +3,14 @@ require "rails_helper"
 RSpec.feature "link checks", :onschedule do
   describe "all pages (external)" do
     specify "all linked-to pages exist" do
-      results = LinkChecker::Results.new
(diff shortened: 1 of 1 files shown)


peteryates (approved):
Looks like a good step - less drastic than firing up a Redis or something to store results!

PR #2077, opened 2021-12-06
Amending name and contact details

Tomcowan1:
Change to details requested by Rick in the Market Regulation Team. Original request came from the Pilgrim Partnership.

### DIFF
--- a/app/views/content/assessment-only-providers.md
+++ b/app/views/content/assessment-only-providers.md
@@ -50,10 +50,10 @@ provider_groups:
       link: http://www.nett.org.uk/
       telephone: '01255431949'
       email: teach@nett.org.uk
-    - header: SAF ITT
-      name: Su Bernard
-      telephone: 01234 827145
-      email: SBernard@saf.org.uk
+    - header: The Pilgrim Partnership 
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
Change looks fine to me @Tomcowan1 👍 

However, the build has failed one check (not sure why) so haven't approved - one to check with @peteryates !

NB might be worth asking the Accreditation team whether the provider wants their website adding which I presume is https://pilgrimpartnership.org/ - but that can be done later!

Tomcowan1:
Thanks @gemmadallmandfe have messages Pete about the fail - will pick up tomorrow. Good point about the web address.

peteryates:
This is just down to a flaky build @Tomcowan1 -  retried it and it seemed to work!

Tomcowan1:
Thanks @peteryates I think this is good to go when you're ready @gemmadallmandfe 

I've spoken to Support who will ask if there's a web address. If so, I'll add in a separate PR.

gemmadallmandfe (approved):
Happy for this to be merged

If/when you add the website address @Tomcowan1 we should sort out the alphabetical ordering!

PR #2085, opened 2021-12-07
Fix Google Optimize redirect experiments

ethax-ross:
### Trello card

[Trello-2685](https://trello.com/c/sNHiFAcU/2685-setup-a-b-test-for-events-in-our-test-environment)

### Context

Google Optimize will drop a cookie for 5 seconds after performing a redirect; if that cookie is present it will not perform another redirect. The idea is that this prevents redirect loops happening if you incorrectly configure an experiment. The issue with it is that it means a user can be redirected to a variant, then click the link again and end up seeing the control if they do it quickly.

It doesn't feel great manually clearing this, but we're unlikely to be running lots of experiments (which is where a redirect loop becomes likely) and this ensures that users always see the correct variant every time.

- Fix Google Optimize redirect experiments

### Guidance to review

Also adds test coverage to the cookie clearance logic added in an earlier PR.

### DIFF
--- a/app/webpacker/javascript/cookie_preferences.js
+++ b/app/webpacker/javascript/cookie_preferences.js
@@ -30,6 +30,19 @@ export default class CookiePreferences {
     );
   }
 
+  static get cookieDomains() {
+    const hostname = window.location.hostname;
+    const rootDomain = hostname.replace(/(.*?)\./, '');
+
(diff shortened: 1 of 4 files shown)


peteryates (approved):
Looks good, had a good click around and it seems like it's working correctly

PR #2092, opened 2021-12-08
Fix accessibility issues with autocomplete input

ethax-ross:
### Trello card

[Trello-2506](https://trello.com/c/Wtt0YC5h/2506-take-a-closer-look-at-search-accessibility)

### Context

We see a warning on Silktide that the input has no label. To fix this we're going to try adding an aria-label to the
auto-complete input and also co-locate the label next to the input. One or both of these may not be necessary, but we
won't know until the next Silktide report.

- Fix accessibility issues with autocomplete input

### Guidance to review

Screenshot 2021-12-08 at 14 46 01

### DIFF
--- a/app/components/header/extra_navigation_component.html.erb
+++ b/app/components/header/extra_navigation_component.html.erb
@@ -7,7 +7,7 @@
     </li>
 
     <%= tag.li(class: %w[extra-navigation__link extra-navigation__search], data: { controller: "searchbox", "searchbox-search-input-id-value" => search_input_id }) do %>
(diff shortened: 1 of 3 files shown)


peteryates (approved):
Makes sense, will be interesting to see what Silktide thinks.

The next step would probably be to wrap the input in the label altogether:

```html
<label>
  Search
  <input name="search", type="search"/>
</label>
```

This is valid as-is but I don't know if the search widget would affect that when it kicks in.

PR #2093, opened 2021-12-09
Add autocomplete attribute to text input fields

ethax-ross:
### Trello card

[Trello-2751](https://trello.com/c/e1rgnH7J/2751-accessibility-issue-identify-the-purpose-of-fields-programmatically)

### Context

Silktide has thrown up a warning around our text input fields not being accessible due to lacking `autocomplete` attributes. Chrome does a decent job in my experience of inferring it from the input names, but it doesn't hurt to be explicit.

Add `autocomplete` attributes to all text input fields. Disable autocomplete for the verification code field and use `on` when a specific value doesn't make sense (i.e. for the search input).

- Add autocomplete attribute to text input fields


peteryates (approved):
Looks great

PR #2095, opened 2021-12-10
Track cookie opt-in success rate

ethax-ross:
### Trello card

[Trello-2640](https://trello.com/c/PMZfAK9g/2640-create-grafana-dashboard-for-cookie-opt-ins-opt-outs)

### Context

We want to be able to track some metrics that trigger from the client-side. As all our metrics go via Prometheus we will need to send the information to Rails from the client in order to track it.

- Add ability to track client-side metrics

Add `ClientMetricsController` to accept incoming client metric information from Javascript.

Add `app_client_cookie_consent_total` metric as the first client-side metric to track the opt-in success rate.

- Update interface to accept setting multiple categories

We want to be able to emit an event when cookies change; currently we use a `setCategory` method that is called multiple times, so the `CookiePreference` class doesn't know when all the cookies have finished changing. By changing this to `setCategories` we encourage callers to set all cookie categories in one go and can emit an event thereafter.

- Send metric when opt-in status changes

When a user updates their opt-in status we want to increment a metric so that we can track the success rate of opting-in to cookies.

When a user changes their cookie preferences (either manually or via 'allow all') emit a `app_client_cookie_consent_total` metric with the respective labels.

### Guidance to review

In an ideal world we would want to track opt-in rate per user, so that if a user opts-out but subsequently opts-in we have their latest status reflected in the metric. Given we are tracking even if they don't accept cookies we have to make sure its completely anonymous, so the best we can do is track the various opt-in/out events to give a decent indication of what portion of users are allowing the different cookie categories.

### DIFF
--- a/app/controllers/client_metrics_controller.rb
+++ b/app/controllers/client_metrics_controller.rb
@@ -1,0 +1,11 @@
+class ClientMetricsController < ApplicationController
+  skip_before_action :verify_authenticity_token
+
+  def create
+    metric_json = JSON.parse(request.body.read)
+
+    ActiveSupport::Notifications.instrument("app.client_metric", metric_json)
+
+    head :ok
+  end
+end
--- a/app/webpacker/controllers/cookie_preferences_controller.js
+++ b/app/webpacker/controllers/cookie_preferences_controller.js
@@ -29,16 +29,18 @@ export default class extends Controller {
   save(event) {
     event.preventDefault();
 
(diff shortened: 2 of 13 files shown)


peteryates:
Looks good - works fine in the test env. If users keep submitting the form their response will be recorded several times - not sure that's easily avoidable though.

peteryates (approved):

PR #2098, opened 2021-12-13
No index event sign up page

ethax-ross:
### Trello card

[Trello-2763](https://trello.com/c/mz5aAzEr/2763-resolve-google-search-console-coverage-issue-for-apply-pages)

### Context

We don't want users to land directly on this page; instead they should go through the view event page.

Google Search Console also shows a redirect error here, as we redirect back to the event page if the event is closed -- if the event also happens to be in the past we will then return a 404 and I think Google has an issue with redirecting to a 404 page.

- No index event sign up page

Remove header method of no-indexing a page given that we specify this in the HTML now (this should have been removed already but I must have missed it -- noindex via header doesn't work all the time because of our static page cache).

### Guidance to review

As part of this I looked at the sitemap and I'm wondering if we would benefit from explicitly surfacing our individual event pages in there (as we don't currently)?

### DIFF
--- a/app/controllers/event_steps_controller.rb
+++ b/app/controllers/event_steps_controller.rb
@@ -7,6 +7,7 @@ class EventStepsController < ApplicationController
   include DFEWizard::Controller
   self.wizard_class = Events::Wizard
 
+  before_action :noindex
   before_action :redirect_closed_events, only: %i[show update]
   before_action :set_step_page_title, only: [:show]
   before_action :set_completed_page_title, only: [:completed]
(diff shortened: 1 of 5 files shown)


peteryates (approved):

gemmadallmandfe (approved):
Looks good to me

I have tested the following:

- **TTT event** noindex appears on every step of sign-up but **not** on event details page
- **online Q&A**  - does not contain noindex
- **provider event** - does not contain noindex

Happy for this to be merged. I will then ask GSC to revalidate the errors.

PR #2100, opened 2021-12-13
edited some wording in AO intro and link text

tcannon3:
Changed wording around experienced teacher and also shortened providers offering assessment only QTS to international teachers

### Trello card
https://trello.com/c/EPcL36d6/1859-revisit-the-content-at-the-top-of-the-assessment-only-page

### DIFF
--- a/app/views/content/assessment-only-providers.md
+++ b/app/views/content/assessment-only-providers.md
@@ -510,6 +510,6 @@ Contact them directly to ask them about entry criteria and to apply.
 
 ### Teachers from outside the UK
 
-If you have experience of teaching outside the UK, you can apply to certain training providers for an assessment only QTS without visiting or training in England. 
+If you have experience of teaching outside the UK, some training providers will let you apply for an assessment only QTS without visiting or training in England.
 
-[View providers offering assessment only QTS to international teachers.](#group--international)
+[See the training providers who offer assessment only QTS to international teachers.](#group--international)
--- a/app/views/content/assessment-only-providers.md
+++ b/app/views/content/assessment-only-providers.md
@@ -512,4 +512,4 @@ Contact them directly to ask them about entry criteria and to apply.
 
 If you trained as a teacher outside the UK, you can apply to certain training providers for an assessment only QTS without visiting or training in England. 
 
-[View teaching training providers offering assessment only QTS to international teachers.](#group--providers-offering-assessment-only-qts-to-international-teachers)
+[View teaching training providers offering assessment only QTS to international teachers.](#group--international)
--- a/app/views/content/assessment-only-providers.md
+++ b/app/views/content/assessment-only-providers.md
@@ -471,7 +471,7 @@ provider_groups:
       name: Mikki Burns
       telephone: 0203 194 3200
       email: institute@tesglobal.com
-  Providers offering assessment only QTS to international teachers:
+  International:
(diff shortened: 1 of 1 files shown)


app/views/content/assessment-only-providers.md
```diff
@@ -502,14 +502,14 @@ keywords:
...
+If you have a degree and enough classroom experience, you may be able to get qualified teacher status (QTS) through a 12-week assessment only (AO) programme.
 
 The accredited teacher training providers listed on this page offer the AO programme.
 
 Contact them directly to ask them about entry criteria and to apply.
 
 ### Teachers from outside the UK
 
-If you trained as a teacher outside the UK, you can apply to certain training providers for an assessment only QTS without visiting or training in England. 
+If you have experience of teaching outside the UK, you can apply to certain training providers for an assessment only QTS without visiting or training in England.
```
gemmadallmandfe, line 513: Is this definitely right @tcannon3 - that you only need experience, not training, as an international AO applicant?

(That might well be the case - just wanted to check!)

app/views/content/assessment-only-providers.md
```diff
@@ -502,14 +502,14 @@ keywords:
   - experienced
 ---
 
-If you’re an experienced teacher with a degree, you may be able to get qualified teacher status (QTS) through a 12-week assessment only (AO) programme.
+If you have a degree and enough classroom experience, you may be able to get qualified teacher status (QTS) through a 12-week assessment only (AO) programme.
```
gemmadallmandfe, line 505: Is this definitely right @tcannon3 - that we want to change from 'experienced' to 'enough classroom experience'?

(Might be the case - just wanted to check that it is accurate!)

gemmadallmandfe:
I think the heading needs changing as well @tcannon3 - 'international' doesn't make sense as 'a region of the provider'?

Perhaps 'Select a region' would be better?

image

app/views/content/assessment-only-providers.md
tcannon3: Yes. According to UCAS and providers, the AO route is for those with substantial classroom experience. The current wording around experienced teachers makes it seem like this is a route for those who are already teachers in which case they would also already have QTS. 
Happy to revise the wording around this though if needed :)

app/views/content/assessment-only-providers.md
tcannon3: I read on TES that international AO applicants can be any member of school staff -unqualified teacher, TA or higher-level teaching assistant with significant teaching experience. Basically it's someone who can prove that they meet all of the Teaching Standards after having significant classroom experience in a British Overseas school

tcannon3: Having said that, I've just double checked their entry requirements and it seems that they want to international applicant to be employed in a teaching role within a school. But that doesn't necessarily mean trained teacher as the candidate could be an unqualified teacher

tcannon3: @gemmadallmandfe Phoebe has given the OK for these changes

app/views/content/assessment-only-providers.md
tcannon3: @gemmadallmandfe Phoebe has OK'd these changes also

Tomcowan1 (approved):

gemmadallmandfe (approved):
Looks good to me

PR #2104, opened 2021-12-14
Refactor event status logic into new model

ethax-ross:
### Trello card

[Trello-2703](https://trello.com/c/jHOXJGrE/2703-clean-up-use-of-constants-file)

### Context

Remove the `GetIntoTeachingApiClient::Constants::EVENT_STATUS` constant, shifting the logic into an `EventSatus` class.

I've also made this class responsible for determining view-ability and whether registration is open as it feels like this should be centralised and comes under the umbrella of 'status'.

- Refactor event status logic into new model

### Guidance to review

I've kept this consistent with the `EventType` model.

### DIFF
--- a/app/controllers/event_steps_controller.rb
+++ b/app/controllers/event_steps_controller.rb
@@ -27,7 +27,7 @@ private
   end
 
   def redirect_closed_events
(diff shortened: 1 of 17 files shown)


peteryates (approved):
**Massive** improvement.

PR #2105, opened 2021-12-15
Refactor teaching subjects into new model

ethax-ross:
### Trello card

[Trello-2704](https://trello.com/c/bO8W7Dn9/2704-clean-up-welcome-guide-references-to-subject-uuids)

### Context

Refactor teaching subjects into new model Remove the `GetIntoTeachingApiClient::Constants::TEACHING_SUBJECTS` and
`IGNORED_PREFERRED_TEACHING_SUBJECTS` constants, shifting the logic into an `TeachingSubject` class.

Remove all references to hard-coded UUIDs with calls to the `TeachingSubject` class.

- Refactor teaching subjects into new model

### Guidance to review

Along the same lines as the `EventType` and `EventStatus` models.

### DIFF
--- a/app/helpers/welcome_content_helper.rb
+++ b/app/helpers/welcome_content_helper.rb
@@ -152,20 +152,22 @@ module WelcomeContentHelper
 private
 
   def mappings(id)
+    key = TeachingSubject.key_with_uuid(id)
+
     {
-      "a42655a1-2afa-e811-a981-000d3a276620" => MATHS,
-      "942655a1-2afa-e811-a981-000d3a276620" => ENGLISH,
+      maths: MATHS,
+      english: ENGLISH,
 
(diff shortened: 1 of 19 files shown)


peteryates (approved):
Massive improvement.

I wonder if later we could put the `:maths` style keys in `TeachingSubject` too so we can use them in welcome guide URLs

ethax-ross:
> Massive improvement.
> 
> I wonder if later we could put the `:maths` style keys in `TeachingSubject` too so we can use them in welcome guide URLs

Ah yeah that's a good shout I forgot we use the guids in the URLs, I did a find/replace on all of them hard coded but switching them out there would be really nice. I'll maybe try and do that this afternoon if I get a minute.

PR #2106, opened 2021-12-15
Fix broken links in funding widget

gemmadallmandfe:
### Trello card
https://trello.com/c/Odqjb1Of

### Context
Three of the links in the funding widget don't work and need replacing - they are anchor links to content further down the page.

In addition, 'T&Cs' should be 'terms and conditions' for readability.


peteryates (approved):
Looks good, links all work

PR #2107, opened 2021-12-15
Refactor remaining constants into OptionSet class

ethax-ross:
### Trello card

[Trello-2703](https://trello.com/c/jHOXJGrE/2703-clean-up-use-of-constants-file)

### Context

We have a number of constants remaining in `GetIntoTeachingApiClient::Constants` that we want to restructure to make referencing them easier.

Pull out the constants into a new `OptionSet` class and add convenience methods. I've lumped all these into one class to keep it concise and extensible.

- Refactor remaining constants into OptionSet class

### Guidance to review

Arguably the hashes could be simplified with keys that map to more concise symbols, but it feels worth keeping this matching the CRM values in case we inadvertently expose them to the user. The only exception to this is the mailing list channels which I've renamed slightly as they will never be user-facing.


peteryates (approved):
Nice 👍🏽

PR #2113, opened 2021-12-17
Show chat as offline on weekends

ethax-ross:
### Trello card

[Trello-2779](https://trello.com/c/fblVO89g/2779-ensure-chat-button-is-unavailable-at-weekends)

### Context

We currently show the chat as offline when outside of working hours, however we aren't taking into account weekends when the chat should be offline all the time.

- Show chat as offline on weekends


peteryates (approved):
Looks good

PR #2118, opened 2021-12-20
Change email listings text to mailto links

peteryates:
Adding this PR as an optional one - it introduces [some usability problems on desktop](https://adamsilver.io/blog/the-trouble-with-mailto-email-links-and-what-to-do-instead/) but it's a widely used pattern so people are likely to understand its shortfalls already.


gemmadallmandfe:
@peteryates looks OK to me - my only slight concern is where we have multiple email addresses in that field (see Smile and Educationline entries). However, because they are separated by semi-colons, it does seem to kind of work!

gemmadallmandfe (approved):
Looks good to me

PR #2119, opened 2021-12-20
Update to v2 of the API client

ethax-ross:
### Trello card

[Trello-2777](https://trello.com/c/oYLSzOTp/2777-monkey-patch-api-client-models-to-support-snakecase-attributes-in-initializer?filter=member:rossoliver15)

### Context

The API client v2 uses `snake_case` attributes in the initializers instead of `camelCase`; this updates all the references.

The initializer now also checks for attributes that don't exist on the model, raising an exception if any are provided. As some of our steps have additional attributes to the payload we need to strip these out prior to initializing the request models.

- Update to v2 of the API client


peteryates (approved):
Much better

PR #2125, opened 2021-12-21
Remove Welsh entry from the list

peteryates:
It crept back in via the master sheet but we should be limiting to just England for the time being.


gemmadallmandfe (approved):
Looks good to me

PR #2126, opened 2021-12-21
Same link text going to multiple destinations

gemmadallmandfe:
### Trello card
https://trello.com/c/A1JXUPQh

### Context
Silktide has highlighted an issue whereby the same link text is used for multiple links on a page that point to different destinations. This makes it difficult for people using screen readers to know which link is which.

In addition, we also have one instance where different link text is used to point to the same page, which again is not good from an accessibility perspective.


gemmadallmandfe:
@peteryates The `/research` page button text changes are particularly clunky - suspect I need to shorten them further but suggestions welcome!

peteryates (approved):
Huge improvement

PR #2127, opened 2021-12-22
Remove and reorganise photos

peteryates:
Lots of the removed ones used to be used in the old hero and CTAs; those components have since been replaced along with their images.

Also try to restructure the content directory a bit by continuing the process of grouping related image by directory.

There should be only one visual difference, I opted to pick a less awkward crop of the homepage CTA, now the subject is not looking out of frame and there's no partial face cut off on the left.

| Before | After |
| ------ | ------ |
| Screenshot from 2021-12-22 10-14-58 | Screenshot from 2021-12-22 10-15-08 |


ethax-ross (approved):
Nice 👍

gemmadallmandfe:
Happy for this to be merged on 4 Jan - will approve then.

Two questions though in the meantime:

* the home page CTA photo has been flipped - which is the correct way round for the photo? I'm a bit uncomfortable with flipping photos of people in general even if it looks better aesthetically

* presume you have checked that there aren't any photos now missing on pages?

peteryates:
@gemmadallmandfe - I'm not sure which the original was, there were two variants in our images folder and we've used both at some point! Happy to switch this out for another photo, we did have another that looks 'advisery' ([this one](https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/webpacker/images/content/hero-images/0015.jpg)) but then we'd have three male photos on the page so left it to balance it out.

I checked for the filename as I deleted each file, they don't appear in the codebase. Additionally when there's an asset missing the app will throw a 404 for the entire page, our tests should have caught them. I'm confident we're using

gemmadallmandfe:
@peteryates I had a look at the image and there is a clock which implies the image that is correct is the one where she is facing to her left :-)

image

However, let's leave as is with your new crop for now as we are using this flipped version on the events page already anyway

When we get some new images, we'll make sure that we are using them the right way round.

peteryates:
Ah well spotted. Also the map on the other side, South America is backwards! 🤦🏽‍♂️

gemmadallmandfe (approved):
Happy for you to merge this @peteryates

PR #2129, opened 2021-12-22
Import latest batch of agency updates

peteryates:


gemmadallmandfe:
Changes all look fine except Wales region has been re-introduced @peteryates ! I removed it from the Google Sheet - it will need removing from the Excel file if you are using that

gemmadallmandfe (approved):
Looks good to me

PR #2157, opened 2022-01-06
Make welcome guide use path instead of query

peteryates:
### Trello card

https://trello.com/c/k6j7B40B/2806-investigate-workarounds-for-crm-welcome-guide-email-problems

### Context and changes

We ran into a problem where Dynamics' link tracking appends `?p0=` to the links it generates making the subsquent query param be ignored. Here's an example query:

```
?p0=?degree_status_id=222750001&preferred_teaching_subject_id=962655a1-2afa-e811-a981-000d3a276620
    ^
    |
    | from here is ignored
```

To combat it, this PR changes the format, we go from this:

```
/welcome?degree_status_id=222750001&preferred_teaching_subject_id=962655a1-2afa-e811-a981-000d3a276620
```

To this:

```
/welcome/email/subject/maths/degree-status/graduate
```

Two things have changed:

* we're using the path instead of the query string (see the new routes)
* we're referring to records using a slug instead of a ID or GUID. This makes it easier and gives the user a better experience

In this app we convert the string name to 'slug' by lowercasing and replacing spaces with underscores. The CRM doesn't know about them yet so ideally we'd add this data (and hard code it) in the CRM and then use that as the source.

### DIFF
--- a/app/models/option_set.rb
+++ b/app/models/option_set.rb
@@ -32,6 +32,14 @@ class OptionSet
       lookup_const(category).fetch_values(*keys)
     end
 
+    def lookup_by_value(category, value)
+      lookup_const(category).invert.fetch(value)
+    end
+
+    def lookup_by_values(category, *values)
+      lookup_const(category).invert.fetch_values(*values)
+    end
+
     def lookup_const(category)
       const = const_get(category.to_s.pluralize.upcase)
       const.transform_keys { |k| k.parameterize(separator: "_").to_sym }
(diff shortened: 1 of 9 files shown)


app/models/option_set.rb
```diff
@@ -32,6 +32,14 @@ def lookup_by_keys(category, *keys)
       lookup_const(category).fetch_values(*keys)
     end
 
+    def lookup_by_value(category, value)
```
ethax-ross, line 35: Missing specs?

peteryates: Ah yeah - thanks. Added them

ethax-ross (approved):
Looks good to me 👍

PR #2158, opened 2022-01-07
Use formatted_event_description helper

peteryates:
This makes the new template match the behaviour of the old one more closely. Some events with Word-style formatting behaved a bit differently in the test environment.

### DIFF
--- a/app/views/teaching_events/show/_event-summary.html.erb
+++ b/app/views/teaching_events/show/_event-summary.html.erb
@@ -6,7 +6,7 @@
   </div>
 <% end %>
 
-<%= tag.div(safe_html_format(@event.description)) %>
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
@peteryates I'm finding this one a bit hard to test as the events with really poor formatting in Test don't seem to be present in the Review app. Is this because they are pointing to two different CRM environments?

Having said that, given that this doesn't affect production, I'm happy to approve and merge if you give the go ahead.

peteryates:
Yeah it's going to be subtle at the best of times, just uses a slightly different way of rendering the description (the same one as the main events page) that fixes MS Word formatting.

gemmadallmandfe (approved):
Happy to approve so this can be tested properly in Test

PR #2159, opened 2022-01-07
Make event location text line up with other text

peteryates:
### Trello card

https://trello.com/c/dwXgoOLp/2813-add-location-icon-to-summary-card-is-on-new-events-listing-page

### Context

Originally we were going to have a (map pin) icon here on the left but the aspect ratios of the icons make it painful, we're best holding off until we have standard ones that can be used interchangeably.

Make the location text line up with the text above and below.

| Before | After |
| ---- | ------ |
| Screenshot from 2022-01-07 10-43-30 | Screenshot from 2022-01-07 10-43-48 |


gemmadallmandfe (approved):
Looks good to me

PR #2164, opened 2022-01-07
Centralised store for alt text

peteryates:
### Trello card

https://trello.com/c/QBpkcj5p/2615-investigate-whether-alt-text-can-be-defined-at-an-image-level-rather-than-set-on-each-page

### Context

This is the first step towards centralising alt text and how we deal with it. The idea is to have a single 'index' of images with alt text and throw an error if we don't have it (so we find out sooner rather than later!)

The two methods on Image are most important:

* `#build_args` returns a splattable array of arguments that can be passed directly into `#image_pack_tag`
* `#alt` returns just the alt text for a given image path

- Add Ellie's alt text keyed by file path
- Add Image model and ImageHelper for structured alt

### Structure

We store some images in multiple sizes so they can be served to visitors
at a reasonable size/quality depending on their device.

These variants are the same image so they should share alt text. The
structure in images.yml is now as follows:

```yaml
"image/path":
   alt: "a nice picture"
   variants:
     - "other/image/path"
     - "yet/other/image/path"
```

When loaded these will be transformed into:

```yaml
"image/path": "alt text"
"other/image/path": "alt text"
"yet/other/image/path": "alt text"
```

And they can be returned by:

```ruby
Image.new.build_args("image/path")
=>  ["image/path", { alt: "a nice picture" }]

Image.new.alt("image/path")
=> "a nice picture"
```


spec/helpers/image_helper_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      allow(Image).to receive(:new).and_return(image_double)
+      image_args(path)
+    end
+
+    specify "passes the path to Image#build_args" do
+      expect(image_double).to have_received(:build_args).with(path).once
+    end
+  end
+
+  describe "#alt" do
```
ethax-ross, line 18: It feels a bit odd that the expectation isn't on the args/alt text for these tests - this reads better in my mind but I don't feel strongly about it (also noticed a typo in the `#alt` describe, should be `#image_alt`):

```
let(:path) { "a/b/c" }
let(:image_double) { instance_double(Image, build_args: { some: "args" }, alt: "xyz") }

before { allow(Image).to receive(:new).and_return(image_double) }

describe "#image_alt" do
  it { expect(image_alt).to eq(image_double.alt) }
end

describe "#image_args" do
  it { expect(image_args(path)).to eq(image_double.alt) }
end
```

spec/models/image_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        specify "returns the correct image path and alt text" do
+          expect(subject).to eql([path, { alt: "Male and female teacher talking in a staff room." }])
+        end
+      end
+
+      context "with a bad image path" do
+        let(:path) { "media/something-really-invalid/abc.pdf" }
+
+        specify "raises an error" do
+          expect { subject }.to raise_error(Image::UnregisteredImageError, /no image information found/)
```
ethax-ross, line 30: Worth having more of an integration test somewhere that iterates over each image and ensures it has alt text specified?

ethax-ross (dismissed):
Looks great, just a couple of suggestions. I did wonder if it would be worth automating the lookup of variant images as the format is standardised (I think), but I reckon how you have it is going to be simpler - certainly at the outset anyway.

spec/models/image_spec.rb
peteryates: These will be caught by the content specs (and/or request specs); if any page is generated using `#image_args` (just blog and hero for the mo) and it fails to find an entry it'll fail with:

```
ActionView::Template::Error:
  no image information found for 'media/images/content/hero-images/0003.jpg', see config/images.yml
```

ethax-ross: Ah ok cool, I guess its a bit fallible if someone includes an image without using the `Image` class though 🤔

ethax-ross (approved):
Looks good to me ✅

spec/models/image_spec.rb
peteryates: Good call, added another commit with a content spec that ensures there's an `alt` attribute on every `img` element.

ethax-ross (approved):

gemmadallmandfe:
@EllieNodder Do you want to have a look at this in the review app (!) to see that it makes sense with your alt text changes?

peteryates:
The review app _shouldn't_ look any different to how it did before.

The bit to be aware of is the new 'central store' of alt text: https://github.com/DFE-Digital/get-into-teaching-app/pull/2164/files#diff-db1e40c240b0c34a95d8becb8cb7fdbc81ac87e816562bb5bda6858af733b6b5

PR #2169, opened 2022-01-11
Enable the /teaching-events route

peteryates:
This will enable the new event list and view pages. The new path, `/teaching-events`, will not be indexed just yet as it's going to be included in a round of A/B tests first.


gemmadallmandfe (approved):
Happy to approve

PR #2177, opened 2022-01-12
Update mailing list sign up postcode step

ethax-ross:
> :warning: I'll need to update the frontend tests after this ships to account for the new step structure.

### Trello card

[Trello-2732](https://trello.com/c/eMl4jPKM/2732-improve-copy-in-mailing-list-funnel-to-encourage-people-to-give-their-postcode?filter=member:rossoliver15)

### Context

The mailing list currently gives users an option to enter their postcode, however the number of users opting to do this could be better. In an attempt to improve the uptake in this option we are making the copy and form clearer.

- Add send_event_emails attribute to postcode step

We want to show a yes/no toggle on the postcode step, so we only show the text input when the user selects 'yes'.

Add necessary attribute to populate the form with.

Change postcode presence to be conditional based on if they select 'yes'.

- Pull in radio functionality from govuk-frontend

Pull in the radio button functionality from govuk-frontend in addition to the error summaries. We only do s will allow the toggle to show field JS to work. We only pull this into the registration pages; update the filename to be more generic.

Move JS to pull in after the footer instead of the head tag; previously if you submitted the form, went back (via Turbolinks page change) and toggled the radio it would not show the field as the JS was not re-initialising.

- Update postcode step to include radio buttons

Include yes/no radio buttons; when 'yes' is selected we show the postcode field.

### Guidance to review

|  Postcode Step |  Enter Postcode State |  Select Radio Error | Enter Postcode Error  |
|---|---|---|---|
|  Screenshot 2022-01-12 at 10 17 29 |  Screenshot 2022-01-12 at 10 17 48  |  Screenshot 2022-01-12 at 10 20 50  |  Screenshot 2022-01-12 at 10 21 07  |

### DIFF
--- a/app/views/mailing_list/steps/_postcode.html.erb
+++ b/app/views/mailing_list/steps/_postcode.html.erb
@@ -1,9 +1,14 @@
+<h1>Would you like to hear about teacher training events?</h1>
+
 <%= f.govuk_error_summary %>
 
-<%= f.govuk_text_field :address_postcode, autocomplete: "postal-code", label: { tag: 'h1', size: "xl", text: t('helpers.label.mailing_list_steps_postcode.address_postcode') } do %>
-
 <p>
-  Stay up to date about events in your area.
+  We'll let you know via email about upcoming teacher training events near you
 </p>
 
+<%= f.govuk_radio_buttons_fieldset(:send_event_emails, legend: nil) do %>
+  <%= f.govuk_radio_button :send_event_emails, "true", link_errors: true do %>
+    <%= f.govuk_text_field :address_postcode, autocomplete: "postal-code", label: { text: t('helpers.label.mailing_list_steps_postcode.address_postcode') } %>
+  <% end %>
+  <%= f.govuk_radio_button :send_event_emails, "false" %>
 <% end %>
--- a/app/views/mailing_list/steps/_postcode.html.erb
+++ b/app/views/mailing_list/steps/_postcode.html.erb
@@ -1,4 +1,4 @@
-<h1>Would you like to hear about teacher training events?</h1>
+<h1>Would you like to hear about teacher training events in your area?</h1>
 
 <%= f.govuk_error_summary %>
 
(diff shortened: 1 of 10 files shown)


SaimaBodiat (dismissed):
looks good to me

gemmadallmandfe:
@SaimaBodiat Worth showing Gary before this merged, from an interaction perspective?

SaimaBodiat:
Hi @gazcarr, would you mind having a look at this from an interaction perspective, before it's merged? thank you

app/views/mailing_list/steps/_postcode.html.erb
```diff
@@ -1,9 +1,14 @@
-<%= f.govuk_error_summary %>
+<h1>Would you like to hear about teacher training events?</h1>
```
gemmadallmandfe, line 1: I think this needs amending slightly - everyone who signs up to the mailing list gets told about events

The key here is that users are being asked if they want to hear about *local* events by supplying their postcode

The h1 probably needs amending slightly to reflect this
eg '... in your area'

gazcarr:
My only comment is around the length of the text inputs, they should ideally match the size of the content, so for example the postcode input should be roughly 6 characters long (maybe 7 to allow for a space). 

This is the same for the subjects dropdown, we could easily half that.

gemmadallmandfe:
> My only comment is around the length of the text inputs, they should ideally match the size of the content, so for example the postcode input should be roughly 6 characters long (maybe 7 to allow for a space).
> 
> This is the same for the subjects dropdown, we could easily half that.

NB I think postcodes can be (at least) eight characters including space (eg "AB12 3CD")

Suggest we look at the subject dropdown separately @gazcarr (don't disagree!)

ethax-ross:
@gazcarr @gemmadallmandfe I've updated the heading copy and restricted the width of the postcode field to 10 characters (the govuk design system only supports certain increments and 10 is the closes to 8).

I'll leave making the subject select narrower to another PR as its not trivial; we need to restrict the width of the overall form to do that and the layout is shared with the events sign up form, which requires the full length layout as it has a sidebar.

gemmadallmandfe:
Looks good to me @SaimaBodiat - happy for you to approve if you are happy with it!

SaimaBodiat (approved):
looks good to me 👍

PR #2185, opened 2022-01-13
Only apply the abbr substitution in paragraphs

peteryates:
### Trello card

https://trello.com/c/BA9vQXqT/2841-restrict-acronym-expander-to-body-copy-only

### Context

Any potential abbreviation with a parent element other than `<p>` will be skipped.

### DIFF
--- a/lib/acronyms.rb
+++ b/lib/acronyms.rb
@@ -30,7 +30,7 @@ private
   def replace_acronyms(document)
     document.traverse do |node|
       next unless node.text?
-      next if node.parent.name == "a"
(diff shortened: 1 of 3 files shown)


gemmadallmandfe:
@peteryates I've checked the following examples and they look good to me:

- Ways to train flags
- Blue h2 headings (eg on /what-is-a-pgce)

I won't approve this ticket in case you want to get tech eyes on this though - but I'm happy for it to be merged when it's had that tech review

ethax-ross (approved):

PR #2190, opened 2022-01-14
Make deployment workflow a reusuable action

sfawcett123:
Github now offers re-usable actions, this replaces benc-uk/workflow-dispatch which is full of issues, it will also allow the removal of a few steps, which should marginally increase performance, but will definitely reduce false negatives

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -270,6 +270,9 @@ jobs:
     needs: [  build  ]
     if: github.ref != 'refs/heads/master'
     runs-on: ubuntu-latest
+    concurrency: Review_${{github.event.inputs.pr}}
+    environment:
+       name: Review 
     steps:
       - name: Check out the repo
         uses: actions/checkout@v2.4.0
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -270,6 +270,9 @@ jobs:
     needs: [  build  ]
     if: github.ref != 'refs/heads/master'
     runs-on: ubuntu-latest
+    concurrency: Review_${{github.event.inputs.pr}}
```
felixtheflex, line 273: what happens if not PR? ( github.event.inputs.pr)  e.g main\dev branch -, why not use `${{ github.ref }}`?

ethax-ross (approved):

.github/workflows/build.yml
sfawcett123: If it were not a PR it would not be in the review stage, the workflow only triggers on  workflow_dispatch  pull_request push:    branches: [ master ] ( workflow dispatch is only used for debugging ) and the review job only triggers on NOT master branches.  so it will only hit this job if a PR

PR #2206, opened 2022-01-18
Fix OWASP Action

sfawcett123:
Fix OWASP Action


peteryates (approved):
This makes sense to me

PR #2215, opened 2022-01-20
Removed reference to Transititon to Teach

tcannon3:
Also reworded the sentence to make sense as it is now referring to a single link rather than multiple

### Trello card
https://trello.com/c/cEZuk8Dn/2844-remove-references-to-transition-to-teach


gemmadallmandfe (approved):
Looks good too me

PR #2217, opened 2022-01-21
Adding content to make the user journey clearer

EllieNodder:
Adding content about how users can apply for scholarships

### Trello card

https://trello.com/c/VhYWFJy3/2705-review-and-update-links-from-find-and-apply-to-git

### Context

Find links to this GIT page saying 'find out how to apply for a scholarship', but the GIT page doesn't provide this information. If you click on the scholarship links on this page, they tell the user how to apply, so need to point this out to improve user journey.

### DIFF
--- a/app/views/content/funding-your-training.md
+++ b/app/views/content/funding-your-training.md
@@ -121,7 +121,7 @@ You’ll need a first, 2:1, 2:2 degree or a PhD or master's degree to be eligibl
 
 [Find out more about your eligibility for teaching bursaries](https://www.gov.uk/government/publications/funding-initial-teacher-training-itt/funding-initial-teacher-training-itt-academic-year-2022-to-2023).
 
-For a scholarship, each professional scholarship body sets its own criteria. You can find out how to apply for a scholarship by looking at each individual provider:
+You can apply for a scholarship to help pay for your teacher training course. The scholarships are offered by independent institutions. Each institution has their own eligibility criteria:
 
 * the [Royal Society of Chemistry](https://www.rsc.org/awards-funding/funding/teacher-training-scholarships/) (chemistry)
 * [BCS The Chartered Institute for IT](https://www.bcs.org/get-qualified/certification-and-scholarships-for-teachers/bcs-computer-teacher-scholarships/) (computing)
--- a/app/views/content/funding-your-training.md
+++ b/app/views/content/funding-your-training.md
@@ -121,7 +121,7 @@ You’ll need a first, 2:1, 2:2 degree or a PhD or master's degree to be eligibl
 
 [Find out more about your eligibility for teaching bursaries](https://www.gov.uk/government/publications/funding-initial-teacher-training-itt/funding-initial-teacher-training-itt-academic-year-2022-to-2023).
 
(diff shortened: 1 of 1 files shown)


app/views/content/funding-your-training.md
```diff
@@ -121,7 +121,7 @@ You’ll need a first, 2:1, 2:2 degree or a PhD or master's degree to be eligibl
 
 [Find out more about your eligibility for teaching bursaries](https://www.gov.uk/government/publications/funding-initial-teacher-training-itt/funding-initial-teacher-training-itt-academic-year-2022-to-2023).
 
-For a scholarship, each professional scholarship body sets its own criteria. These bodies include:
+For a scholarship, each professional scholarship body sets its own criteria. You can find out how to apply for a scholarship by looking at each individual provider:
```
Tomcowan1, line 124: Hi @EllieNodder for this, I think users will expect a link to the criteria, rather than to the application itself (which fits with the 'discovery mode' MO of the page. Would suggest changing the new wording to "To see the criteria for each provider, visit:"

Also, can we change the Maths link to http://teachingmathsscholars.org/about (it's the page with the criteria on).

EllieNodder: @Tomcowan1 will change the maths link. Changing the text to 'how to apply' was because [this page on Find](https://www.find-postgraduate-teacher-training.service.gov.uk/course/2AT/3CXF#section-financial-support) says 'find out how apply' and then links to this GIT page, which doesn't mention anywhere about applying. If you don't think we should be saying 'find out how to apply', do we perhaps need to then change the link text on Find? Maybe just 'find out about the different bursaries and scholarships available and their criteria' instead?

Tomcowan1: I think the wording on Find could be optimised. Is applying for a scholarship the user need here (on Find), or is it just finding out about what a scholarship is, and who offers them? One to discuss with Emma.

Also, I wonder if on our we're being a bit confusing/ jumping into the middle of the story. Should we add something simple like 


> "You apply for a scholarship through an institute that aims to support teachers of particular subject. To be eligible you need to meet their eligibility criteria. Each institute has their own criteria.
> 
> Institution names and links.


What do you think?

EllieNodder: @Tomcowan1 Will discuss the Find wording with Emma.

I feel like the first sentence for GIT is a bit confusing for me? Could we make it simpler?

> You can apply for a scholarship if you're planning to teach certain subjects. The different institutes that offer these scholarships have their own eligibility criteria:

Institution names and links

Or...


> Scholarships are available for certain subjects. The different institutes that offer these scholarships have their own eligibility criteria:

Institution names and links

Tomcowan1: Cool - let's go with something simpler:

"You can apply for a scholarship to help pay for your teacher training course. The scholarships are offered by independent institutions. Each institution has their own eligibility criteria."

EllieNodder: sounds good to me! I'll add it in

gemmadallmandfe (approved):
Looks good to me!

@Tomcowan1 Do you want to see this again before I merge it?

PR #2221, opened 2022-01-21
Make teaching events attend guidance match events

peteryates:
### Trello card

https://trello.com/c/TnMyQbOh/2880-fix-how-to-attend-display-in-new-teaching-events

### Context

Make the 'how to attend' section match the old one in events more closely.

Currenly the new and old events pages are sharing helpers, when the new supersedes the old we should move these to a single component and simplify the logic (and the big opening `if` statement) further.

### DIFF
--- a/app/views/teaching_events/show/_how-to-attend.html.erb
+++ b/app/views/teaching_events/show/_how-to-attend.html.erb
@@ -1,12 +1,36 @@
-<section class="teaching-event__how-to-attend">
-  <h2>How to attend</h2>
+<%
+    if @event.open? && [
+      can_sign_up_online?(@event),
+      is_event_type?(@event, "School or University event"),
+      @event.provider_website_url,
+      @event.provider_contact_email
+    ].any?
+%>
+  <section class="teaching-event__how-to-attend">
+    <h2>How to attend</h2>
 
-  <p>Registration for this event is free.</p>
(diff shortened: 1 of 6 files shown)


app/views/teaching_events/show/_how-to-attend.html.erb
```diff
@@ -1,12 +1,36 @@
-<section class="teaching-event__how-to-attend">
-  <h2>How to attend</h2>
+<%
+    if @event.open? && [
```
ethax-ross, line 2: Is it worth replacing the `allow_registration?` method with this logic?

ethax-ross (approved):
Looks good, one minor suggestion but can be done as part of the clean up you already have planned

app/views/teaching_events/show/_how-to-attend.html.erb
peteryates: Yeah, calling helpers from the presenter didn't feel quite right and I didn't want to duplicate them. Having it in a single method would be much nicer.

gemmadallmandfe:
Looks good (from what I can test in dev) @peteryates 

Do you want to raise a ticket about the additional clean up work for after the split test?

peteryates:
I think the best time to do the cleanup is when we drop the old (or new!) events listings, then we can organise them in one place without needing to make them shareable.

PR #2228, opened 2022-01-24
Only send virtual page view on Turbo page change

ethax-ross:
When we were using Turbolinks the `turbolinks:load` event was only called when Turbolinks performed the page change (even though the documentation says it should be called on the initial page load). Turbo calls `turbo:load` on the initial page load. To keep the behaviour consistent we need to ignore the first `turbo:load` event.

We are seeing high bounce rates in GA and it _could_ be related to this behaviour.


peteryates (approved):
Makes sense

PR #2239, opened 2022-01-26
Remove underlines from the hero h1

peteryates:
### Trello card

https://trello.com/c/JSzS7iRo/2846-remove-the-header-underline-inspire-the-next-generation-etc

### Context and changes

This also allows us to simplify the padding as we no longer need to take the `border-bottom` into account.

The h1 font size has also been increased to make the heading stand out from other similarly-sized text in/around the hero. The reduced padding (from removing the underline/bottom-border) means it doesn't take up much more space but definitely looks more heading-like.

| Before | After |
| ------- | ------- |
| Screenshot from 2022-01-26 11-45-18 | Screenshot from 2022-01-26 11-45-16 |
| Screenshot from 2022-01-26 11-45-32| Screenshot from 2022-01-26 11-45-34 |
| Screenshot from 2022-01-26 11-46-08 | Screenshot from 2022-01-26 11-45-56 |
| Screenshot from 2022-01-26 11-46-39 | Screenshot from 2022-01-26 11-46-33 |

### DIFF
--- a/app/webpacker/styles/sections/hero.scss
+++ b/app/webpacker/styles/sections/hero.scss
@@ -147,34 +147,19 @@ $mobile-cutoff: 800px;
 
     h1 {
       background-color: $yellow-dark;
-      padding: .8em .75em 1.2em 1.75em;
+      padding: .8em .75em .8em 1.75em;
       margin-left: -1em;
 
       @include mq($from: tablet) {
         margin-top: 2em;
-        padding: .8em 2em 1.2em;
+        padding: .8em 2em;
       }
 
       > span {
         @include font-size("large");
 
-        // user a border bottom to simulate a underline that
-        // looks like the branding. It's not perfect as the line
-        // extends beyond the edges of the text by a few px
-        border-bottom: 3px solid $black;
-
-        // proper underline styles - switch to this when (if) we
-        // start using a standard font. while we're relying on
-        // user-installed default sans fonts we run the risk
-        // of descenders breaking up the line
-        //
-        // text-decoration: underline;
-        // text-underline-offset: .3em;
-        // text-decoration-thickness: .1em;
-
         @include mq($from: tablet) {
           @include font-size("xxlarge");
-          border-bottom: 6px solid $black;
         }
       }
     }


gemmadallmandfe:
I think this looks better @peteryates, particularly with the font size increase - although we need to be mindful of pages with long titles such as this one:
image

@gazcarr My only concern is that we've moved away from the brand quite a bit with all these tweaks - is there anything we should do with this header to rectify that?

We should probably share with the brand team as well before implementing - whilst they didn't disagree with removing the underline, we just need to make sure we are keeping that consistent feel with campaigns

gazcarr:
@gemmadallmandfe I don't think there is much else we can do to these headings.

I am not sure about the increase in font size. Normally there is a hierarchy and font sizes are considered based on the size of the body copy. @peteryates what was the reason for the size increase?

peteryates:
The reason I added it was to make a hierarchy. Currently on prod the font size of the black banner, the header, subtitle and button are all more or less the same, unless you're familiar with the branding it may not be obvious which the header is.

Screenshot from 2022-01-26 14-04-31

gazcarr:
@peteryates In the design files I have:

black banner: 19px
h1: 28px
h2: 24px

peteryates:
Currently in prod (and unchanged for many months) the sizes are:

* `<h1>` =  `54px` (`3.375rem`)
* subtitle =  `28px` (`1.75rem`)
* black bar = `19px` (`1.188rem`)

We don't use px directly anywhere, the font sizes also [scale depending on screen size](https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/webpacker/styles/font-sizes.scss) - they get closer as they get smaller.

gemmadallmandfe:
@peteryates @gazcarr Are we happy with this one now?

gazcarr:
@gemmadallmandfe @peteryates yes good with me

gemmadallmandfe (approved):
Happy for this to be merged to get rid of the accessibility issue @gazcarr @peteryates 

However, I am still a little concerned about the h1 getting lost, particularly with the urgent call to arms banner on the home page. We have a separate ticket in the backlog about reviewing this header/promo area (not least as the skip to main content has issues which can't be resolved without a redesign) and I will put a note on there about the h1 visibility.

peteryates:
@gemmadallmandfe thought so, pushed an update.

PR #2243, opened 2022-01-26
Changing based on feedback from the funding team.

Tomcowan1:
### Trello card

https://trello.com/c/plti1RTV/2648-review-degree-requirement-for-bursaries

### DIFF
--- a/app/views/content/funding-your-training.md
+++ b/app/views/content/funding-your-training.md
@@ -117,7 +117,7 @@ You may be eligible for a teaching bursary or scholarship when training to teach
 
 These are tax-free amounts of money you receive to train in certain subjects. You do not need to pay them back.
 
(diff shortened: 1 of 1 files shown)


Tomcowan1:
Hi @gemmadallmandfe 

I've decided to edit the wording on the funding page slightly to meet requirements. I have added "bachelor's" and "honours" as these words are understood by graduates, and also reflect the language in Apply and on Get an Adviser. 

I have not included 'British' or '(or overseas equivalent)'. We have recently had feedback that a higher number of international candidates are applying or expecting bursaries when they are not eligible. We have updated the content on the page to reflect this (i.e "If you’re not a UK citizen, you are unlikely to get help funding your training"). I feel adding in the overseas elements will dilute this message.

Also, after checking Apply data, people are applying with the correct degrees and classifications and there is no noticeable trend in ordinary or other types of degree being entered.

gemmadallmandfe (approved):
Looks good to me

PR #2249, opened 2022-01-28
Change to arrow icon on homapage

gazcarr:
### Trello card
https://trello.com/c/pGmdGTqC

### Context
Check icon should not be used in isolation, provide an alternative

Switch the icon to the arrow

### DIFF
--- a/app/views/content/home/_directory.html.erb
+++ b/app/views/content/home/_directory.html.erb
@@ -2,7 +2,7 @@
   <%= render Content::GenericBlockComponent.new(
     title: "Life as a teacher",
     icon_image: "icon-school-black.svg",
-    icon_alt: "mortarboard icon",
+    icon_alt: "",
     icon_size: "40x30",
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
Looks fine to me but I'll leave it to @peteryates to approve :-)

peteryates (dismissed):
Looks good to me.

I wonder if the icons are decorative, perhaps we should change the alt to `""` - thoughts @gemmadallmandfe?

gemmadallmandfe:
> Looks good to me.
> 
> I wonder if the icons are decorative, perhaps we should change the alt to `""` - thoughts @gemmadallmandfe?

Yes - I have been meaning to mention this @peteryates! I agree that they are decorative

Can we change all three alts to "" and then merge this?

peteryates:
@gemmadallmandfe thought so, pushed an update.

peteryates (approved):

PR #2273, opened 2022-02-01
Improve Core Web Vitals scores

ethax-ross:
### Trello card

[Trello-2912](https://trello.com/c/4JkZiePw/2912-investigate-increase-in-core-web-vitals-cls-warnings-for-desktop)

### Context

Our Core Web Vitals scores are indicating a number of pages that could be improved. The majority are related to Cumulative Layout Shift and appear to be pretty much globally across pages. A smaller subset is a long Largest Contentful Paint; this appears to be related to a particular event that is no longer active, but I expect the image was particularly large and resulted in a longer LCP (although it was only 0.1s over the 2.5s threshold!). As this event is in the past I think its safe to ignore this one and focus on the CLS issues.

- Add explicit width to get boosted image

An image without a width/height causes cumulative layout shift.

- Remove aspect-ratio sizing of images

Sizing images with an aspect ratio results in Cumulative Layout Shift; we need to be  explicit with the dimensions to avoid that.

- Prevent search in header causing CLS

The search link in the header, along with the search autocomplete input, cause
an amount of Cumulative Layout Shift.

To prevent this we ensure there is enough space for the autocomplete input to render into. The JS disabled search link is causing the majority of the CLS, though I have absolutely no idea why and no combination of moving it around/messing with the CSS appears to resolve it. Instead, we're ripping it out of the header and putting a link in the footer so that those users can still access search (there have only been 35 page views of it in the last 6
months).

Load `js_enabled.js` pack synchronously so that we don't get flashes of JS enabled/disabled content that is later hidden (also a cause of CLS).

- Reduce content shift from image in hero section

Defaulting the `fit-content` declarations to 0 reduces the amount of content shift when laying out the image in the grid.

### DIFF
--- a/app/components/footer_component.html.erb
+++ b/app/components/footer_component.html.erb
@@ -53,7 +53,9 @@
       <div class="site-footer-bottom__logo-container">
         <%= image_pack_tag 'media/images/dfelogo-white.svg', alt: "Department for Education", size: "124x73" %>
         <%# image_pack_tag 'media/images/plan-for-jobs.svg', alt: "Plan for jobs", size: "124x73" %>
-        <a href="https://www.nhs.uk/conditions/coronavirus-covid-19/coronavirus-vaccination/book-coronavirus-vaccination/"><%= image_pack_tag 'media/images/get-boosted-now.png', alt: "Get boosted now", height: "80" %></a>
+        <a href="https://www.nhs.uk/conditions/coronavirus-covid-19/coronavirus-vaccination/book-coronavirus-vaccination/">
+          <%= image_pack_tag 'media/images/get-boosted-now.png', alt: "Get boosted now", width: "274", height: "80" %>
+        </a>
       </div>
       <div class="site-footer-bottom__links-container">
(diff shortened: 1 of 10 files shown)


peteryates (approved):
Makes sense, everything looks like it's working fine.

gemmadallmandfe:
Thanks for this @ethax-ross 

* Get boosted logo - looks good to me

* Javascript/search - have tested this and it works fine. Given the very small number of searches by people with Javascript disabled, I think that is a reasonable approach; search is still available to them

* Image aspect ratio/hero images - could you just run me through this on the site to show the impact? I can see that some (but not all) of the hero images have changed to have a different crop (reduced height) but just want to understand the implications better

gemmadallmandfe (approved):
Thanks for talking me through the image changes @ethax-ross 

As discussed, I don't think it is worth pinning the hero images to combat the reduction in height on some of them - that whole promo section needs revising properly anyway.

Given that this has had tech eyes on it already, happy for this to be merged tomorrow (Thurs)

gemmadallmandfe:
@ethax-ross Once these changes are in production, we need to ask Google Search Console to validate the CLS fix on both mobile and desktop - are you happy to do this?

ethax-ross:
Yep I can do that @gemmadallmandfe

PR #2289, opened 2022-02-09
Remove alt text from blog post as it's centralised

peteryates:
Simply tidying up a bit of front matter


gemmadallmandfe (approved):
Looks good to me (presuming centralised alt text is marked as decorative?)

PR #2293, opened 2022-02-11
Account for links that are missing a href attribute

ethax-ross:
### Trello card

[Trello-2954](https://trello.com/c/xw7E8YEN/2954-fix-bug-in-external-link-icon-helper)

### Context

Some of our links - such as those in the event descriptions - go through the `sanitize` Rails method. We allow `href` attributes here, however under the hood Rails uses `Loofah` to only allow certain whitelisted protocols. If a link is fat-fingered to contain an invalid protocol, for example `hhttp://test.com`, then it will be stripped out by this method.

Currently if this happens we get an exception as the code to add external link icons encounters a `nil` `href` attribute. Instead, we want to continue rendering the page (albeit with the broken link) and alert in Sentry so that we are aware and can ask the events team to fix it.

- Account for links that are missing a href attribute

Update external link icon to account for `nil` `href` attributes (raising a message in Sentry with details, but continuing to render).

Clean up tests around the external link icons which weren't actually testing anything as the test markdown files had been changed; the new tests should prevent that from happening again.

### Guidance to review

The code to add external link icons could be cleaned up, but I think we're going to be changing this from a design-perspective in the near future anyway, so I'll not get into it here.


jkempster34 (approved):
Nice 💯💯

PR #2294, opened 2022-02-11
Use Redis for session storage

ethax-ross:
### Trello card

[Trello-2953](https://trello.com/c/B3NLLZNm/2953-move-to-active-record-session-store)
[Sentry](https://sentry.io/organizations/dfe-teacher-services/issues/3000094940/?project=5276949&query=is%3Aunresolved)

### Context

We are seeing cookie overflow errors in production; this happens when more than 4kb is stored in a cookie (the current session storage medium). I imagine this happens when someone signs up to multiple events as we key the session information by event id so its going to be additive.

To avoid this we can tell Rails to use Redis as the session store in production.

- Use Redis for session storage

### Guidance to review

 I've set the TTL to 1 day; I figure if a user hasn't completed a form by then its not unreasonable that they should start again. When we ship this people mid-form will get reset, but there's not much we can do about that...


peteryates (approved):
Makes sense

PR #2303, opened 2022-02-15
Make 'Not now' button accessible

ethax-ross:
### Trello card

[Trello-2924](https://trello.com/c/4wtGkNbp/2924-ensure-mailing-list-bar-buttons-can-be-understood-by-screen-readers)

### Context

Silktide flagged the mailing list banner 'Not now' button as inaccessible because it doesn't provide a `href` and is therefore not behaving like a link.

To fix this we can give the anchor link a role of `button`. I've also given it an explicit `href` that does nothing.

- Make 'Not now' button accessible


gemmadallmandfe:
I looked at this yesterday from a user perspective and all looked good.

Will leave for @peteryates to approve from a tech perspective!

peteryates (approved):
Looks good

PR #2304, opened 2022-02-15
Run tests in parallel during CI

ethax-ross:
### Trello card

[Trello-2970](https://trello.com/c/tVU8E9Hg/2970-the-unit-test-on-git-can-take-a-considerable-time)

### Context

The test suite takes around 7 minutes to complete in CI; we can speed this up by running it in parallel.

- Run linting in separate job

The linters currently run with the test suites, but as we want to parallelize the test suite there's no point running the linters multiple times; they can also be ran in parallel if they are split out.

- Parallelize test suite with Knapsack

Uses the Knapsack gem to parallelize the test suite; we can use the pro version for free as our tests run in under 10 minutes in total.

- Optimize Docker image

Minor updates to speed up the docker build.

- Optimize redirect tests

The redirect tests were taking a fair chunk of the suite run time at 1m 30s. To cut down the time we can combine the two types of tests (so always test with a query string) and cache the results of subsequent requests to ensure we don't check the same redirect locations multiple times.

- Avoid re-downloading the Docker image

We build the Docker image to a registry and then have to download it multiple times in future jobs which is slow. Instead we can build it and save it as an artefact (~10s) and then run the image from the artefact (~20s instead of ~1m).

- Multi-stage docker image

Our docker image is used for both running tests and production so its fairly bloated. We also do a bunch of image processing so need several libraries for this.

To make the docker image smaller (saving ~500mb down from ~1GB) we can create another stage that only has the app files and required dependencies.

In future we should probably look to have multiple dockerfiles; one for test and one for production, then we can avoid including the test gems and packages (chromium/driver).

- Run QA deploy in parallel with development

We can save time by running the QA and development deployments in parallel. It should be safe to assume in the dev deploy fails then in all likelihood the QA deploy will fail for the same reason.

### Guidance to review

Knapsack is free for us at the moment but it does mean another account/token to manage... we can just downgrade to the non-pro version if we start exceeding the 10m total test runtime limitation. With pro mode you get test balancing and it gives you a nice performance breakdown of the tests by spec file which isn't trivial to get out of rspec otherwise.

This PR brings the CI runtime down from ~16m to ~11m (with re-runs closer to 8m).

The UI In Knapsack is pretty swish too:

Screenshot 2022-02-17 at 12 58 46


peteryates (approved):
Nice ⚡

PR #2312, opened 2022-02-17
Remove external link checking specs

peteryates:
### Trello card

https://trello.com/c/SqBJ6ZOK/2957-missing-link-checker-times-out

### Context and changes

- Remove external link specs
- Add some docs on link checking


sfawcett123:
Comment looks good, I think we now need to get the results to output to Slack and once we have that we can retire the previous method

sfawcett123:
this might be the key, https://api.slack.com/methods/files.upload

ethax-ross (approved):

PR #2319, opened 2022-02-18
Split dockerfile for test/release

ethax-ross:
We don't need to do a bunch of time-consuming tasks (mainly related to asset preprocessing) for the test docker image.

If we split the dockerfile (so we can either build it for test or production) we should be able to start running the tests much sooner and the longer, production docker build process won't delay things.

Copies the `node_modules` from previous builds to speed up the `yarn install`.

We install the test packages (chromium/driver) in the initial `apk add` to benefit from the layer cache; it's more efficient to remove these from the production image than to add them separately later to the test image.

Brings the build time from ~11m down to ~8m.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,8 +7,8 @@ on:
     branches: [ master ]
 
 jobs:
-  build:
-    name: Build
+  build_release:
+    name: Build (Release)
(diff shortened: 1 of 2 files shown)


peteryates (approved):
Nice, all looks sensible to me.

PR #2323, opened 2022-02-18
Fix headers in iQTS for accessibility

laurahermionetennant92:
Fix headers in iQTS for accessibility


EllieNodder:
@laurahermionetennant92 this has failed to build for some reason. @peteryates would you be able to take a look when you get chance? :)

EllieNodder (approved):
Looks good to me!

PR #2324, opened 2022-02-21
Add canonical tags to all web pages

ethax-ross:
### Trello card

[Trello-2986](https://trello.com/c/4Dm3u0rf/2986-avoid-serving-web-pages-on-the-assets-domain)

### Context

We serve the app assets from another domain in production (app-assets-getintoteaching.education.gov.uk) and Google is indexing some of our pages on this domain instead of the production domain (as we serve the web pages as well as the assets).

To prevent this and avoid it happening again in the future we can indicate to Google that we want getintoteaching.education.gov.uk to be used as the canonical domain.

- Add canonical URL tags to all pages

- Set DOCKER_SHA env variable

We do this so that we can setup a manual release to test.

### Guidance to review

I also tried to make it so the routes didn't exist for asset domains (hence the second commit in this PR, which I've left in in case we need to push to test in the future), but it ended up feeling a bit hacky and I think the canonical tags should resolve the issue.


peteryates (approved):
Looks good.

PR #2333, opened 2022-02-24
Improve Network Policies

sfawcett123:
Create the network policy in Terraform

### DIFF
--- a/terraform/paas/variables.tf
+++ b/terraform/paas/variables.tf
@@ -21,9 +21,13 @@ variable "azure_vault_secret" {
   default = "CONTENT-KEYS"
 }
 
+variable "paas_monitoring_space" { }
+
 variable "paas_space" {
   default = "sandbox"
 }
+
+variable "paas_monitoring_app" { }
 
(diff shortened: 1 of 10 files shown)


terraform/paas/variables.tf
```diff
@@ -21,10 +21,18 @@ variable "azure_vault_secret" {
   default = "CONTENT-KEYS"
 }
 
+variable "paas_monitoring_space" {
+  default = "sandbox"
```
saliceti, line 25: If we always expect to pass a value, we can remove the default

terraform/paas/variables.tf
```diff
@@ -21,10 +21,18 @@ variable "azure_vault_secret" {
...
+}
+
 variable "paas_space" {
   default = "sandbox"
 }
 
+variable "paas_monitoring_app" {
+   default  = "prometheus-dev-get-into-teaching"
```
saliceti, line 33: If we always expect to pass a value, we can remove the default

saliceti (approved):

PR #2334, opened 2022-02-25
Output Broken links to SLACK

sfawcett123:
Github actions does not like multiple lines when using SET-OUTPUT so put the data into an environment variable. Additionally the markdown from SLACK is not the same as anyone else's markdown, so ran through a convertor to make the format nicer.
Also removed the old link checker as it was just hanging and timing out.


.github/workflows/lychee.yml
```diff
@@ -37,14 +37,25 @@ jobs:
...
 
       - name: Slack Notification
         if: ${{ steps.lychee.outputs.exit_code != 0 }}
         uses: rtCamp/action-slack-notify@master
         env:
            SLACK_COLOR: ${{env.SLACK_ERROR}}
-           SLACK_MESSAGE: |
-             ${{ steps.lychee-output.outputs.LYCHEE_OUTPUT }}
+           SLACK_MESSAGE: | 
+                         ${{ steps.convert.outputs.text }}
```
peteryates, line 59: Nice - didn't know you could do this! 🤓

sfawcett123: it is this action 

```
      - name: Slack Markdown Converter
        id: convert
        if: ${{ steps.lychee.outputs.exit_code != 0 }}
        uses: LoveToKnow/slackify-markdown-action@v1.0.0
        with: 
          text: |  
                ${{ env.LYCHEE_OUTPUT }} 
```

the main change is where we stop using SET-OUTPUT and sticj the result in an environment variable. SET-OUTPUT is rubbish for multi line text

peteryates (approved):

PR #2339, opened 2022-02-28
Fix cumulative layout shift on desktop

ethax-ross:
- Fix cumulative layout shift on desktop

Loading the Javascript async in the `<head>` appears to result in fairly random Cumulative Layout Shift (CLS) on the page. It appears to effect the header section in particular, but I'm not certain why (it event causes hidden elements to result in a layout shift!).

Changing the Javascript to defer loading clears _most_ of this up, and doesn't appear to have any negative impact.

- Fix intermittent JS exception

We are seeing an intermittent exception where the `labelTarget` is no longer available; it appears to get removed from the DOM on occasion when changing pages. I think it may be something to do with the Turbolinks page cache, but I'm not certain what yet.

Putting in a conditional check to ensure the target is available quietens the error for now, and as the input still appears to have the correct `aria-label` set when this happens it should still remain accessible.

### DIFF
--- a/app/views/sections/_head.html.erb
+++ b/app/views/sections/_head.html.erb
@@ -13,10 +13,10 @@
   <% end %>
   <%= google_optimize_script %>
   <%= javascript_pack_tag 'sentry', 'data-turbolinks-track': 'reload', data: { "sentry-dsn": sentry_dsn, "sentry-environment": Rails.env } %>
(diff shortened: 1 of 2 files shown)


peteryates (approved):
This makes sense and looks like it does stop the issue from happening - we must be doing something in one of the stimulus controllers that causes this but having had a look through I can't see any smoking guns.

PR #2352, opened 2022-03-01
Apply latest internship provider updates from spreadsheet

peteryates:
### Context and changes

This site has a issue with TLS when using www.bfet.co.uk but not when
using bfet.co.uk


SaimaBodiat (approved):
looks good :)

PR #2362, opened 2022-03-02
Remove external link icons

peteryates:
### Trello card

https://trello.com/c/I8LV7KMd/2613-remove-external-link-icons

### Context and changes

- Replace external link icons with descriptive text
- Remove manual external link icon from events page
- Remove external link icon and remaining references

### Guidance to review

I think it's worth evaluating how the site looks (and reads) without the icons before merging. Does all the extra `visually-hidden` 'link opens in new window' text add to the noise for screen reader users?

### DIFF
--- a/app/webpacker/controllers/link_controller.js
+++ b/app/webpacker/controllers/link_controller.js
@@ -13,22 +13,14 @@ export default class extends Controller {
   }
 
   openExternalContentLinksInNewWindow() {
-    const links = this.contentTarget.querySelectorAll('a');
+    const links = Array.from(this.contentTarget.querySelectorAll('a'));
 
-    links.forEach((l) => {
-      if (this.isExternal(l)) {
+    links
+      .filter((l) => this.isExternal(l))
+      .forEach((l) => {
         l.setAttribute('target', '_blank');
         l.setAttribute('rel', 'noopener');
-
-        // add hidden text to inform screen reader users that
-        // link will open in a new window
-        const linkOpensInNewWindow = document.createElement('span');
-        const text = document.createTextNode('(Link opens in new window)');
-        linkOpensInNewWindow.classList.add('govuk-visually-hidden');
-        linkOpensInNewWindow.appendChild(text);
-        l.appendChild(linkOpensInNewWindow);
-      }
-    });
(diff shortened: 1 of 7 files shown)


ethax-ross (commented):
Starting looking at this but don't think its working yet?

Screenshot 2022-03-08 at 08 04 46

peteryates:
@ethax-ross - it looks like there was a prettier fail on the last push so the review app was out of date. The `icon-external.svg` is removed so if it's still there when it regenerates I'll be worried! 😓

app/webpacker/controllers/link_controller.js
```diff
@@ -13,22 +13,14 @@ export default class extends Controller {
   }
 
   openExternalContentLinksInNewWindow() {
-    const links = this.contentTarget.querySelectorAll('a');
+    const links = Array.from(this.contentTarget.querySelectorAll('a'));
 
-    links.forEach((l) => {
-      if (this.isExternal(l)) {
+    links
+      .filter((l) => this.isExternal(l))
```
ethax-ross, line 19: We should probably modify links in either Ruby or JS, but not both; one for another day though 😬

ethax-ross (approved):

gemmadallmandfe:
@peteryates Looks good to me

I have one minor question - for screen readers, external links are currently marked as `(Link opens in new window)` but in this PR they are `(opens in new window)` - is that deliberate?

When I tested with VoiceOver, the latter almost seemed part of the link text so a bit harder to understand. The slight pause between the link text and the `opens in new window` had also disappeared - not sure if the lack of initial capital letter affected this?

However, if your changes follow gov.uk best practice, ignore my comments!

Otherwise happy for this to be merged tomorrow once we've discussed this

peteryates:
@gemmadallmandfe I tried to follow the [design system's guidance] (https://design-system.service.gov.uk/styles/typography/#opening-links-in-a-new-tab) more closely than we did.

In their example they say 'tab' but I think this depends on platform/device.

gemmadallmandfe (approved):
Thanks for clarifying @peteryates - happy for you to merge this tomorrow :-)

PR #2367, opened 2022-03-03
Updating-content-around-RTTA-eligibility

EllieNodder:
### Trello card

https://trello.com/c/F4WbBob9/2993-consider-most-accurate-way-of-describing-eligibility-for-an-rtta

### DIFF
--- a/app/components/footer/talk_to_us_component.html.erb
+++ b/app/components/footer/talk_to_us_component.html.erb
@@ -23,7 +23,7 @@
             <h3>Get a teacher training adviser</h3>
             <p>
               An experienced teaching professional can help if you're ready to get into
-              teaching or if you're returning to teaching and qualified to teach chemistry, computing, maths, modern foreign languages or
+              teaching or if you already have QTS and want to teach chemistry, computing, maths, modern foreign languages or
               physics.
             </p>
 
--- a/app/components/footer/talk_to_us_component.html.erb
+++ b/app/components/footer/talk_to_us_component.html.erb
@@ -22,9 +22,7 @@
           <div>
             <h3>Get a teacher training adviser</h3>
             <p>
-              An experienced teaching professional can help if you're ready to get into
-              teaching or if you already have QTS and want to teach chemistry, computing, maths, modern foreign languages or
-              physics.
+              Teacher training advisers provide free, one-to-one support wherever you are in your journey into teaching. Whether it's just an idea, or you're looking to get back into the classroom, you could get personalised help and support.
             </p>
 
             <a href="/tta-service" class="button">
--- a/app/views/content/blog/top-tips-for-returning-teachers.md
+++ b/app/views/content/blog/top-tips-for-returning-teachers.md
@@ -87,4 +87,4 @@ Once you have made the decision to return to teaching what’s the best way of f
 
 Remember though that teacher recruitment goes in cycles, with most teaching roles being advertised from January onwards. Setting up a job alert will mean you don’t miss out on your perfect role.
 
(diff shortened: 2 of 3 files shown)


app/components/footer/talk_to_us_component.html.erb
```diff
@@ -23,7 +23,7 @@
             <h3>Get a teacher training adviser</h3>
             <p>
               An experienced teaching professional can help if you're ready to get into
-              teaching or if you're returning to teaching and qualified to teach chemistry, computing, maths, modern foreign languages or
+              teaching or if you already have QTS and want to teach chemistry, computing, maths, modern foreign languages or
```
gemmadallmandfe, line 26: At the risk of making this footer block even longer, I think we should say the word 'return' in here somewhere otherwise it is a bit misleading?

And should we say qualified teacher status rather than QTS? (not sure about this one - happy for you to disagree!)

eg 'If you have qualified teacher status and you want to return to teach....'

EllieNodder: @gemmadallmandfe given that we might soon be adding even more subjects to RTTA eligibility, do we maybe just get rid of the subject eligibility text in this footer altogether, given it's already quite a heavy sentence and replace it with something more about what advisers do?

i.e. 'Teacher training advisers provide free, one-to-one support wherever you are in your journey into teaching. Whether it's just an idea, or you're looking to get back into the classroom, you could get personalised help and support.'

Might also make people more likely to click it, if it explains what they could get out of it, rather than what makes them eligible? (they then find this info out on the next page)

Alternatively, if you think we should keep the subjects in, break it up into two sentences?

'An experienced teaching professional can help if you're ready to get into teaching. You can also get support if you have qualified teacher status and are returning to teach chemistry, computing, maths, modern foreign languages or physics.'

EllieNodder:
Have received policy guidance from Phoebe.

app/components/footer/talk_to_us_component.html.erb
gemmadallmandfe: @EllieNodder I've been having similar thoughts re (not) listing out subjects :-)

I think a problem with the current Get an adviser block is that the RTTA bit overshadows the TTA bit, because it is SO much longer; if you were looking to do ITT in (say) English and glanced at it, you may think it is not for you as your subject isn't there.

I really like your 'Teacher training advisers....' example as it focuses more on the benefits/journey stage rather than the eligibility (I'd maybe add something around ready to apply/support with applications?)

However, I'm happy to leave it to you and @valentine-carter as the content experts!

EllieNodder: @valentine-carter thoughts on this? A little wary about adding something to do with applications, just because there's not the room to talk about everything they can do, and I don't want to put people off who aren't at the application stage by minimising what TTAs can help with i.e. they can also help with finding school experience, information around financial support, etc.

valentine-carter: I think it would be better to add in more of what they can help with rather than list the subjects. Unless they can only help people who do those subjects? I agree that include specifics can put off the people that it doesn't apply to.

EllieNodder: TTAs can help graduates/career changers/finders interested in teaching any subject as long as they have/are getting a degree 2:2 or higher. TTAs can also help ex teachers returning to teaching but only if they want to teach those specific subjects.

Shall we go with the above suggestion then?

'Teacher training advisers provide free, one-to-one support wherever you are in your journey into teaching. Whether it's just an idea, or you're looking to get back into the classroom, you could get personalised help and support.'

valentine-carter: Yes, I expressed my reply badly, apologies!

valentine-carter (approved):

gemmadallmandfe (approved):
Looks good to me

PR #2368, opened 2022-03-04
Fix abbr underline styles on Safari mobile

peteryates:
### Context

Fix the `abbr` styles on Safari mobile.

### DIFF
--- a/app/webpacker/styles/markdown.scss
+++ b/app/webpacker/styles/markdown.scss
@@ -89,7 +89,7 @@
 
   @mixin abbr-styles {
     text-underline-offset: .2em;
-    text-decoration-thickness: 2em;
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
@peteryates I did some further testing and the original issue was not just limited to Safari - it was also apparent in Chrome and Edge on iPhone.

However, this fix works for all three browsers :-)

gemmadallmandfe (approved):
Looks good to me!

peteryates:
Thanks @gemmadallmandfe - under the hood on iOS all browsers have to use Safari's rendering engine Webkit - I think that would explain why it happens across all three. On Android/Chrome (Blink) it looked ok:

PR #2371, opened 2022-03-07
Migrate Cypress integration tests into the app

ethax-ross:
### Trello card

[Trello-2978](https://trello.com/c/gTxvgRKZ/2978-pull-cypress-tests-into-the-respective-apps)

### Context

We have another repository of integration tests written with Cypress. Often we find a change in the app has a knock-on effect of breaking these tests and its a pain to dive into a separate repository to fix them. Having them as part of the app means we're less likely to inadvertently break them and, if they do break, it's going to be easier to fix them. The CI run is also over-complicated as they are in a different repository; bringing them into the same repo makes this much simpler and hopefully quicker to run as well.

- Migrate Cypress integration tests into the app

### Guidance to review

You can run them locally with `rspec --tag integration`. You will also need to set these environment variables (found in the dev/test `INFRA-KEYS` key vault in Azure):

```
HTTP_USERNAME
HTTP_PASSWORD
MAILSAC_API_KEY
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -549,10 +549,9 @@ jobs:
           SLACK_MESSAGE: Failure with initialising QA deployment for ${{env.APPLICATION}}
           SLACK_WEBHOOK: ${{ steps.keyvault-yaml-secret.outputs.SLACK-WEBHOOK }}
 
-  cypress:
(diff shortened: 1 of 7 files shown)


peteryates (approved):
This is **so much nicer**

PR #2373, opened 2022-03-08
Events image replace

gazcarr:
### Trello card
https://trello.com/c/a73OtvQU/2975-update-imagery-on-website-to-show-events

### Context
Replace events image on the homepage 

Check the images on the box calling out 'Events'


SaimaBodiat:
Hi - We also need the events landing page image changing and i'm not too sure the image here conveys teacher training event. Perhaps this one (https://drive.google.com/file/d/1PLvHxqOyzcxAkthz-oAW8pIDBU3uU1mf/view?usp=sharing ) is better as it shows a provider talking to a candidate and you can also see the stalls in the background which would give candidates a better idea of what an event looks like.

gazcarr:
@SaimaBodiat OK, the new image is up on the events landing page. 👍

SaimaBodiat:
@ethax-ross Looks good to me! We should also take the opportunity to maximise the use of key word searches so changing the image description. Perhaps "Teacher training provider advising a potential teacher trainee at an in person teacher training event"? think that is 101 characters so need to think about making that succinct.

config/images.yml
```diff
@@ -143,6 +143,12 @@
   variants:
     - "media/images/content/hero-images/0024--mobile.jpg"
     - "media/images/content/hero-images/0024--tablet.jpg"
+    
+"media/images/content/hero-images/0025.jpg":
+  alt: "Helper talking to an attendee at an in person event."
```
SaimaBodiat, line 148: We should take the opportunity to maximise the use of key word searches so changing the image description. Perhaps "Teacher training provider advising a potential teacher trainee at an in person teacher training event"? think that is 101 characters so need to think about making that succinct.

gemmadallmandfe: SEO definitely important as discussed - including 'teacher training event' definitely a good idea

However, do be mindful of screen readers which is the main purpose of alt text - this is very verbose/repetitive if you are listening to the description being read out?

SaimaBodiat: Yes makes sense - was an initial draft and suggestion and was hoping others could input/review before changes were made.

gazcarr: @peteryates do we need to have a 'variant image' when adding the alt text? Not quite sure how this works

peteryates: No, the variant entry is only needed when there are multiple versions of the same photo. I've pushed @EllieNodder's alt text for the homepage images.

SaimaBodiat (approved):
Looks good to me! :)

peteryates (approved):
Looks good

PR #2375, opened 2022-03-09
Make it more obvious which SHA the input wants

peteryates:
### Context

The `SHA` input was confusing, I thought it meant the git SHA but it actually wants the Docker tag.


saliceti (approved):
Thanks

PR #2381, opened 2022-03-14
Remove alt text from CTA icons

peteryates:
### Trello card

https://trello.com/c/A5arEJ69/3025-mark-icons-as-decorative

### Context

The alt text is generated from the icon used, it's not helpful and should be removed.


gemmadallmandfe:
I've retested on my iPhone and was able to recreate it (but also able to resize the pages back to normal). However, I then tested _again_ in a private Safari tab and all looked fine.

Given that Ross can't recreate and there is no reason that this should cause width issues, happy for this to be merged if you agree @peteryates (and then I'll check in production!).

gemmadallmandfe (approved):

PR #2389, opened 2022-03-18
Fixing-disorderly-headers

valentine-carter:
### Trello card

https://trello.com/c/KTDmQu8J/2913-fix-skipped-headings-where-h3s-are-used-instead-of-h2s

Changed headers on Early years, Come to England to teach and Assessment only pages.

### DIFF
--- a/app/views/content/assessment-only-providers.md
+++ b/app/views/content/assessment-only-providers.md
@@ -55,7 +55,7 @@ provider_groups:
       link: http://www.nett.org.uk/
       telephone: '01255431949'
       email: teach@nett.org.uk
-    - header: The Pilgrim Partnership 
+    - header: The Pilgrim Partnership
       name: Bev Collier
       telephone: 01234 907896
       email: bcollier@pilgrimpartnership.org
@@ -108,7 +108,7 @@ provider_groups:
       link: https://www.lincolnshirescitt.co.uk/aor
       name: Angela Murphy
       telephone: 01522 889297
-      email: scittenquiries@prioryacademies.co.uk      
+      email: scittenquiries@prioryacademies.co.uk
     - header: Nottingham Trent University
       link: https://www.ntu.ac.uk/
       name: Admissions team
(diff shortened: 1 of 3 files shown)


gemmadallmandfe:
@valentine-carter The **AO page** and the **Come to England** page look fine to me.

The **early years** changes you have made look fine.

However, the rest of the early years page looks a bit odd with three h2s (Ways to train plus the two tables).

I'd suggest:
* considering whether the content in the two tables would be better presented not in a table - it seems like quite a lot of effort in terms of the frontmatter
* if you think it should be in a table, the headings should probably be h3s rather than h2s OR the 'Ways to train' heading should be removed?
* if the 'Ways to train' heading is to be kept, I think it should be reworded as we are already using this in our top nav - confusing for Google

Let me know what you think - if further changes are needed to the early years page, I'd suggest raising a separate ticket. I can then approve/merge this one!

valentine-carter:
> @valentine-carter The **AO page** and the **Come to England** page look fine to me.
> 
> The **early years** changes you have made look fine.
> 
> However, the rest of the early years page looks a bit odd with five h2s (Ways to train plus the four tables).
> 
> I'd suggest:
> 
> * considering whether the content in the four tables would be better presented not in a table - it seems like quite a lot of effort in terms of the front matter
> * if you think this content should be in tables, the headings should probably be h3s rather than h2s OR the 'Ways to train' heading should be removed?
> * if the 'Ways to train' heading is to be kept, I think it should be reworded as we are already using this in our top nav - confusing for Google?
> 
> Let me know what you think - if further changes are needed to the early years page, I'd suggest raising a separate ticket. I can then approve/merge this one!

Yes, I agree. I'd like to take a look at the early years page properly. I was trying to resolve the issue at hand without making too many big changes but I think it would be benefit from a refresh. I couldn't work out what was going on with the styling of the table to do an easy fix, it was most confusing.

gemmadallmandfe (approved):
Looks good to me (early years page to be addressed separately!)

PR #2391, opened 2022-03-21
Send sub_channel to BAM tracking pixel

ethax-ross:
### Trello card

[Trello-2925](https://trello.com/c/TDCO8mZL/2925-investigate-request-from-bam-re-tracking-sign-ups-by-iphone-safari-users)

### Context

BAM are struggling to track conversions due to Safari blocking their session cookie. To work around this while still maintaining user privacy we are going to accept a `sub_channel` parameter as part of the mailing list sign up; when the user reaches the completion page we should include the `sub_channel` in the tracking pixel request.

- Send sub_channel to BAM tracking pixel

Also fixes an issue where the channel/sub-channel was dropped when a candidate hit a validation error and the page reloaded.

### DIFF
--- a/app/models/mailing_list/steps/name.rb
+++ b/app/models/mailing_list/steps/name.rb
@@ -7,6 +7,7 @@ module MailingList
       attribute :last_name
       attribute :email
       attribute :channel_id, :integer
+      attribute :sub_channel_id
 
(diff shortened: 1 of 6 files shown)


peteryates (approved):
Looks good

PR #2398, opened 2022-03-25
remove video and quotes from top tips for returning teachers blog

EllieNodder:
### Trello card

https://trello.com/c/mr8ENLHO/3062-remove-private-video-from-top-tips-for-returning-teachers-blog

Removed the video that has now been set to private. Also removed two of the RTTA quotes - these come before we mention what an RTTA is, which doesn't seem to make sense to me for the user. The second quote is also very long. Have discussed these changes with Valentine.


gemmadallmandfe (approved):
Looks good to me

PR #2419, opened 2022-04-04
Add BAM tracking data to DOM on welcome guide completion

ethax-ross:
### Trello card

[Trello-3087](https://trello.com/c/V9TkAKtI/3087-bam-magic-pixel-not-working-on-safari)

### Context

Previously we were only including the sub-channel for the BAM tracking pixel in the DOM for the normal (non-welcome guide) variant of the mailing list completion page.

Ensure it renders for all mailing list completion pages.

- Add BAM tracking data to DOM on welcome guide completion

### Guidance to review

Also removes redundant checks for accept privacy policy error message in feature specs.


peteryates (approved):
Looks good

PR #2428, opened 2022-04-06
Prevent pages from being indexed by search engines

ethax-ross:
### Context

We want to prevent certain pages from being indexed by search engines:

- The entire callback booking form
- The cookie preferences page
- All the mailing list steps apart from the first step
- The search page (for non-JS users only)
- The generic thank you page used by the marketing team (also needs removing from the sitemap)
- All the PDF documents/assets (served under /packs)

- Prevent pages from being indexed by search engines

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -28,6 +28,10 @@ protected
     []
   end
 
+  def noindex
+    @noindex = true
+  end
+
 private
 
   def declare_frontmatter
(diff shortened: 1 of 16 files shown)


gemmadallmandfe:
@ethax-ross I've looked at all these pages in the review app and all looks good to me.

The only thing I can't check is the media assets, although I can see you have disallowed them in the robots.txt.

I haven't approved in case you want tech eyes on this before merging - but I'm happy for it to be merged tomorrow (Thurs).

Once it has been merged, I'll try and request re-indexing of the pages in question to try and encourage quicker removal from the index.

peteryates (approved):
Looks good

PR #2431, opened 2022-04-07
Prevent duplicate canonical tags on events pages

ethax-ross:
### Trello card
[Trello-3113](https://trello.com/c/9XC6fJq0/3113-set-canonical-urls-to-prevent-pages-with-query-parameters-being-indexed-by-google-search-console)

### Context

We have custom canonical tags on the new teaching events pages that point back to the original events pages (we do this to keep search engines happy because they are A/B tested).

We have since added a canonical tag to _all_ web pages (to avoid an issue where our assets domain was being indexed). This has resulted in multiple canonical tags on the events pages. To avoid this we make the canonical tag overridable
by templates.

- Prevent duplicate canonical tags on events pages

Also adds full domain to the custom events canonical tags (to be inline with the others).

### DIFF
--- a/app/helpers/text_formatting_helper.rb
+++ b/app/helpers/text_formatting_helper.rb
@@ -7,6 +7,6 @@ module TextFormattingHelper
   end
 
   def safe_html_format(html)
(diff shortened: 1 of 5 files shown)


peteryates (approved):
Looks good

PR #2436, opened 2022-04-07
Ruby 3.1.1

peteryates:
### Trello card

N/A

### Context

Ruby 2.7 is quite old now, upgrading to 3.1 makes sense for various reasons (longer support window, performance, some syntactic niceties)

- Upgrade to Ruby 3.1.1
- Fix the failing tests caused by the Ruby upgrade
- Update rubocop-govuk to the latest versions
- Fix some Rubocop offences


ethax-ross (approved):
Looks good 👍

PR #2442, opened 2022-04-11
Migrate non-production secrets to Azure keyvault

ethax-ross:
### Trello card

[Trello-1660](https://trello.com/c/eRpJcktw/1660-migrate-all-secrets-from-rails-credentials-to-the-azure-keyvaults)

### Context

We currently store our secrets in a mixture of Rails credentials and Azure keyvaults; we want to store them all in the same place and Azure keyvaults is more secure, so we are migrating them to there.

- Migrate rolling/preprod secrets to Azure

Azure keyvault has been updated with the Rails credentials and I've checked they all also map to ENV variables rather than Rails credentials.

- Remove userresearch environment/secrets

As far as I can tell the user research environment in Cloud Foundry just uses the `test` Azure keyvault, which sets `RAILS_ENV=preprod` so the `userresearch` environment is never actually used.


peteryates (approved):
Looks good

PR #2456, opened 2022-04-13
Fix accessibility issue on events list

ethax-ross:
### Trello card

[Trello-3127](https://trello.com/c/5sTO2dNB/3127-investigate-accessibility-issue-with-lists-on-returning-from-overseas-page)

### Context

Silktide is flagging the events list as a navigation component and therefore should be in `ul/li` tags. I think its probably inferred this from the class name (which contains `list`) and the fact it's multiple adjacent anchor tags. This feels a bit like a false positive but its worth fixing just to placate Silktide.

- Fix accessibility issue with events listing


peteryates (approved):
Makes sense.

PR #2462, opened 2022-04-14
Fix mobile usability issues

ethax-ross:
### Trello card

[Trello-3139](https://trello.com/c/vrtGDHcJ/3139-investigate-mobile-usability-issues-flagged-in-google-search-console)

### Context

We are seeing errors on the Google Mobile Usability checker. Strangely I can't replicate them; if I boot up a review instance running master and check it then it passes. We think its likely due to the breadcrumb and/or mobile menu button text/clickable area being small on mobile, so increasing those may resolve the issues.

- Increase breadcrumb clickable area on mobile

The breadcrumb font size and clickable area on mobile is too small; increasing it to improve usability.

- Tweak font-size of menu button to match list items

The font-size of the menu button is marginally smaller than the menu items on the expanded navigation; this tweaks it to match.

### DIFF
--- a/app/webpacker/styles/components/breadcrumb.scss
+++ b/app/webpacker/styles/components/breadcrumb.scss
@@ -8,8 +8,8 @@
       display: flex;
       flex-wrap: wrap;
       box-sizing: border-box;
-      padding: .5rem $indent-amount;
-      margin: 2em 0 0;
+      padding: 0.4em $indent-amount;
+      margin: 2em 0 0 -1em;
       list-style: none;
 
       * {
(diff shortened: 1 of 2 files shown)


gemmadallmandfe:
Thanks @ethax-ross - looks OK to me (if a little spaced out on mobile - but I guess that is the point!)

cc'ing @gazcarr for info / to have a quick look

gemmadallmandfe:
As @gazcarr is on leave this week, happy for this to be merged to production @ethax-ross so we can start getting GSC to validate the fix; an increasing number of pages with mobile usability issues are being flagged at the moment

@gazcarr this breadcrumb change is something for you to be aware of (and possibly tweak) as we implement the new nav pages

PR #2463, opened 2022-04-14
Transition to search box on click

ethax-ross:
### Trello card

[Trello-3140](https://trello.com/c/l1nMZXW7/3140-investigate-cumulative-layout-shift-cls-issues-on-desktop-core-web-vitals)

### Context

We have an issue whereby the JS search autocomplete library we use causes Cumulative Layout Shift. I've tried various techniques to fix this, such as ensuring enough space is available for the input that gets added before it happens, but nothing seems to work.

Instead, we can display a search button initially and only pull in the autocomplete input after the user clicks it; it doesn't add any extra clicks from the user (they would have to click into the input to search anyway and we auto-focus it) and it should hopefully eliminate the CLS.

- Transition to search box on click

### Guidance to review

I've also made the search input wider, which I think works better with this method as its not imposing until its opened and now you get clearer search results with more horizontal room.

Remove the co-locating of the search label next to the input to see if Silktide complains or not (if it doesn't complain then the aria-label is sufficient on the input, if it does then we need to add this back in).

Weirdly in my testing this resolves the CLS locally for lighthouse and dev tools performance section, but in the review app I still see a CLS issue 🤷‍♂️


gemmadallmandfe:
This looks fine I think @ethax-ross (although would be interested in the views of @gazcarr as well).

However, whilst testing on my iPhone 11, I have found an existing problem with Search in production.

The site fits the width of the screen perfectly fine - until the search box is accessed. As soon as that happens, the site width expands to beyond the width of the screen, resulting in horizontal scrolling.


https://user-images.githubusercontent.com/87642394/163403581-8cc0b11d-7660-41f2-85ad-2d47e00ba41c.MP4

gemmadallmandfe:
> Have we overcomplicated the search by making it inline/dynamic?
> 
> Would the user experience be worse if it was a link to a search page?

That is a good question. I actually quite like the search page for non-JS users: https://getintoteaching.education.gov.uk/search?search%5Bsearch%5D=teacher+training+adviser

I think the main question for me is how many people are using the search in the first place - and what for.

I'm going to raise a spike ticket around search - we could look at whether to get rid of the inline/dynamic functionality at that point?

gemmadallmandfe:
@gazcarr Could you have a look at this PR and see what you think of Ross's changes to the search box, to try and address a Cumulative Layout Shift that Google has flagged as a negative page experience?

@ethax-ross I ran Webaim/Wave on the review app and it flagged a possible small accessibility issue re a form label:

image

gazcarr:
I prefer the search box not to be there to start, however, I'd prefer the interaction/reveal to be vertical rather than horizontal. Mainly because this is how the majority of other sites and I'd argue this is more expected behavior from a user perspective.

I am not sure if this would also fix the Cumulative Layout Shift issue as well?

gazcarr:
Here is the original design:
https://xd.adobe.com/view/00383228-3203-43f6-8b85-fca6e3ab12fc-556b/

I prefer this to link off to a search page if I'm being honest. Search is meant to be quick and easy - there aren't many other sites that link you off to a seperate page.

gemmadallmandfe:
Thanks @gazcarr 

@ethax-ross Is it possible to reveal the search bar vertically rather than horizontally as @gazcarr suggests?

@gazcarr Noted re the separate search page. I've created a spike ticket in the backlog to look at search in more detail (https://trello.com/c/oCeUGiyP) - starting with how many people actually use it.

However, as far as this PR is concerned, the focus is just on trying to improve our on-page experience according to Google's Core Web Vitals 😱

gemmadallmandfe:
@ethax-ross @gazcarr I like the vertical reveal and bigger search field

However, I do have some questions!

* the white text on pink bar fails WCAG 2.1 AA colour contrast from looking at WebAim/Wave

* I'm a bit worried that we are essentially promoting search over Register your interest by adding the pink, without knowing how many people use search - would be interested in @valentine-carter's views on highlighting search more prominently

* I quite like that the extra nav bar (Register your interest / Search) is now not hidden in the hamburger menu on mobile - but is it a bit disjointed visually?

gazcarr:
Thanks @gemmadallmandfe 

My opinion is we remove the word 'search' which solves the first two points.

On the last one, on live we currently bundle search and register into the menu button on a mobile - I'd like to sort that with this piece of work so that we are no longer doing that. So the demo you see here has brought them back out of the menu button and back to the top of the page.

I'd also like to make the menu button pink (if accessibility allows) just to meet the style guide recommendations

valentine-carter:
As far as promoting the search bar more goes I would have to see that data because I will confess to a slight bias again internal search. This looks good to me though.

gemmadallmandfe:
This looks good to me @ethax-ross - have tested on iPhone 11 too and works well (I like having a wider search box - easier to use)

@gazcarr Are you happy with this? Suggest we address your menu button colour/position in a separate PR so that we can get this change shipped asap and I can get Search Console to revalidate the Core Web Vitals issues

gazcarr:
I'm ok with this, for now, I think there are a couple of minor tweaks to be made to icons etc. The menu button is a separate PR I guess. For a temporary fix though, this is good to go.
@gemmadallmandfe @ethax-ross

peteryates (approved):
Nice. Looks better and works better imho

PR #2464, opened 2022-04-14
Omit pages that are noindexed from site search results

ethax-ross:
### Trello card

[Trello-3146](https://trello.com/c/BSTGiZKI/3146-add-noindex-as-filter-on-internal-search)

### Context

- Omit pages that are noindexed from site search results

If a page has frontmatter with `noindex: true` we don't allow Google to index it, however we do still show it in the search results. We should ensure pages that are noindexed should also be omitted from our local site search.

### Guidance to review

You can test this by searching for `test` - no results should appear.

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -72,6 +72,6 @@ private
   end
 
   def searchable_pages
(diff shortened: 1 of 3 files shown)


peteryates (approved):
Looks good

PR #2470, opened 2022-04-19
removing top CTA on TTA page

EllieNodder:
### Trello card

https://trello.com/c/fnhcIjKT/3128-update-teacher-training-adviser-page-for-further-quantitative-test-on-context


valentine-carter:
Yes, I agree there's no need for an inline link :)

PR #2477, opened 2022-04-27
Add workflow to bust docker cache

NickGraham101:
### Context
Docker build layers are cached to optimise build performance but sometimes the base image is updated so we need to rebuild it.

Add a build-no-cache work that is scheduled once a week on Sunday and can be run manually as required.

### Guidance to review
Check the new workflow runs as expected., example [here](https://github.com/DFE-Digital/get-into-teaching-app/actions/runs/2232395955) which published image [here](https://github.com/DFE-Digital/get-into-teaching-app/pkgs/container/get-into-teaching-frontend/20741658?tag=master)

### DIFF
--- a/.github/workflows/build-no-cache.yml
+++ b/.github/workflows/build-no-cache.yml
@@ -1,0 +1,76 @@
+name: Rebuild master docker image
+on:
+  workflow_dispatch:
+
+  schedule:
+    - cron: '0 12 * * 0'
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


NickGraham101:
[Add workflow to bust docker cache](https://trello.com/c/HwLOMQ9D/546-add-workflow-to-bust-docker-cache)

saliceti (approved):

PR #2478, opened 2022-04-27
Add an aria-label to the primary nav

peteryates:
### Trello card

https://trello.com/c/H4N1jvTT/3130-build-new-navigation-pages

### Context and changes

When we soon add the category pages to the site we'll be introducing a second level of navigation and have two `<nav>` elements on the page.

To make the site easier to use for people with assistive technology we should label them separately so screen readers will call them by distinct names.

The main one in the header will be announced as 'Primary navigation'.

Screenshot from 2022-04-27 11-21-38


peteryates:
@gemmadallmandfe - we could probably do with labelling some of our other landmarks and sections better too.

gemmadallmandfe (approved):
Looks good to me

gemmadallmandfe:
I've raised a separate ticket on the backlog to cover a review of landmarks more generally: https://trello.com/c/mTd3SLZe

PR #2495, opened 2022-05-03
Ensure postcodes are not sent to Google Analytics

ethax-ross:
### Trello card

[Trello-416](https://trello.com/c/YxfD4ndn/416-investigate-alternatives-to-passing-postcode-in-urls)

### Context

A UK postcode is considered Personally Identifiable Information (PII) and therefore should not be sent to external services such as Google Analytics or output in URL/made available in the browser history.

We want to avoid this happening by encrypting postcodes before they are rendered in the URL of a GET request.

- Add simple Encryptor lib

Add a simple `Encryptor` library that can encrypt/decrypt string values. We are using [Hashids](https://hashids.org/) which is intended for obfuscating your database ids when used in URLs and results in a short hash when given a postcode (mapped to character integer values).

This isn't going to be as secure as something like `ActiveSupport::MessageEncryptor` but I think for the purpose of obfuscating PII it should be sufficient and results in a much shorter hash.

The solution should be easily transferrable to other services as we need to do similar in other projects.

- Add concern to encrypt ActiveModel attributes

We plan on encrypting the attributes of a form model when rendering the GET request that contains PII in the URL query string. In order to do this we need to be able to query the encrypted attributes of a model (to construct the URL) and then be able to instantiate the same model using the encrypted attributes (decrypting the fields into the model attributes).

The usage will be along the lines of:

```
model = Model.new(attributes_from_post_request)
url = model_path(model.encrypted_attributes)
redirect_to url
model = Model.new_decrypt(attributes_from_get_request)
```

- Encrypt postcode in event search URL

Encrypt the postcode in the event search URL so as not to expose it to Google Analytics/the browser history.

In order to do this we include the `EncryptedAttributes` concern into the `Events::Search` model and change the form method from `get` to `post`. This ensures the postcode is transmitted securely to the controller action, whereby we can encrypt it and redirect to the `GET` search endpoint, decrypt it and display the relevant results to the user. This method ensures the postcode is not leaked to the URL and yet the search results URL can be copied/bookmarked.

Removes a redundant route for event archives.

- Encrypt postcode in teaching event search URL

Ditto with respect to the previous commit.

### Guidance to review

I've given the various search forms (events, event categories and teaching events) a good click about with various filter combinations and it seems fine, but it would be worth others doing some exploratory testing.

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -5,3 +5,4 @@ PREVIEW_PAGES=1
 VCAP_APPLICATION="{\"application_name\": \"app-name\",\"space_name\": \"space-name\",\"organization_name\": \"org-name\"}"
 GTM_ID=123-ABC
 CF_INSTANCE_INDEX=app-instance
+SECRET_KEY_BASE=d144e21816f984ea2adf984d3eadc8f93e8c71f5c05368127365b13cb840ea6e0d6822a77f3342b0913b0522dd47ea4b80a794ee2da0433f4b075337650d1170
--- a/Gemfile
+++ b/Gemfile
@@ -44,6 +44,8 @@ gem "faraday-encoding"
 gem "faraday-http-cache"
 gem "faraday_middleware"
 
+gem "hashids"
+
 gem "dotenv-rails", ">= 2.7.6"
 
 gem "govuk_design_system_formbuilder", ">= 2.8.0"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -247,6 +247,7 @@ GEM
       activesupport (>= 6.0)
       deep_merge (~> 1.2.1)
     hashdiff (1.0.1)
+    hashids (1.0.6)
     html_tokenizer (0.0.7)
     htmlentities (4.3.4)
     httpclient (2.8.3)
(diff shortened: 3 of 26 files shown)


peteryates (approved):
Nice, works well.

gemmadallmandfe:
Have just started testing this @ethax-ross - can see it working on /events (although will do some more testing) :-)

Needs to be applied to the new event promo component as well though? 
Eg at the bottom of https://review-get-into-teaching-app-2495.london.cloudapps.digital/train-to-be-a-teacher/if-you-have-a-degree

ethax-ross:
@gemmadallmandfe I've updated that page so it correctly encodes the postcode now

gemmadallmandfe:
@ethax-ross I tested this in the promo module on https://review-get-into-teaching-app-2495.london.cloudapps.digital/train-to-be-a-teacher/if-you-have-a-degree and got the following error:

image

gemmadallmandfe:
Sorry - didn't mean to close that!

The postcode search worked fine on the /events page though, using the same postcode

gemmadallmandfe:
I've retested this and am still getting the 422 error @ethax-ross :-(

peteryates (approved):
Working nicely for me!

PR #2499, opened 2022-05-03
Remove empty <aside> element when not fullwidth

peteryates:
The empty aside element is pushing content to the right with the new simplified flex ordering. It's no longer needed so let's drop it.


ethax-ross (approved):
Looks good

PR #2502, opened 2022-05-04
Fixes for mobile usability issues

ethax-ross:
- Increase mobile/tablet font-size

Ensures the minimum font-size of paragraphs will be 16px, which by all indications is what Google checks for when testing if text is too small to read.

- Add bottom margin to list items

We are seeing a 'clickable items too close together' usability issue; we think this may be from where we have lists of links. Adding a small bottom-margin to the list items should hopefully improve their usability.

- Prevent table headings wrapping

Table headings were wrapping mid-word and on mobile the headings were touching; this prevents word wrapping and adds a little horizontal padding.

### DIFF
--- a/app/webpacker/styles/font-sizes.scss
+++ b/app/webpacker/styles/font-sizes.scss
@@ -1,6 +1,6 @@
 html {
-  @include mq($until: mobile) { font-size: 90%; }
-  @include mq($from: mobile, $until: tablet) { font-size: 95%; }
+  @include mq($until: mobile) { font-size: 106%; }
+  @include mq($from: mobile, $until: tablet) { font-size: 106%; }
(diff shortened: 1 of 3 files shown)


gemmadallmandfe:
List items (padding under bullets) looks fine to me @ethax-ross 

Font size increase on mobile/tablet - definitely feels easier to read for me on my iPhone :-D 

However, I think we might need to do a couple of other tweaks - for example, the column headings on the Funding table now wrap on an iPhone 11 (should we address via content or dev?)

Is there an easy way to identify other issues like this?

cc'ing @gazcarr for info

gemmadallmandfe (approved):
Looks good to me - I've had a look around and can't find any other issues

Happy for this to be merged once @peteryates has had a look too

PR #2509, opened 2022-05-04
Make flex shrink to prevent horiozontal scroll

peteryates:
### Trello card

https://trello.com/c/Z6YZFaz3/3187-fix-horizontal-shift-on-new-navigation-pages-on-mobile

### Context

The promo flexbox was set to `1 0` rather than `0 1` so it was maintaining its width rather than shrinking. Also tidied up the CSS a little.


ethax-ross (approved):
Looks good

PR #2515, opened 2022-05-09
Convert steps to be an ordered list

ethax-ross:
### Trello card

[Trello-3222](https://trello.com/c/dh1lZwnR/3222-dac-audit-no-list)

### Context

This was picked up on the DAC audit; they recommend this as an ordered list rather than sections.

- Convert steps to be an ordered list


peteryates (approved):
Looks good, definitely makes more sense semantically

PR #2521, opened 2022-05-11
Make pagination component accessible

ethax-ross:
### Trello card

[Trello-429](https://trello.com/c/QHcrgFN3/429-dac-audit-current-page-location)

### Context

Screen readers only read out the page number when they get to the pagination component which doesn't provide enough context. Instead, we should prefix the number with "Page" or "Current page".

- Make pagination component accessible


peteryates (approved):
Ah nice - didn't ever occur to me that this would be problematic 🤦🏽‍♂️

PR #2523, opened 2022-05-11
Ensure dialog cannot be accessed by screen readers when dismissed

ethax-ross:
### Trello card

[Trello-3232](https://trello.com/c/JmquPZQn/3232-dac-audit-bleed-through)

### Context

Currently the dialog is 'hidden' by pushing it out of the browser viewport. This hides it from the page but not from screen readers, which are still able to access it via tabbing (even though the `tabindex` is `-1` - I couldn't replicate the issue but it was reported by DAC so may be another screen reader).

To fix this we can ensure the element is hidden with `display: none;`.

- Ensure dialog cannot be accessed by screen readers when dismissed

### DIFF
--- a/app/webpacker/styles/dialog.scss
+++ b/app/webpacker/styles/dialog.scss
@@ -8,7 +8,7 @@
 }
 
 .dialog {
-  display: flex;
+  display: none;
   position: fixed;
   left: 0;
   top: 0;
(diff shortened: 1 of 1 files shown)


gemmadallmandfe (commented):
I can replicate the original issue using VoiceOver on my iPhone 11.

I've tested the fix and that seems to work fine.

(NB the tab order in the footer is a bit screwy on mobile - the social media links appear visually below the Home / Steps etc links but receive focus first as per the order they are in on desktop. Will raise a separate ticket for this.)

Will leave @peteryates to approve this one from a tech perspective!

peteryates (approved):

PR #2532, opened 2022-05-12
Ensure links have sufficient contrast on focus

ethax-ross:
### Trello card

[Trello-3230](https://trello.com/c/8fA6EVmX/3230-dac-audit-colour-contrast-text)

### Context

Our links aren't standard across the site and fail on contrast checks in places.

- Fix chat online button focus state

I'm not sure why the `color: white` was necessary here originally but its not any longer and it was overriding the focus state
causing white text on yellow background which is a contrast issue.

Remove specific text color styling to use defaults (remains white but has black on orange for focus state).

- Standardise white link focus states

We have white links op dark backgrounds in a number of places; mostly with custom CSS rules that don't correctly set the focus state to use black text (on an yellow background).

This aims to somewhat standardise the links through the `git-link` class. 

- Tweak focus dark blue color for contrast

We have a dark blue we use on some links to ensure sufficient contrast with the grey background. DAC picked up that the
contrast is still lacking when used in a focus state on a yellow background.

Darken the blue color to meet AAA accessibility contrast ratios.

### Guidance to review

More could definitely be done to clean up the links in general but I'll leave that for another day; this is primarily to fix the accessibility issues identified by DAC.

### DIFF
--- a/app/components/footer/mailing_list_component.html.erb
+++ b/app/components/footer/mailing_list_component.html.erb
@@ -5,7 +5,7 @@
     </div>
     <div class="mailing-list-bar__right">
       <a id="mailing-list-bar-accept" href="/mailinglist/signup/name" class="mailing-list-bar__sign-up button button--white button--nowrap" data-action="click->mailing-list#beginMailingListJourney">Sign up here</a>
-      <a id="mailing-list-bar-dismiss" href="javascript:void(0)" class="mailing-list-bar__close" data-action="click->mailing-list#dismiss" role="button">Not now</a>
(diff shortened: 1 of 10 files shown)


gemmadallmandfe:
@ethax-ross I've had a look at the Chat online button in a couple of places - all looks good.

The darker blue text on grey also passes WCAG 2.1 AA now.

However, I'll leave @peteryates to approve next week, given that there are more file updates than I understand!

peteryates (approved):
Much nicer.

PR #2535, opened 2022-05-12
Ensure a pointer cursor is used on all buttons

ethax-ross:
### Trello card

[Trello-3242](https://trello.com/c/wbt0nJbu/3242-dac-audit-missing-mouse-pointer)

### Context

DAC suggest that a pointer type cursor is used on all buttons to indicate to users that they are clickable.

- Ensure a pointer cursor is used on all buttons

### Guidance to review

It's worth noting that this goes against the CSS recommendations/standard, which suggests a cursor type pointer
should only be used for "clickable links". Buttons have a high affordance and therefore the pointer cursor is not necessary
(i.e. its obvious you can click on it because it looks like a button).

### DIFF
--- a/app/webpacker/styles/links-and-buttons.scss
+++ b/app/webpacker/styles/links-and-buttons.scss
@@ -41,6 +41,7 @@ a {
   @include button;
   @include chevron;
   display: inline-block;
+  cursor: pointer;
 
   &:disabled {
     opacity: 0.7;
--- a/app/webpacker/styles/forms.scss
+++ b/app/webpacker/styles/forms.scss
@@ -21,6 +21,10 @@ input {
   padding-right: .4em;
 }
 
+button, input[type=submit] {
+  cursor: pointer;
+}
+
 form {
   .govuk-label,
   h3.govuk-fieldset__heading {


app/webpacker/styles/links-and-buttons.scss
```diff
@@ -41,6 +41,7 @@ a {
   @include button;
   @include chevron;
   display: inline-block;
+  cursor: pointer;
```
peteryates, line 44: Not so sure about this bit. A link in a green box is still a link.

peteryates (approved):
Accepted but I'm a bit uneasy about using the pointer cursor for link-style buttons.

app/webpacker/styles/links-and-buttons.scss
ethax-ross: According to the CSS spec a link is when we _should_ use a pointer type cursor; I don't think this line is actually necessary as we do this already (I added it to be explicit, but anchor tags default to pointer cursors). It's when we have a non-link button that DAC tell us to go against the standards and also use a pointer type cursor 🤷‍♂️

gemmadallmandfe:
@peteryates @ethax-ross Let's discuss before this is merged - we need to be sure that this is the right thing

app/webpacker/styles/links-and-buttons.scss
peteryates: Ah sorry - I got muddled between pointer and default (arrow).

I think it's because when it's a link people have extra options - middle click to open in a new tab, right click to bookmark, drag to desktop etc. With a button/`input[type='submit']` they don't, but if we use the same cursor there's no visual distinction.

peteryates:
Having done a bit more reading, including [this post](https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b) from Adam Silver, I think we shouldn't use the pointer on `<button>` and `<input type="submit">` elements.

peteryates (changes requested):
Yeah, I think we should ignore DAC on this one. Buttons definitely shouldn't have pointers - do we have any more context on why they suggested it?

peteryates (approved):
Looks good 👆🏽

gemmadallmandfe:
Adding comment for audit trail - having reviewed other GOV.UK services plus our own TTA funnel and Get School Experience, we have agreed that a pointer should be used for consistency on buttons in the funnels

PR #2542, opened 2022-05-19
Hide mail icon from screen readers

ethax-ross:
### Trello card

[Trello-3260](https://trello.com/c/KpZgHkAh/3260-dac-audit-decorative-css-images)

### Context

Currently a screen reader will focus on the icon and read out 'bullet' (as its a font-based icon). This isn't very helpful and there is associated link text that gets read out so we should just be hiding this from screen readers.

- Hide mail icon from screen readers


peteryates (approved):
👍🏽

PR #2545, opened 2022-05-23
adding apprenticeships info

EllieNodder:
### Trello card

https://trello.com/c/vZrOtNgH/3228-add-apprenticeships-to-if-you-have-a-degree-page

### Context

Difficult to direct users somewhere given there's no apprenticeships page (apart from the [UCAS page](https://www.ucas.com/teaching-option/postgraduate-teaching-apprenticeship)). Not sure if this suggestion is okay, or doesn't really work? Open to suggestions @gemmadallmandfe


gemmadallmandfe:
Thanks @EllieNodder! We should definitely point people to Find - whether or not we mention the filter is probably more a question for @valentine-carter.

Once this has been resolved either way, it would be good to run this past Phoebe (or Sul) and then get it deployed.

However, I don't think we should worry too much given that this page will change in the near future re routes rewrite.

valentine-carter:
I think the bit about the filter works well and will be useful for users. I'm happy that this can now go to policy.

EllieNodder:
@gemmadallmandfe @valentine-carter seen by Sul who's happy with the changes 👍

gemmadallmandfe (approved):
Happy for this to be merged

PR #2547, opened 2022-05-23
Allow blog posts to be published without an image

peteryates:
### Trello card

https://trello.com/c/tc8OmGZG/3272-allow-blog-posts-to-be-published-without-a-main-image

### Context

The only place where an image is _required_ is when generating the structured data that's used by search engines. Now, if there's nothing there the image will be set to `nil` and omitted from the output.

I added a little bit more spacing beneath the header so the headers on posts without images don't blend into the body.

@gemmadallmandfe - here's how a post on the index looks without an image. I think it looks fine, if we have a bunch of no-image ones all together, them having the same fallback might look funny.

Screenshot from 2022-05-23 16-05-52


gemmadallmandfe:
Thanks @peteryates !

I agree - that looks fine without a thumbnail image (you've changed my mind on that one).

The only issue is the structured data (which incidently is not in production yet) - given that blogs can then be presented in carousels, not having an image might be a bit limiting

I'd suggest we address that when we bring the structured data ticket in (I'll put it in the next sprint, once the My story merge has been completed) - we may need a default image for that at least?

@valentine-carter Are you happy with this change re your My Story merging?

peteryates:
We can always merge this PR into @valentine-carter's and see how it looks with some 'real' posts.

gemmadallmandfe:
@peteryates If it is easier to push this one out first rather than combine with @valentine-carter's PR, I'd suggest we use `/blog/jills-story-getting-back-into-the-classroom` as the test post

This is a My story case study that Ellie moved into a blog post a while back at the request of the returners team - I'd like to remove the main image from it as it is such poor quality

If you can add that to this PR, we can review/approve/merge prior to Valentine's changes

PR #2550, opened 2022-05-24
Breadcrumb structure updates

ethax-ross:
### Trello card

[Trello-3263](https://trello.com/c/kGXvHq8S/3263-tweak-header-to-address-skip-to-main-content-and-breadcrumb-issues)

### Context

The way the current hero/header/breadcrumb sections are structured causes accessibility issues. We want to move the breadcrumbs up above the hero section and then make the hero section part of the main content; that way the 'skip to main content' link will work as expected.

We also want the breadcrumbs to be styled consistently across the website, making use of the GOV.UK breadcrumb component and its ability to collapse on mobile.

Finally, we believe the 'home' link in the navigation is now redundant (given proximity to the breadcrumb 'home' link and logo that can be used to go to the home page) so we want to remove it.

- Restyle breadcrumbs above hero section
- Remove home link from main navigation
- Restructure header and hero sections

### Guidance to review

| Desktop (no hero)      | Desktop (hero) | Mobile (no hero)      | Mobile (hero) |
| ----------- | ----------- | ----------- | ----------- |
| Screenshot 2022-05-24 at 14 46 27      | Screenshot 2022-05-24 at 14 47 11       |  Screenshot 2022-05-24 at 14 48 04   |  Screenshot 2022-05-24 at 14 47 54 |


gazcarr:
@ethax-ross all good from a design perspective 👍

peteryates (approved):
Much nicer. Glad that dangling 'Home' is gone

gemmadallmandfe:
My view @ethax-ross @peteryates @gazcarr is that if you have a breadcrumb on some pages, you need it on all, otherwise it could be confusing to users.

I've tried to find WCAG guidance on this and only come up with this:
https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-location.html

image

However, I do agree that it looks a bit odd when it is just 'Home' on the second level pages!

The alternative would be to include the current page in the breadcrumb as well, which Scope / RNIB do eg

image

**But** this contradicts the guidance in the design system:
https://design-system.service.gov.uk/components/breadcrumbs/

So my view is that we leave as is. As part of this nav work, we will be moving orphan pages around, so fewer pages will just have 'Home' on them anyway. We can always review again at that point

PR #2551, opened 2022-05-24
Add border to buttons for high contrast mode

ethax-ross:
### Trello card

[Trello-3254](https://trello.com/c/getLeIBt/3254-dac-audit-inverted-hit-area)

### Context

When in Windows High Contrast mode the buttons are indistinguishable from the background. To get around this we can add a transparent border that will differentiate the component in High Contrast Mode.

- Add border to buttons for high contrast mode

### Guidance to review

We use a media query to apply this only when high contrast is enabled as it results in a minor misalignment of buttons due to the 2px border width.

Screenshot (7)

Screenshot (8)

### DIFF
--- a/app/webpacker/styles/header/logo.scss
+++ b/app/webpacker/styles/header/logo.scss
@@ -21,6 +21,10 @@
   display: block;
   overflow-x: clip;
 
+  @media (forced-colors: active) {
+    forced-color-adjust: none;
+  }
+
   a {
     text-decoration: none;
     display: block;
--- a/app/webpacker/styles/links-and-buttons.scss
+++ b/app/webpacker/styles/links-and-buttons.scss
@@ -8,6 +8,10 @@
   @include font-size("small");
   border: none;
 
+  @media (forced-colors: active) {
+    border: 2px solid transparent;
+  }
+
   &:hover {
     background: darken($bg, 5%);
     color: $fg;


gemmadallmandfe:
This looks much better to me @ethax-ross although I'll leave @peteryates to look too

My only comment is the logo - particularly if we are removing Home from the top nav. Not sure what the best approach is with this?

image

gemmadallmandfe:
I've tested in WHC mode and the logo doesn't change now

image

I don't know whether this strictly meets accessibility or whether we are supposed to have an alternate contrasting logo (like GOV.UK). However, for the purposes of addressing this PR / the issue flagged in the audit, I think this is an improvement.

Tagging @peteryates to review from a tech perspective and for any accessibility insight re the logo!

peteryates (approved):
Overall I think this looks good. Left a comment, `outline` might be useful here as it doesn't occupy space.

I think there are a few other bits we could improve on high contrast mode - for example there's no border/outline around the mailing list bar. Perhaps worth a separate ticket @gemmadallmandfe?

ethax-ross:
@peteryates good suggestion with outline; I'm not sure if that will work the same way as border in high contrast mode though and I can't test it at the moment as I don't have access to a Windows computer 😞 it sounds like we should raise a separate ticket to tackle the logo and some other areas so I'll merge this for now and we can try an outline as part of that.

PR #2553, opened 2022-05-25
Revert changes to postcode step

ethax-ross:
### Trello card

[Trello-2995](https://trello.com/c/1jPnQFRQ/2995-revisit-the-postcode-step-in-the-mailing-list-funnel)

### Context

We found that the new postcode step design (that places the postcode input behind a yes/no question) performed less well than our original design.

Revert back to the original design with minor copy updates.

- Revert changes to postcode step

### Guidance to review

Screenshot 2022-05-25 at 09 54 22

### DIFF
--- a/app/models/mailing_list/steps/postcode.rb
+++ b/app/models/mailing_list/steps/postcode.rb
@@ -2,26 +2,15 @@ module MailingList
   module Steps
     class Postcode < ::DFEWizard::Step
       attribute :address_postcode
-      attribute :send_event_emails, :boolean
 
-      validates :send_event_emails, inclusion: { in: [true, false] }
       validates :address_postcode, postcode: true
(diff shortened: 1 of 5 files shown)


gemmadallmandfe:
Looks good to me

@valentine-carter are you happy with this? Copy feels slightly different from that on the other steps but this is the only optional field, so I think that is probably OK for now

(Next step would be to split test making this mandatory)

valentine-carter:
@gemmadallmandfe I like this copy (I am biased) and I'd be interested to see if it's more effective to have a slightly more personable voice.

gemmadallmandfe (approved):
Happy for this to be merged (should we wait until Monday morning?)

ethax-ross:
Yeah I'll leave this until Monday now 👍

PR #2556, opened 2022-05-25
Imageless blog posts

peteryates:
### Trello card

https://trello.com/c/tc8OmGZG/3272-allow-blog-posts-to-be-published-without-a-main-image

### Context

The only place where an image is _required_ is when generating the structured data that's used by search engines. Now, if there's nothing there the image will be set to `nil` and omitted from the output.

I added a little bit more spacing beneath the header so the headers on posts without images don't blend into the body.

@gemmadallmandfe - here's how a post on the index looks without an image. I think it looks fine, if we have a bunch of no-image ones all together, them having the same fallback might look funny.

Screenshot from 2022-05-23 16-05-52

Replaced #2547


peteryates:
@gemmadallmandfe - fixed in #2558

gemmadallmandfe (approved):
Looks good to me - @valentine-carter happy with this as well so can be merged

PR #2559, opened 2022-05-26
Standardise page headings across the site

ethax-ross:
### Trello card

[Trello-3271](https://trello.com/c/Yj3hKNXU/3271-standardise-heading-styles-across-the-site)

### Context

Currently we style the core `h1`, `h2`, etc tags to cover the majority of typical scenarios where they are used. We often find our headings deviate from the core styles on a number of page types (markdown, longform, custom, etc). This results in overriding the core classes in multiple places, often with tweaks to font sizes, margins and in some cases the background
highlighting/font colours.

Due to this approach it's difficult to know how many heading styles we have and where/when they should be used. Instead, we want to take the GOV.UK approach of base styles for the core heading tags and classes that can be used to cleanly override a heading style.

- Standardise headings across the site

This commit aims to clear that up with a dedicated `headings.scss` that abstracts the styling of each heading type into classes as well as applying defaults to the core `h1-4` tags. It means if we find a scenario where we are using a `h3` and need it styled like a `h2` then we can simply apply the appropriate class.

Also includes a couple of helper classes to remove margins/change font weights/apply an overlap layout. If we stick to using these it should be easier to make the site consistent.

- Remove TypeDescriptionComponent and styles

This is left over from when we had event type descriptions summarised at the top of the events page; they were later moved to be in each event group on the index page so this is now redundant.

- Add view component preview for typography

Makes it easier to visualise the different headings/select the one you need.

### Guidance to review

[Component preview](https://review-get-into-teaching-app-2559.london.cloudapps.digital/rails/view_components/content/typography/default)

screencapture-0-0-0-0-3000-rails-view-components-content-typography-default-2022-05-26-12_38_07

The welcome guide and subject specific layouts are outliers and have their own bespoke heading/caption styles going on. I'd like to bring them inline with the new heading classes as well but opted to tackle those in a separate PR.

> ⚠️ this needs a thorough click-about to check I haven't missed/broken headings anywhere for desktop and mobile viewports. There will be minor changes with production in places to achieve better consistency across the site.


peteryates (dismissed):
CSS wise this is **much nicer**. I've done a sweep through the pages and nothing looks far out, there are a few minor differences in sizes/spacings but it looks great overall.

gemmadallmandfe:
@ethax-ross Nice work!

I've looked through all of the pages in the sitemap (except for all of the My story ones) plus all of the non-event ones that are indexed by Search Console but not in the sitemap

The only things that look slightly odd are:

* last two h2s (Further information and Contact) on https://review-get-into-teaching-app-2559.london.cloudapps.digital/international-qualified-teacher-status (although Valentine is currently working on this page anyway so this is probably a red herring)

* Find promo on https://review-get-into-teaching-app-2559.london.cloudapps.digital/train-to-be-a-teacher/if-you-dont-have-a-degree (Note that there are two Find promos - the other one is fine; this one that goes to UCAS needs to be included in these changes as well)

image

gazcarr:
Other than what Gemma has spotted I think this seems to be good

peteryates (approved):
Looks great!

PR #2562, opened 2022-05-27
Add current page to breadcrumb component

ethax-ross:
### Trello card

[Trello-3302](https://trello.com/c/ttjy7K6l/3302-redesign-breadcrumb-to-show-current-page)

### Context

Our website structure is largely flat, which means we often have a sole 'Home' link at the top of the page. It looks out of place and it isn't clear that its a breadcrumb, so we want to always include the current page (not as a link) to make sure there's at least two breadcrumbs in the component at all times.

- Add current page to breadcrumb component

### Guidance to review

The collapse on mobile behaviour has been dropped (otherwise the user is left with just a home link and the text for the current page, given its the last breadcrumb).

### DIFF
--- a/app/components/header/breadcrumb_component.html.erb
+++ b/app/components/header/breadcrumb_component.html.erb
@@ -1,10 +1,13 @@
 <section class="container breadcrumbs">
-  <nav class="govuk-breadcrumbs govuk-breadcrumbs--collapse-on-mobile" aria-label="breadcrumbs" role="navigation">
+  <nav class="govuk-breadcrumbs" aria-label="breadcrumbs" role="navigation">
     <ol class="govuk-breadcrumbs__list">
       <% helpers.breadcrumb_trail do |crumb| %>
-        <% next if crumb.current? %>
         <li class="govuk-breadcrumbs__list-item">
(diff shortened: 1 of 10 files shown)


gazcarr:
@ethax-ross happy with the styling, when you land on the blog page, there is no 'blog' in the breadcrumb like there are on other top-level pages

ethax-ross:
@gazcarr well spotted I've pushed the fix

gemmadallmandfe:
Looks good @ethax-ross !

Couple of things from me:

1) As @gazcarr says, this doesn't work on the Blog page - nor on the main Events page - presumably because of the way these pages work. Is it is possible to make them consistent?

2) The current page in the breadcrumb looks like it is picking up `title`. However, for some pages, the title is different from the `heading` for SEO purposes. Given that the `heading` is the one that appears on-site and is the one that users will see, I think we should use that instead unless there are good reasons not to?

eg `/train-to-be-a-teacher/if-you-have-a-degree`
Breadcrumb has 'Ways to train if you have a degree' but on page and navigating to it via the parent page, the user would know it as 'If you have a degree'
image

gemmadallmandfe (approved):
Looks good to me if @peteryates agrees!

peteryates (approved):

peteryates:
Looks way better!

PR #2565, opened 2022-05-30
Improve contrast on button/link focus states

ethax-ross:
### Trello card

[Trello-3300](https://trello.com/c/t730tosw/3300-review-link-and-button-states-across-the-site)

### Context

Our buttons and links are accessible in _most_ use cases, but we have several edge cases where they are used on a darker/lighter background and are not passing colour contrast checks.

Update buttons/link focus states to use a black background with purple outline to ensure they can always be clearly differentiated from surrounding content/backgrounds.

- Improve contrast on button/link focus states

Cleans up the CSS around links and buttons.


ethax-ross:
@gazcarr thoughts on this ^ ?

gazcarr (changes requested):
make outline 2px

peteryates (approved):
Looks good, outline's definitely easier to spot now.

gazcarr:
The chat to us on the "funding your training" page has a black link colour on the CTA

image

ethax-ross:
Well spotted @gazcarr I've updated that now

gemmadallmandfe:
Sorry for the delay in looking at this.

Firstly - top job in terms of consolidating these! Definitely the right thing to do.

However, I have to say that I find the focus state of white font on black background with pink outline quite hard to read in a lot of cases - and particularly on mobile. I think it is because the text goes right up to the edge of the black for inline links and then there are the distracting lines as well.

image

image

image

Even the top nav is hard work on my eyes!

image

It's not so bad for the buttons because the text doesn't go up to the edges:

image

@gazcarr Can we discuss? I don't think it is quite right as is

**Missing focus states**
@ethax-ross I've spotted the following

1) Related content eg on `/train-to-be-a-teacher/if-you-have-a-degree`

image

2) Accordion arrows eg on `/non-uk-teachers/international-qualified-teacher-status`

image

gemmadallmandfe:
Focus state looks easier to read to me - what do you think @ethax-ross @peteryates @gazcarr ?

The link colour on grey also looks better to me.

I've only had a brief look around - will have more of a clickaround tomorrow

peteryates:
Rebased it so it didn't clash on the now-removed `latest_event_component_spec.rb`

peteryates (dismissed):
Looks good to me, focus appear to be consistent (other than the forms which keep their GOV.UK Design System yellow)

gazcarr:
Gemma has spotted a few issues... 

Generally thought the design seems to work well IMO and I think we've found a good balance here

gemmadallmandfe:
I've had another click around @ethax-ross and this is looking pretty good.

The only additional oddity I have found (sorry!) is the Back link in the funnels eg 
image

Not entirely sure whether this is a different GDS state deliberately.

Otherwise happy for this to be be deployed tomorrow (Tues)!

gemmadallmandfe (approved):
Thanks @ethax-ross - I _think_ this is good to go now!

gemmadallmandfe (approved):

PR #2570, opened 2022-05-31
Add landmark roles to improve document structure

ethax-ross:
### Trello card

[Trello-3309](https://trello.com/c/mTd3SLZe/3309-review-labelling-of-landmarks-and-sections)

### Context

As we use semantic HTML elements in the majority of places adding the `role` attribute isn't strictly necessary but it is advisable:

> It is also good practice to include the role redundantly with the associated semantic element.

Add `role` landmark for banner, complementary, main, contentinfo, navigation and search.

I've avoided adding `region` landmarks as these will be inferred from where we use `section` elements as long as a heading exists and there are a lot of them so it would be a maintenance burden IMO.

- Add landmark roles to improve document structure

### Guidance to review

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/landmark_role

### DIFF
--- a/app/components/footer_component.html.erb
+++ b/app/components/footer_component.html.erb
@@ -13,7 +13,7 @@
 <%= render Footer::VideoPlayerComponent.new %>
 <%= render Footer::CookieAcceptanceComponent.new %>
 
-<footer class="site-footer">
+<footer class="site-footer" role="contentinfo">
   <div class="site-footer__wrapper limit-content-width">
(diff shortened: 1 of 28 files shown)


peteryates (approved):
Nice. I've browsed around a few pages and the page structure is definitely more apparent.

👌🏽 

Screenshot from 2022-06-01 14-49-18

PR #2585, opened 2022-06-01
Remove space between autocomplete result items

peteryates:
### Context and changes

The margin-bottom on accessible-autocomplete was pushing our autocomplete items apart when they should have been flush.

Thanks to @mylesjarvis for the heads up!

| Before | After |
| ------ | ------- |
| Screenshot from 2022-06-01 11-58-22 | Screenshot from 2022-06-01 11-58-44 |


ethax-ross (approved):
Looks good!

PR #2592, opened 2022-06-06
Fix broken link checker Slack notifications

ethax-ross:
### Trello card

[Trello-3310](https://trello.com/c/MiLrSzmd/3310-fix-slack-notifications-for-broken-link-checker)

### Context

The `lychee` action we use appears to have recently changed so that the `exit_code` is only written via `exit` so that a subsequent action can read it if the `fail` argument is `true`.

In doing this I still wasn't able to read `exit_code` directly, but as the step now fails on a non-zero exit code we can use `continue-on-error: true` and then check the `outcome` of the step to achieve the same result.

- Fix broken link checker Slack notifications

### Guidance to review

I've tested this manually and it posted to our tech channel.

### DIFF
--- a/.github/workflows/lychee.yml
+++ b/.github/workflows/lychee.yml
@@ -15,10 +15,12 @@ jobs:
       - name: Link Checker
         id: lychee
         uses: lycheeverse/lychee-action@v1.5.0
+        continue-on-error: true
         env:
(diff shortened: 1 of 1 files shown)


peteryates (approved):
Looks good. Sorry this was my fault for letting the update (which looked very minor!) through 😞

PR #2595, opened 2022-06-06
Change the photo on non-UK teachers page

peteryates:
### Context and changes

The yellow background of the hero has a funny effect on the photo that makes it look much yellower than it does in isolation.

| Before | After |
| ------- | ---------|
| Screenshot from 2022-06-06 14-54-26 | Screenshot from 2022-06-06 14-54-31 |


gazcarr:
agreed @peteryates 

All good with me @gemmadallmandfe

gazcarr (approved):

PR #2599, opened 2022-06-07
Embedding a GOV.UK section for School Experience

peteryates:
### Trello card

https://trello.com/c/WHNPHSff/3276-design-embedding-govuk-start-pages-into-git

### Context and changes

Currently going from GIT to GOV.UK-style pages (like School Experince) can be a bit jarring for users. This is an attempt to ease the transition by showing that they're about to go to a more-official looking site.

- Split up get-school-experience page into partials
- Make SimpleComponent more flexible
- Add a 'GOV.UK' like embedded section for SE

| Desktop | Mobile |
| -------- | --------- |
| Screenshot from 2022-06-07 12-02-25 | Screenshot from 2022-06-07 12-02-37 |

### Limitations

The GOV.UK Transport font is not used here because getting it to play nicely with our styles is would involve lots of additional work. We'd need to load it separately (so all users aren't impacted by the font download) and style things individually because the Design System is very all-or-nothing. If we think it's needed I'm happy to revisit, but in my opinion it looks ok as-is.

### DIFF
--- a/app/components/calls_to_action/simple_component.html.erb
+++ b/app/components/calls_to_action/simple_component.html.erb
@@ -6,14 +6,16 @@
       <%= tag.span(title, class: "call-to-action__heading") %>
     <% end %>
 
-    <% if text.present? %>
+    <% if content.present? || text.present? %>
       <p class="call-to-action__text">
-        <%= helpers.safe_html_format(text) %>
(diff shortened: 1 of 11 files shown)


peteryates:
Closing for now as it's rather out of date.

PR #2600, opened 2022-06-07
Update logo alt text for accessibility

ethax-ross:
### Trello card

[Trello-3343](https://trello.com/c/d1Iw9gNG/3343-dac-audit-label-in-name-logo?filter=member:rossoliver15)

### Context

It was picked up in the DAC report that the logo alt text is "Go to home page" even though the image has the text "Teaching
Every Lesson Shapes a Life". The user using voice assist expected "Click teaching every lesson shapes a life" to take
them to the homepage. The recommendation was to add alt text that covers both scenarios.

Update the alt text to include "Teaching every lesson shapes a life" and "Go to the Get Into Teaching homeage" (the former
being what DAC expect and the latter being what other GOV.UK services appear to use).

- Update logo alt text for accessibility

### Guidance to review

I couldn't replicate the DAC report for this; using 'Tap teaching every lesson shapes a life' on mobile safari doesn't work (but 'tap <visible link>' does). I also couldn't replicate it in Chrome with various voice assist extensions, so we're just going by the suggested solution in the DAC report (that having both separated by a pipe will work).

### DIFF
--- a/app/components/header/logo_component.html.erb
+++ b/app/components/header/logo_component.html.erb
@@ -1,7 +1,7 @@
 <div class="logo-wrapper">
   <div class="logo">
     <a href="/" class="logo__image">
-      <%= image_pack_tag 'media/images/getintoteachinglogo.svg', alt: "Go to Home page", size: "180x62", data: { "lazy-disable": true } %>
+      <%= image_pack_tag 'media/images/getintoteachinglogo.svg', alt: "Go to the Get Into Teaching homepage, Teaching Every Lesson Shapes a Life logo", size: "180x62", data: { "lazy-disable": true } %>
     </a>
   </div>
 </div>


peteryates (dismissed):
Looks good. Not sure about the `|` character in the alt though. When screen readers read it will they say 'pipe'? Maybe a comma here instead?

gemmadallmandfe:
Agree that 'Go to the Get into teaching homepage' appears sensible and in keeping with GOV.UK @ethax-ross 

I have tested this on my iPhone using Voice Control and can't get it to work for `Tap teaching every lesson shapes a life` or `Tap Go to home page`. However, I can get `Tap Home` to work as it picks up the breadcrumb link (which is more prominent now at the top of the page).

In terms of the alt text, VoiceOver on iPhone is already trying to read the text on the logo so it now says `Teaching every lesson shapes a life' twice!

https://user-images.githubusercontent.com/87642394/172449893-c89ebf6c-02e2-45fd-af2f-07a4005b3e72.MP4

The pipe in the alt text is read out as `vertical line`

My thoughts are:

* Keep the 'Go to the Get into Teaching home page' as this is in line with GOV.UK and improves what we have at the moment
* If we add the 'Teaching every lesson shapes a life logo` in the alt text, I'd suggest replacing the pipe with a comma (or full-stop? does that work?) as @peteryates suggests

I think that we are probably not going to come up with a perfect solution for all voice assistants but this feels like a small improvement?

ethax-ross:
@gemmadallmandfe updated to replace pipe with a comma

PR #2601, opened 2022-06-07
Move the details descriptions to an expandable box

peteryates:
### Trello card

https://trello.com/c/ENQrIOP1/2897-improve-events-page-redesign-of-where-definitions-should-be-placed

### Context

The details descriptions at the top of the teaching events list take up a lot of room and are quite ugly. This reformats them so they take up more vertical space but places them inside a `<details>` element so they're hidden by default.

|            | Desktop | Mobile |
| -------- | -------- | ----------|
| Collapsed | Screenshot from 2022-06-07 13-54-03 | Screenshot from 2022-06-07 13-54-20 |
| Expanded | Screenshot from 2022-06-07 13-54-07 | Screenshot from 2022-06-07 13-54-25 |

### DIFF
--- a/app/views/teaching_events/index.html.erb
+++ b/app/views/teaching_events/index.html.erb
@@ -6,12 +6,12 @@
   <header>
     <h1 class="heading-l heading--box-yellow">Get Into Teaching events</h1>
 
+    <p>
+      All events listed here are designed to help you get into teacher training. They are all free to attend.
+    </p>
+
     <details class="teaching-events__descriptions">
       <summary class="teaching-events__descriptions--summary">What are the event types?</summary>
-
-      <p>
-        All events listed here are designed to help you get into teacher training. They are all free to attend.
-      </p>
 
       <dl>
         <div>
--- a/app/views/teaching_events/index.html.erb
+++ b/app/views/teaching_events/index.html.erb
@@ -6,26 +6,30 @@
   <header>
     <h1 class="heading-l heading--box-yellow">Get Into Teaching events</h1>
 
-    <p>
-      All events listed here are designed to help you get into teacher training. They are all free to attend.
-    </p>
+    <details class="teaching-events__descriptions">
+      <summary class="teaching-events__descriptions--summary">What are the event types?</summary>
 
-    <dl>
-      <div>
-        <dt>DfE Train to Teach events</dt>
-        <dd>Watch presentations, hear from teachers, speak to advisers and meet local training providers.</dd>
-      </div>
+      <p>
+        All events listed here are designed to help you get into teacher training. They are all free to attend.
+      </p>
 
-      <div>
-        <dt>DfE Online Q&amp;As</dt>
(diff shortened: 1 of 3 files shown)


gemmadallmandfe:
This looks good to me.

@gazcarr @EllieNodder Are you happy with this?

Two copy suggestions/questions from me @EllieNodder:

* Suggest we reduce the first line slightly as this will help get it onto two lines rather than three on (some) mobiles:
`All events listed ~~here~~ are designed to help you get into teacher training. They are ~~all~~ free to attend.`

* Is the wording 'What are the event types' right? Should this be slightly more descriptive or is it OK?

EllieNodder:
> This looks good to me.
> 
> @gazcarr @EllieNodder Are you happy with this?
> 
> Two copy suggestions/questions from me @EllieNodder:
> 
> * Suggest we reduce the first line slightly as this will help get it onto two lines rather than three on (some) mobiles:
>   `All events listed are designed to help you get into teacher training. They are free to attend.`
>   ie removing 'here' and the second 'all'
> * Is the wording 'What are the event types' right? Should this be slightly more descriptive or is it OK?

Maybe 'Explore the different types of events' instead of 'What are the event types?' to make it more active? 'Event types' sounds a bit off to me. And agree with getting rid of 'all' and 'here'.

Also, should it be 'All events listed are designed to help you get into **teaching**' rather than 'teacher training'? What about returners? (Does that also mean the nav label shouldn't 'teacher training events', too? 😬)

app/views/teaching_events/index.html.erb
```diff
@@ -10,22 +10,26 @@
       All events listed here are designed to help you get into teacher training. They are all free to attend.
     </p>
 
-    <dl>
-      <div>
-        <dt>DfE Train to Teach events</dt>
-        <dd>Watch presentations, hear from teachers, speak to advisers and meet local training providers.</dd>
-      </div>
+    <details class="teaching-events__descriptions">
+      <summary class="teaching-events__descriptions--summary">What are the event types?</summary>
```
gemmadallmandfe, line 14:
```suggestion
      <summary class="teaching-events__descriptions--summary">Explore the different types of events</summary>
```

app/views/teaching_events/index.html.erb
```diff
@@ -10,22 +10,26 @@
       All events listed here are designed to help you get into teacher training. They are all free to attend.
```
gemmadallmandfe, line 10:
```suggestion
      All events listed are designed to help you get into teacher training. They are free to attend.
```

gemmadallmandfe:
Change the opening para wording and also the accordion wording as suggested by @EllieNodder 

I will leave the 'teacher training' references as is for now - the nav label will be disappearing anyway and the teacher training events make up the vast majority of those listed.

gemmadallmandfe (approved):
Looks good to me - happy for this to be merged @peteryates if it you are happy with it

PR #2602, opened 2022-06-07
Make the Scribble box wider

peteryates:
### Trello card

https://trello.com/c/hC7dfoiJ/2829-address-width-issue-re-scribble-confirmation-message

### Context and changes

Scribble is an embeddable 'event' iframe placed on the page. When it's in the article element which occupies 66% of the width it gets cramped and isn't very responsive.

Here we're moving it outside of the `<article>` element and placing it below so it can occupy 100% of the width. A small margin is applied so it will line up with the content above.

### Testing

Tricky to test locally as there are no events but it should now appear where this "lorem ipsum" paragraph is 😅 

Screenshot from 2022-06-07 14-19-03

### DIFF
--- a/app/views/teaching_events/show.html.erb
+++ b/app/views/teaching_events/show.html.erb
@@ -22,7 +22,6 @@
     <article>
       <%= render(partial: "teaching_events/show/event-summary") %>
       <%= render(partial: "teaching_events/show/attending-training-providers") %>
-      <%= render(partial: "teaching_events/show/scribble") if @event.scribble_id.present? %>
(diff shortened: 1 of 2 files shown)


gemmadallmandfe:
Given that `\teaching-events` is already in production but does not have (much) traffic going to it, I'm happy for this to be merged and check in prod that it works, as long as @ethax-ross is happy from a tech eyes perspective!

gemmadallmandfe:
@ethax-ross Do you want to have a quick look at this from a tech perspsective? Can then be merged as far as I'm concerned!

ethax-ross (approved):
LGTM

PR #2603, opened 2022-06-07
Make the teaching-event quote flush with footer

peteryates:
### Trello card

N/A

### Context

The bottom of the teaching event page looks a bit odd. This change removes the forced `margin-bottom` from the `.main-content` so it's flush. The `margin-bottom` of the bottom bits of most pages means it looks fine. It does look a bit broken on the stories page but that will soon be gone so shouldn't be a worry.

* [ ] Remove stories pages before merging this

| Before | After |
| ------ | ------- |
| Screenshot from 2022-06-07 14-55-39 | Screenshot from 2022-06-07 14-55-23 |

### DIFF
--- a/app/views/layouts/teaching_event.html.erb
+++ b/app/views/layouts/teaching_event.html.erb
@@ -9,7 +9,7 @@
     <main id="main-content" role="main">
       <%= render Content::HeroComponent.new(@front_matter) %>
 
-      <section class="main-body">
+      <section class="main-body main-body--no-bottom-margin">
         <%= yield %>
       </section>
     </main>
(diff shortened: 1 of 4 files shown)


gemmadallmandfe:
@peteryates Slightly in two minds about this - definitely looks better on desktop but on mobile the stacking means that the Talk to us looks like it is very much associated with the quote

image

Would be interested in @gazcarr's thoughts!

gazcarr:
@peteryates @gemmadallmandfe 

Can we add a bit of extra padding (double) at the bottom of the yellow section?

peteryates:
Done, added a bit more padding on both desktop and mobile:

| Desktop | Mobile |
| -------- | -------- |
| Screenshot from 2022-06-08 15-15-50 |  Screenshot from 2022-06-08 15-19-25 |

gemmadallmandfe:
Looks better to me - are you happy with this @gazcarr ?

PR #2609, opened 2022-06-09
Log when we perform a redirect

ethax-ross:
### Trello card

[Trello-3349](https://trello.com/c/5g14tyJ3/3349-look-into-how-many-redirects-are-still-used-which-ones-we-can-get-rid-of)

### Context

We want to be able to track which of our custom redirects are still being used (so that we can remove them when we find no traffic going through them any longer).

Update redirect route to log out the details, which we can then pull into a Kibana visualisation.

- Log when we perform a redirect

### Guidance to review

Initially I was going to do this as a metric but as it would likely end up with low cardinality and infrequent data points it didn't feel like a good fit. Kibana also has a longer retention period than our metrics so should prove to be a more reliable data store.


peteryates (approved):
Looks good

PR #2616, opened 2022-06-13
Tweak padding on highlighted headings

ethax-ross:
### Trello card

[Trello-3348](https://trello.com/c/Ulmo5Ytv/3348-small-bug-fix-on-blog-headings)

### Context

These headings have a highlight/background over multi-line text; we're finding on some browsers the highlighting of the next line overlaps the text of the previous. Tweaking the padding to be slightly less should improve this.

- Tweak padding on highlighted headings

### DIFF
--- a/app/webpacker/styles/headings.scss
+++ b/app/webpacker/styles/headings.scss
@@ -15,7 +15,7 @@
 
   span {
     box-decoration-break: clone;
-    padding: 0.15em 0;
+    padding: 0.1em 0;
     background: $backgroundColor;
     box-shadow: 1rem 0 0 $backgroundColor, -1rem 0 0 $backgroundColor;
   }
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
Looks good to me - however, I can't replicate the original issue so will leave for @peteryates to test as flagged the original problem

peteryates:
I'm still seeing it. Maybe it's just a Firefox thing.

Here's Chrome on the left, Firefox on the right:

Screenshot from 2022-06-13 16-19-47

gemmadallmandfe:
Given that this now appears to relate to a very small subset of users (see Trello ticket) and the page is still readable for those users, I'm happy for this to be closed @ethax-ross

ethax-ross:
Closing as this is only effecting Linux users, which make up a very small portion of our traffic and the text is still readable even with the slight visual glitch.

PR #2619, opened 2022-06-13
Require an address city for all buildings

ethax-ross:
> :warning: pending sign-off from events team before we merge

### Trello card

[Trello-2963](https://trello.com/c/l0Yw3U53/2963-investigate-why-some-provider-events-are-missing-location-in-new-events-listing-design)

### Context

We want to ensure all buildings are entered with an address city as we use this on the listings page to indicate where the event is (even if its an online event as it may be location-relevant).

- Require a city address for all buildings
- Fix GOV.UK JS not loading in internal event pages

### Guidance to review

It looks like the JS needs to be evaluated after the page has loaded and it hasn't been working on the internal events pages. Switch to using the `frontend.js` pack that only pulls in the parts we actually use (radio buttons in the internal events form).

### DIFF
--- a/app/models/internal/event_building.rb
+++ b/app/models/internal/event_building.rb
@@ -17,7 +17,7 @@ module Internal
     validates :address_line1, length: { maximum: 255 }
     validates :address_line2, length: { maximum: 255 }
     validates :address_line3, length: { maximum: 255 }
-    validates :address_city, length: { maximum: 100 }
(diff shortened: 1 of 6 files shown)


gemmadallmandfe:
Closing this PR - locations have been added in the CRM by the Events team and the front end provider form changed to make City mandatory

PR #2613, opened 2022-06-10
Replace events with teaching events

peteryates:
### Trello card

https://trello.com/c/BqqVPfpe/3332-deploy-new-events-design-to-production

### Context

Started making the move, bulk is 'done' but it needs:

* [x] the remains of old event functionality removing
* [x] remove noindex from new event pages
* [x] remove canonical tags from new event pages
* [x] track down any last remnants of old events views (this is done in #2620)
* [x] update backlinks / copy round the site
* [ ] some final checks

PR #2620, opened 2022-06-14
Cleaning up after events and stories

peteryates:
**This follows on from #2613**

### Trello card

https://trello.com/c/MNnVEaB0/3361-do-some-post-event-and-story-removal-tidying

### Context and changes

General cleaning up of things we no longer use

- Remove story component
- Remove latest event component
- Remove renderer component
- Remove events category
- Change the index path in events integration spec
- Remove story component from CTA injector
- Add a wildcard redirect for /event-categories
- Remove event_categories paths
- Remove the #latest_event_for_category method
- Remove remaining references to stories from specs


config/redirects.yml
```diff
@@ -410,3 +410,4 @@ redirects:
   "/why-teach(*reason)": "/blog"
   "/content/stories(*story)": "/blog"
   "/my-story-into-teaching(*story)": "/blog"
+  "/event-categories/(*category)": "/events"
```
gemmadallmandfe, line 413: @peteryates As per my comments on the Events deployment PR, this doesn't work - we have a TTT page that we want people to go to! Plus I think the others should probably go to pre-filtered lists. Happy to discuss

gemmadallmandfe: @peteryates This is covered by https://github.com/DFE-Digital/get-into-teaching-app/pull/2613 now so I think needs removing here?

ethax-ross (approved):
Looks good to me 🗑️ 🔥

PR #2622, opened 2022-06-15
Whitelist blog tags with a centralised config file

ethax-ross:
### Trello card

[Trello-3164](https://trello.com/c/ikHA33tm/3164-control-the-creation-of-blog-tags-centrally-to-enable-better-curation-of-blog-posts)

### Context

We want to stop new tags being added ad-hoc to blog posts; instead, they must be part of the centralised `tags.yml` file before they can be used.

If a new tag is used without being declared we will raise an exception, which will be flagged by our existing content specs that ensure all content on the website can be rendered successfully.

If a blog post has an invalid tag (such that its not declared in the whitelist of available tags) we want to raise this as a problem. The difficulty here is that its often going to be a content editor that runs into this, so whilst we want a test to fail to prevent a deployment we also want to make it clear to content editors that there is an issue/what the issue is (and delving into the CI to find out wouldn't be a great experience for them).

To achieve this, we can surface the content errors on the `rolling` environment (which is used for review apps) within the page being edited. The tests will still fail to prevent a deployment and the content editor will be able to resolve the issue without developer assistance.

- Raise an error when an invalid blog tag is used
- Surface content errors on rolling
- Update content documentation around blog tags

### Guidance to review

I've made this a somewhat generic solution with a view that we could potentially use `ContentError` elsewhere (to prevent large image files from being added, for example).

I opted to render the errors inline with the article being edited, but if we expand this it may make sense to `rescue_from` in the `ApplicationController` and render a generic 'there are content errors on this page' template (to avoid rescuing in individual actions).

Screenshot 2022-06-15 at 11 22 30

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -32,6 +32,11 @@ protected
     @noindex = true
   end
 
+  def add_content_error(error)
+    @content_errors ||= []
+    @content_errors << error
+  end
+
 private
 
   def declare_frontmatter
--- a/app/controllers/blog_controller.rb
+++ b/app/controllers/blog_controller.rb
@@ -14,6 +14,14 @@ class BlogController < ApplicationController
   def show
     @post = ::Pages::Blog.find(request.path)
 
+    begin
+      @post.validate!
+    rescue ::Pages::ContentError => e
+      raise e unless helpers.display_content_errors?
+
+      add_content_error(e)
+    end
+
     breadcrumb "Blog", blog_index_path
     breadcrumb @post.title, request.path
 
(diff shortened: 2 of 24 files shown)


peteryates (commented):
I like it - this'll help reduce the number of tags where we have closely-related ones too

peteryates (approved):

gemmadallmandfe (approved):
Looks good to me - I have had a look around the front end and blog posts/tags/tag pages all look fine, plus the content guidance makes sense.

Beyond that, I can't test but happy that Pete has had tech eyes on it!

PR #2629, opened 2022-06-17
Make event search params more analytics friendly

peteryates:
### Trello card

https://trello.com/c/doE2bUqu/2701-set-up-redirects-to-make-pre-filtered-event-links-easier-to-use

### Context

This is the first step in making the event search query string easier to read. Here we're replacing the Dynamics-generated IDs (eg 222_750_008) with a readable string:

#### Before

```
&teaching_events_search[type][]=222_750_008&teaching_events_search[type][]=222_750_009
```

#### After

```
&teaching_events_search[type][]=online_event&teaching_events_search[type][]=school_or_university_event
```

The query string is complex because we're binding the search form to an object (which makes it easier for us to show errors, etc), but in doing so it means we're using Rails' conventions for query params.

We could simplify more so it's more like:

```
?event_types=online_event,school_or_university_event
```

But we'd need to make bigger changes to the way the filter and form works, re-implementing a lot of what Rails gives us from scratch.

- Add lookup_by_query_params methods to EventType
- Change the search over to use named event params

### DIFF
--- a/app/models/event_type.rb
+++ b/app/models/event_type.rb
@@ -9,10 +9,10 @@ class EventType
 
   QUERY_PARAM_NAMES =
     {
-      "train_to_teach_event" => 222_750_001,
-      "question_time" => 222_750_007,
-      "online_event" => 222_750_008,
-      "school_or_university_event" => 222_750_009,
+      "ttt" => 222_750_001,       # Train to Teach event
+      "tttqt" => 222_750_007,     # Question Time
+      "onlineqa" => 222_750_008,  # Online event
+      "provider" => 222_750_009,  # School or University event
     }.freeze
 
   attr_accessor :type_id
--- a/app/models/event_type.rb
+++ b/app/models/event_type.rb
@@ -7,6 +7,14 @@ class EventType
       "School or University event" => 222_750_009,
     }.freeze
 
+  QUERY_PARAM_NAMES =
+    {
+      "train_to_teach_event" => 222_750_001,
+      "question_time" => 222_750_007,
+      "online_event" => 222_750_008,
+      "school_or_university_event" => 222_750_009,
+    }.freeze
+
   attr_accessor :type_id
 
   delegate(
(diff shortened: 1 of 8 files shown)


app/models/event_type.rb
```diff
@@ -7,6 +7,14 @@ class EventType
       "School or University event" => 222_750_009,
     }.freeze
 
+  QUERY_PARAM_NAMES =
```
gemmadallmandfe, line 10: Are these param names taken from the CRM?

I'm asking because

a) I don't think they are really as useful as they could be eg 'online_event' is misleading as it is only Online Q&A in reality but could be read as any online event

b) they are longer than needed which increases URL length therefore making it harder to read in GA and harder to get right in links in emails etc.

If we have flexibility, I would perhaps suggest something like

      "ttt" => 222_750_001,
      "tttqt" => 222_750_007,
      "onlineqa" => 222_750_008,
      "provider" => 222_750_009,
      
I've deliberately included 'ttt' in the question time events as it then is easy to query all ttt events in GA.

NB event types are likely to change in the next two terms anyway, but good to get some principles in place!

peteryates: The ones above ("School or University event" etc) _are_ from the CRM, but there's no reason for the `QUERY_PARAM_NAMES` to follow suit (other than consistency). I'll change them and add a comment so it's clear what's what

ethax-ross (approved):
Looks good. Who are we doing this for though? If its for us/DfE then I can see how this is helpful, but if its for users/external entities then I'm not sure `tttqt` says much more than `222_750_007` 😅

gemmadallmandfe:
> Looks good. Who are we doing this for though? If its for us/DfE then I can see how this is helpful, but if its for users/external entities then I'm not sure `tttqt` says much more than `222_750_007` 😅

Ha! My view was that we were doing it for us:
* to help with GA analysis
* to make it easier to link to pre-filtered results eg from GIT emails

However, I'm happy for you to disagree @ethax-ross !

PR #2633, opened 2022-06-21
Add original request to redirect logs

ethax-ross:
### Trello card

[Trello-3414](https://trello.com/c/sdcYUyVi/3414-investigate-high-traffic-to-legacy-urls)

### Context

We are seeing a large number of redirects for certain patterns; to help debug where the traffic is coming from we want to log the original request URL as well as the redirect information.

- Add original request to redirect logs


peteryates (approved):
Makes sense to me

PR #2635, opened 2022-06-22
Fix the sitemap for the new event structure

peteryates:
### Trello card

https://trello.com/c/cCltcNPa/3418-add-core-events-pages-to-the-sitemap

#### Remove event category pages from sitemap

The old event categories page no longer exist and have been removed. In their place the 'about-train-to-teach-events' page has been added.

#### Remove trailing slashes from events and blog

It appears that Google isn't too worried about the presence (or absence) of the trailing slash but it does treat both variations as separate pages.

As we link to `/events` and `/blog` in most places (via the `events_path` and `blog_path` helpers) it makes sense to drop them from `OTHER_PATHS`.

### DIFF
--- a/app/controllers/sitemap_controller.rb
+++ b/app/controllers/sitemap_controller.rb
@@ -11,9 +11,7 @@ class SitemapController < ApplicationController
   OTHER_PATHS = %w[
     /events/
     /blog/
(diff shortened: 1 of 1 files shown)


ethax-ross (approved):
Looks good to me 👍

PR #2637, opened 2022-06-24
Move online event instructions to conditional

peteryates:
### Trello card

https://trello.com/c/jElVr4WU

### Context and changes

When events aren't online we shouldn't be displaying the guidance on log-in instructions and hardware requirements.

Also reword the opening sentence a little.

### DIFF
--- a/app/presenters/teaching_events/event_presenter.rb
+++ b/app/presenters/teaching_events/event_presenter.rb
@@ -49,6 +49,10 @@ module TeachingEvents
       event_building.present?
     end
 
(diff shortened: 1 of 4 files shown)


gemmadallmandfe:
I've looked at the events in the review app and the in-person TTT events look good. However, we don't have any virtual TTT events in the dev CRM so I can't test fully.

I'll leave it to @ethax-ross to review from a tech perspective - then happy for this to be merged

ethax-ross (approved):

PR #2641, opened 2022-06-28
Migrate post-processing logic to middleware

ethax-ross:
> ⚠️ WIP

### Trello card

[Trello-3431](https://trello.com/c/KoMHb8Wb/3431-ensure-that-links-from-404-page-take-the-user-to-the-top-of-the-destination-page?filter=member:rossoliver15)

### Context

We currently leverage an `after_action` to apply post-processing to HTML responses (for images and links). This works for almost all pages, however if we `rescue_from` an exception and render a response this appears to be skipped; I can't figure out why and no other controller lifecycle callbacks seem to catch it.

Migrate the post processing logic into Rack middleware so that it will apply to all responses. Clean up the external link logic and extract it to its own library.

- Migrate post-processing logic to middleware

### Guidance to review

Needs a good click about in case there are any obscure pages/responses this blows up for.


lib/middleware/html_response_transformer.rb
```diff
@@ -0,0 +1,40 @@
...
+      end
+
+      [status, headers, response]
+    end
+
+    def transform(response_body)
+      response_body = NextGenImages.new(response_body).html
+      response_body = ResponsiveImages.new(response_body).html
+      response_body = LazyLoadImages.new(response_body).html
+      response_body = ExternalLinks.new(response_body).html
```
peteryates, line 32: Totally not one for now but it would be nice if this could be chained

```
response_body
  .with_next_gen_images
  .with_responsive_images
  .with_lazy_loading_images
  .with_accessible_external_links
  .without_closing_source_tags
  .html
```

peteryates (approved):
Looks great, **much** nicer in middleware than in a controller

Dockerfile
```diff
@@ -13,7 +13,7 @@ ENV RAILS_ENV=production \
 RUN mkdir /app
 WORKDIR /app
 
-RUN apk add --no-cache build-base tzdata shared-mime-info nodejs yarn \
+RUN apk add --no-cache build-base tzdata shared-mime-info nodejs yarn git \
```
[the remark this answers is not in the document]
ethax-ross: The odd thing is that we have other gems pulled in via `github` with a `ref` that were fine 🤷‍♂️

lib/middleware/html_response_transformer.rb
ethax-ross: Yeah that's a nice suggestion, maybe with a decorator so we don't need to monkey patch the response 🤔

gemmadallmandfe:
Looks good to me

I tested:

* Links on 404 page now return you to the top of the page
* Logo on 404 page does not include inverted state logo as well
* Generally having a click around the site

PR #2644, opened 2022-06-30
Restrict goal completion to live/viewable events

ethax-ross:
### Trello card

[Trello-3433](https://trello.com/c/3wu6pGzV/3433-investigate-why-historic-events-are-showing-up-in-ga-goal-completions)

### Context

At the moment its possible to reach the sign up `/completed` page for an event that is in the past/closed. As we have our goal completions setup to look for requests to this path its possible for a completion event to register in UA after the event
date (if a user has bookmarked the page or has left it open and reloads).

To prevent this we can redirect to the root event path, which will render a `410 gone` response (no longer registering as a goal completion). If the event is still 'live' but registrations are closed, the user will be presented with the event and closure message.

A 'walk in' can still technically sign up for a closed event (as this happensat the time of the event, so potenitally shortly after the start_at date). We need to make sure this is still possible by retaining the `is_walk_in`attribute in the store after sign up (otherwise the completion step will not know the sign up was a walk in and redirect them). 

- Restrict goal completion to live/viewable events

- Use event factory in tests

As the controller action now inspects the date as well as the status of the event it needs to be present on our mock; we can do this easily by swapping it out to use the factory.

### Guidance to review

This PR also simplifies/changes the logic around determining if an event is 'viewable'. Previously we checked it was future dated and not pending. We should be safe to include closed in that conditional now as we no longer have a way of viewing past/archived events (previously you could view a past online Q&A event).

[Example completed page of past event](https://review-get-into-teaching-app-2644.london.cloudapps.digital/events/train-to-teach-london-event-050322/apply/completed)

### DIFF
--- a/spec/requests/event_steps_controller_spec.rb
+++ b/spec/requests/event_steps_controller_spec.rb
@@ -185,5 +185,17 @@ describe EventStepsController, type: :request do
     end
 
     it { is_expected.to have_http_status :success }
+
+    context "when the event is in the past" do
+      let(:event) { build :event_api, :past, readable_id: readable_event_id }
+
+      it { is_expected.to redirect_to(event_path(id: event.readable_id)) }
+
+      context "when the candidate is a 'walk-in'" do
+        before { get event_step_path(readable_event_id, :personal_details, walk_in: true) }
+
+        it { is_expected.to have_http_status :success }
+      end
+    end
   end
 end
--- a/app/controllers/event_steps_controller.rb
+++ b/app/controllers/event_steps_controller.rb
@@ -8,7 +8,7 @@ class EventStepsController < ApplicationController
   self.wizard_class = Events::Wizard
 
   before_action :noindex
-  before_action :redirect_closed_events, only: %i[show update]
+  before_action :restrict_sign_ups, only: %i[show update completed]
   before_action :set_step_page_title, only: [:show]
   before_action :set_completed_page_title, only: [:completed]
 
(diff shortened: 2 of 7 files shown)


gemmadallmandfe:
@ethax-ross I can see this issue in prod - I've tried accessing a completed page for a historic event (`https://getintoteaching.education.gov.uk/events/train-to-teach-national-virtual-event-220622/apply/completed`)

However, I can't test in the review app as I don't know any URLs of historic events in the dev CRM! If you can let me know one, I'll test asap tomorrow

gemmadallmandfe:
@misaka Can you have a look at this from a tech perspective so I can approve tomorrow once I've tested in the review app?

gemmadallmandfe:
Sorry @ethax-ross - I missed the link for testing!

Tested successfully

gemmadallmandfe (approved):
Looks good to me

spec/requests/event_steps_controller_spec.rb
```diff
@@ -179,11 +179,25 @@ def perform_request
   end
 
   describe "#completed" do
+    let(:path) { completed_event_steps_path(readable_event_id) }
```
misaka, line 182: Don't see aspecific reason why this is broken out into a `let`, is this a left-over from aborted changes below?

misaka (dismissed):
LGTM!

spec/requests/event_steps_controller_spec.rb
ethax-ross: Ah yeah I started doing a refactor of the specs above to consolidate the common setup but then decided against it. I've removed the let and fixed up the commit if you could re-tick it when you have a minute.

misaka (approved):

PR #2645, opened 2022-06-30
Handle when event is returned as nil

ethax-ross:
### Trello card

[Trello-3435](https://trello.com/c/CbE3Um28/3435-fix-sentry-error-on-accessing-nil-event)

### Context

When an event is not found in the API the client library is returning `nil`. I'm not sure yet when this changed as it used to throw an exception that we handle. For now, we need to handle the `nil` case and return a 404 (at the moment we get an `nil` object access error). I will look into how/when the client library changed as a separate PR.

- Handle when event is returned as nil
- Redirect /events/search to /events

The old events form redirected the POST request to GET `/events/search` with encrypted parameters (to obfuscate the postcode/PII). The new events form also does this, however we don't have a redirect in place for people that bookmarked
the `/events/search` path and we are seeing a few errors as thats now treated as a GET event (with id of `search`), which is not found.

### Guidance to review

The two commits in this PR are related so I've bundled them together (the missing redirect is causing a number of errors and flagged the need to handle a `nil` event).

### DIFF
--- a/app/controllers/teaching_events_controller.rb
+++ b/app/controllers/teaching_events_controller.rb
@@ -70,7 +70,7 @@ private
     GetIntoTeachingApiClient::TeachingEventsApi.new.get_teaching_event(params[:id]).tap do |event|
       status = EventStatus.new(event)
 
-      raise ActionController::RoutingError, "Not Found" if status.pending?
+      raise ActionController::RoutingError, "Not Found" if event.nil? || status.pending?
       raise(EventNotViewableError) unless status.viewable?
     end
   end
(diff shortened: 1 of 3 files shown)


gemmadallmandfe:
I can see that `/events/search` is currently throwing a 500 error in production - we've had 27 unique page views of `/events/search/xxx` since 16 June. I've tested the redirect and that is fine @ethax-ross. 

I think the rest of the changes need tech eyes on them so I will leave for Misha to review and approve!

misaka (approved):
💯

gemmadallmandfe (approved):
Approving this now that Misha has looked at it as well

PR #2656, opened 2022-07-01
Fix 302 redirect on root sign up path

ethax-ross:
The `dfe_wizard` gem has been updated to ensure the root redirects, for example from `/mailinglist/signup` to `/mailinglist/signup/name` are 301 (permanent) instead of 302 (temporary).


gemmadallmandfe (approved):
Looks good to me - have tested the following:

* mailing list CTA on /subjects/maths
* event register CTA on Manchester TTT event

Can see that both are 301s

PR #2659, opened 2022-07-04
removing TTT event info from promo

EllieNodder:
### Trello card

https://trello.com/c/kSUdvrTW/3442-replace-events-promo-when-summer-programme-ends

### Context

Removing Train to Teach event details from events promo now that the event has passed.


gemmadallmandfe (approved):
Looks good to me

PR #2660, opened 2022-07-05
updating nav descriptions

EllieNodder:
### Trello card

https://trello.com/c/jomu3l6g/3452-look-at-nav-card-descriptions

### Context

Currently, the nav card descriptions in the nav cards file don't match those on the content pages. This PR makes sure that they match up and that the right ones are pulling through onto the live category page.


gemmadallmandfe (approved):
Looks good to me

PR #2662, opened 2022-07-05
Add component for displaying inset text

ethax-ross:
### Trello card

[Trello-3352](https://trello.com/c/bqdPpvtD/3352-highlight-that-a-blog-post-contains-old-terminology)

### Context

Now that we have moved some of the old stories into the blog, we need to highlight that some of them have old content/terminology. To do this we will call it out at the top of the blog article in an inset component.

- Add component for displaying inset text

Add an `InsetTextComponent` that can be used to highlight a portion of text within an article. It optionally takes a title and can render safe HTML so it is possible to include inline links.

- Render inset text in markdown handler

Update the markdown handler to support rendering `InsetText` view components specified by the `inset_text` key within the frontmatter.

Clean up the component rendering logic so that we can easily whitelist new components to be rendered in the future.

- Update blog articles with inset text

We want to call out the first paragraph of these blog articles as inset text on the page to make it clearer to the user that the terminology has changed.

### Guidance to review

Screenshot 2022-07-05 at 09 10 53

### DIFF
--- a/app/components/content/inset_text_component.html.erb
+++ b/app/components/content/inset_text_component.html.erb
@@ -1,0 +1,4 @@
+<%= tag.section(class: "inset-text") do %>
+  <%= tag.h2(title) if title.present? %>
+  <%= tag.p(helpers.safe_html_format(text)) %>
+<% end %>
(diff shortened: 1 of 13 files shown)


gazcarr (approved):

misaka (approved):
🚀

gemmadallmandfe (approved):
Looks good to me - thanks for adding the content guidance @ethax-ross

PR #2677, opened 2022-07-11
adding qts and rewording copy

EllieNodder:
### Trello card

https://trello.com/c/5HOaD98u/3443-update-steps-page-to-reference-qts

### Context

Steps currently doesn't reference QTS, which isn't good for challenging the misconception users have that it's a PGCE they need to teach.

Some of the copy also currently isn't accurate.

### DIFF
--- a/app/views/content/steps-to-become-a-teacher/_step_2_check_your_qualifications.html.erb
+++ b/app/views/content/steps-to-become-a-teacher/_step_2_check_your_qualifications.html.erb
@@ -1,11 +1,13 @@
-<p>Once you have a degree or equivalent qualification you’re ready for postgraduate primary or secondary initial teacher training courses.</p>
+<p>If you have a degree or equivalent qualification, you can do postgraduate primary or secondary teacher training.</p>
 
-<p>You also need the following GCSEs at grade 4 (C) or above, or equivalent qualifications:</p>
+<p>If you do not have a degree and are not studying for one, you can do <a href="/train-to-be-a-teacher/if-you-dont-have-a-degree">undergraduate teacher training</a>.</p>
+
+<p>You need the following GCSEs at grade 4 (C) or above, or equivalent qualifications:</p>
 
 <ul>
   <li>English</li>
   <li>maths</li>
-  <li>science if you want to teach primary</li>
+  <li>science (if you want to teach primary)</li>
 </ul>
 
 <p>You may be able to show you meet the standard in another way if you do not have GCSEs.</p>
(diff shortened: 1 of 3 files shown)


EllieNodder:
@gemmadallmandfe @GDW7 have edited some of the copy so that:

- the content is more accurate and clear
- the links are accessible
- we now mention qualified teacher status (QTS)

In rewriting some of the content under step 4, I've also got rid of 'PGCE'. I think we _could_ still talk about PGCEs if we wanted, but it's not really necessary for this page. And from an SEO/keywords point of view, if people are searching 'PGCE', we want them to land on the PGCE/train to be a teacher/maybe QTS page, rather than steps, so don't see the value in stuffing it in.

But happy for you to disagree and can always add it back in!

GDW7:
Hi @EllieNodder that sounds sensible and the changes look good to me.

gemmadallmandfe (commented):
Thanks @EllieNodder - just left one minor comment. Happy for this to be deployed to prod when you've had a look at this (feel free to disagree!).

Re taking out PGCE - I think that sounds reasonable. Looking in GSC, search queries with 'pgce' don't tend to end up on Steps. I've also had a look at ad campaign data in GA and their PGCE campaigns don't point to Steps either.

I think we just need to inform the advertising team when this goes live.

app/views/content/steps-to-become-a-teacher/_step_2_check_your_qualifications.html.erb
```diff
@@ -1,11 +1,13 @@
-<p>Once you have a degree or equivalent qualification you’re ready for postgraduate primary or secondary initial teacher training courses.</p>
+<p>If you have a degree or equivalent qualification, you can do postgraduate primary or secondary teacher training.</p>
```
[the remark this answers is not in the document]
EllieNodder: Does that suggest then that you can't apply until you've finished your degree though? (Or am I overthinking!)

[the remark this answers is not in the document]
gemmadallmandfe: Yes I think you are right - let's leave as is :-)

app/views/content/steps-to-become-a-teacher/_step_2_check_your_qualifications.html.erb
```diff
@@ -1,11 +1,13 @@
...
 
-<p>You also need the following GCSEs at grade 4 (C) or above, or equivalent qualifications:</p>
+<p>If you do not have a degree and are not studying for one, you can do <a href="/train-to-be-a-teacher/if-you-dont-have-a-degree">undergraduate teacher training</a>.</p>
```
gemmadallmandfe, line 3: @EllieNodder Is this undergrad sentence in the right place? It is now in the middle of the postgrad info - is that confusing?

EllieNodder: My thinking was that the qualifications info was relevant for both postgrad and undergrad courses. And as a user, if I see 'if you have a degree', I'd then be thinking 'what about if I don't have a degree?' so thought it made sense to move that sentence further up. But equally happy to move it back down!

gemmadallmandfe: If the qualifications section also applies to undergrad degrees then let's leave it as is

gemmadallmandfe (approved):
Looks good to me

PR #2679, opened 2022-07-13
removing old funding and tips pages

EllieNodder:
### Trello card

https://trello.com/c/ARWu5Zrw/3437-unpublish-funding-your-training-and-set-up-redirect

AND

https://trello.com/c/wEvxxmbk/3419-unpublish-tips-on-applying-and-set-up-redirect

### Context

Now that our new sections/pages are live, we need to delete the old ones and set up redirects.


gemmadallmandfe:
Redirects tested - all fine @EllieNodder 

However, the advertising team are just checking re the ads as there is still some cpc traffic going to these pages (possibly through sitelink extensions). We should have an answer tomorrow so will hold off deploying for now

gemmadallmandfe (approved):
Looks good to me

PR #2683, opened 2022-07-18
Updating old funding links

EllieNodder:
### Context

The Ukraine page contains links to the old /funding-your-training page. These need to be updated to the new section so we can delete the old page.


gemmadallmandfe (approved):
Looks good to me

PR #2687, opened 2022-07-19
Add figures are for 2021/22 academic year message

GDW7:
### Context
Adding message giving context to salary figures


gemmadallmandfe (approved):
Looks good to me

PR #2711, opened 2022-08-01
Campaign landing pages for mailing list sign up

ethax-ross:
### Trello card

[Trello-3499](https://trello.com/c/aJVZws3a/3499-create-pages-for-first-campaign-landing-test)

- Add mailing list campaign hero image

We will be using this image in the hero section of the mailing list campaign.

Add the image and an entry in the config file for the alt text.

- Add hero component for campaign landing pages

The campaign landing pages for the mailing list sign up will have a restyled hero section.

Add a component for the hero section, along with a preview.

- Support campaign landing pages for mailing list sign up

The marketing team want to be able to deliver campaign landing pages that let the user sign up to the mailing list. These pages will have the same mailing list form but the first step will use a new layout and content. They want to be able to dynamically add new campaign landing pages with ease.

Add a `campaigns.yml` file that contains the page content for each campaign.

Add new routes to support campaigns in the format `/mailinglist/signup/:id/campaigns/:campaign_id`. The routes were slightly awkward to add due to the step parameter needing to be `:id` and not `:step_id` which is how Rails would like to infer the parameter name (the wizard gem uses `:id`).

Add a new layout that pulls in the `CampaignHeroComponent` and ensures the pages are not indexed.

Update the `StepsController` to check if a campaign is present; if it is we load the frontmatter from the `campaigns.yml` according to the campaign id and render the correct layout for the first step of the wizard. We also ensure subsequent steps are served under the new campaign routes.

### Guidance to review

There is a [component preview](https://review-get-into-teaching-app-2711.london.cloudapps.digital/rails/view_components/content/campaign_hero_component/with_default) for the new campaign hero component. You can view the new campaign landing page [here](https://review-get-into-teaching-app-2711.london.cloudapps.digital/mailinglist/signup/name/campaigns/overcoming-hurdles).

I've opted not to memoize the campaign data as its currently a very small file and the page appears to load quickly anyway; we can add it in later if we find it necessary.

~~I think its worth making the hero image better quality if we have the source, it looks pretty fuzzy on a wide desktop screen.~~

| Mobile      |  Tablet      |  Desktop |  Wide |
| ----------- | ----------- | ----------- | ----------- |
| screencapture-0-0-0-0-3000-mailinglist-signup-name-campaigns-overcoming-hurdles-2022-08-01-15_05_39  |  screencapture-0-0-0-0-3000-mailinglist-signup-name-campaigns-overcoming-hurdles-2022-08-01-15_08_27    | screencapture-0-0-0-0-3000-mailinglist-signup-name-campaigns-overcoming-hurdles-2022-08-01-15_04_56   |  screencapture-0-0-0-0-3000-mailinglist-signup-name-campaigns-overcoming-hurdles-2022-08-01-15_09_45  |


GayeCord:
i like the simplicity of this design, but is that the best picture we have? Unless its my new glasses, its very blurred  
image

ethax-ross:
@GayeCord I've updated the PR with the new higher resolution image

gemmadallmandfe:
@EllieNodder Can you sort out the Lorem ipsum in the meta description please :-)

app/controllers/mailing_list/steps_controller.rb
```diff
@@ -40,14 +45,24 @@ def crm_store
...
       is_first_page = @current_step.instance_of? MailingList::Steps::Name
-
+      return "registration_with_campaign" if is_first_page && campaign_id.present?
       return "registration_with_side_images" if is_first_page
 
       "registration"
     end
 
+    def load_campaign_frontmatter
+      campaigns = YAML.load_file(Rails.root.join("config/campaigns.yml"))
```
misaka, line 61: Is there any reason not to pre-load this somewhere instead of doing it for every request? Minor performance issue, but it should be easy to load this at boot time somehow.

ethax-ross: I thought about this but opted not to; put a note in the PR description -- happy to do so if you disagree, but it felt like a premature optimisation to me and it often makes the tests more complex:

> I've opted not to memoize the campaign data as its currently a very small file and the page appears to load quickly anyway; we can add it in later if we find it necessary.

misaka: Sorry, missed that note. I think that's fine, normally that file should, hopefully, remain in cache and the only cost would be a few system calls, no big deal. I guess it risks becoming a performance issue if the host is swamped anyway, and can't afford to keep that file cached, which seems rather unlikely.

misaka (dismissed):

gemmadallmandfe:
Can we just hold fire on merging this just yet - need to respond to some concerns from advertising team

GDW7:
@gemmadallmandfe I've updated the copy in the title - could you take a look and let me know what you think? Also @gazcarr

GDW7:
@gemmadallmandfe I've updated to copy in the hero - it's now referencing free support, not bursaries.

gemmadallmandfe:
Closing this PR for now - whilst we want to test campaign landing pages, we won't be able to pick this up again until early(ish) 2023 due to other priorities.

PR #2721, opened 2022-08-03
Add navigation guidance to content readme

ethax-ross:
### Trello card

[Trello-3395](https://trello.com/c/sbNuXqDi/3395-update-guidance-for-content-editors-re-navigation-changes)

### Context

We have two different types of navigation components that pages can leverage; the main navigation (at the top of a page) and the 'category' pages which have a set of cards to let the user navigate to related content.

Update the readme with guidance on how to use navigation components via frontmatter.

- Add navigation guidance to content readme

### Guidance to review

I thought about including screenshots but figured they'd be a pain to keep relevant, so we'll see if the descriptions are clear enough without them before doing it.

### DIFF
--- a/docs/content.md
+++ b/docs/content.md
@@ -14,9 +14,10 @@ This documentation aims to be a reference for content editors that want to make 
 	* [Calls to Action](#calls-to-action)
 		* [Main Content](#main-content)
 		* [Sidebar](#sidebar)
-  * [Accessibility](#accessibility)
-    * [iframe](#iframe)
-  * [Inset text](#inset-text)
+	* [Accessibility](#accessibility)
+		* [iframe](#iframe)
+	* [Inset text](#inset-text)
+	* [Hero](#hero)
 3. [Creating a Blog Post](#creating-a-blog-post)
 	* [Images](#images)
 	* [Footers](#footers)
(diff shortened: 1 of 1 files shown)


docs/content.md
```diff
@@ -318,3 +321,55 @@ closing_paragraph: my-new-closing-paragraph
...
+```
+
+In order for the pages to appear as cards on the main category page they must have navigational attributes defined in the frontmatter:
+
+```yaml
+navigation: 1
+navigation_description: A brief description that will appear in the card for this page
+```
+
+The `navigation` attribute determines the order of the cards on the page (lower values appear first/higher up the page). The `navigation_description` will appear within the card for this page below a heading (which will match the `heading` or `title` from the page frontmatter).
```
EllieNodder, line 368: I think that we can also put in a 'navigation_title:' so that the title on the card can be slightly different to the page title/heading. Worth adding that in here?

ethax-ross:
@EllieNodder updated with your suggestions, also added a bit of info for the hero section as I noticed that was missing

docs/content.md
```diff
@@ -318,3 +338,64 @@ closing_paragraph: my-new-closing-paragraph
...
+
+In order for the pages to appear as cards on the main category page they must have navigational attributes defined in the frontmatter:
+
+```yaml
+navigation: 1
+navigation_title: Title that appears on the card
+navigation_description: A brief description that will appear in the card for this page
+```
+
+The `navigation` attribute determines the order of the cards on the page (lower values appear first/higher up the page). The `navigation_description` will appear within the card for this page below a heading, which will match the `navigation_title`, `heading` or `title` from the page frontmatter (in that order).
```
EllieNodder, line 386: sorry, can I just check what you mean by 'in that order'?

ethax-ross: So we look at `navigation_title` first, if thats not set we take the `heading` and if that's not set we take the `title`. Does that make sense? I can maybe explain that clearer 🤔

EllieNodder (approved):
Looks good to me :)

PR #2724, opened 2022-08-04
Automatically add sizes to image elements

ethax-ross:
### Trello card

[Trello-3511](https://trello.com/c/uIQ6heOX/3511-address-cumulative-layout-shift-cls-issue-on-subject-template-pages)

### Context

In a number of places on the website we don't explicitly set the image size on the `img` element. This can result in Cumulative Layout Shift (CLS) as the browser can't reserve the appropriate page space for the image until its downloaded it.

We manually set image sizes where the images are static, as it's easy to do. Some of our components/views however can take a dynamic image and render it, making it hard/verbose to manually specify the image sizes everywhere.

We can use `FastImage` to quickly obtain image sizes (in under ~100ms) and then inject them into the `img` elements as part of the existing HTML response transformer. Once an image has been 'sized' the result is cached and subsequent requests will be even faster.

If we find that `FastImage` has a detremental impact on performance we can look to manually add image sizes to the `images.yml` config file, however I want to avoid this as it will be error prone (if an image is updated and the size is not, for example).

- Automatically add size to img elements
- Set image height to be automatic

On some pages we don't specify an image height so the element dimensions are used, which are far bigger than what we want. Instead, we can set the height to `auto` and it will scale the image according to the responsive width.

- Make image caching more robust

Reduce the timeout on loading images to 250ms in case the asset instance is being slow to respond (that way the knock-on effect is minimised; we cache a `nil` result so even if the asset host is down it should only effect the first size request).

Restrict image sizing to our own images; ensure we don't call out to download a potentially malicious file from an external source.

- Confiugure image sizing with env vars

Control the timeout for `FastImage` with an environment variable and add a feature flag so we can easily turn it off and on in case we have issues after shipping.

### Guidance to review

At the moment the cache will persist across requests but is not shared between instances. We could use Redis to do this, but as we only have 2 instances its probably not worth it; especially as it introduces a complexity of needing to purge the Redis keys when the image changes (so probably on every release). We get this for free when the cache is in-memory.

:warning: this change is site-wide and may result in oddly sized images on some pages (if they don't have an explicit/auto height set in CSS); we need to do a thorough check before merging.

:warning: this will be turned off in production initially so we can make sure its behaving in the preprod environment first.


misaka (approved):
🚀

PR #2726, opened 2022-08-05
Optimise blog images

ethax-ross:
### Trello card

[Trello-3510](https://trello.com/c/9AHU0ScM/3510-optimise-images-identified-in-wavemaker-seo-report)

### Context

Where possible these are resized to 1464 x 1100 pixels; most however are smaller than this so have been resized to match this aspect ratio.

- Resize blog images to better fit the template
- Include blog image guidance in content readme


gemmadallmandfe (approved):
Looks good to me - have tested on iPhone 11 too

And as well as optimising the images for SEO, controlling the crop has the added bonus of not cutting off the tops of heads in some cases!

PR #2727, opened 2022-08-05
Optimise event images

ethax-ross:
### Trello card

[Trello-3510](https://trello.com/c/9AHU0ScM/3510-optimise-images-identified-in-wavemaker-seo-report)

### Context

Resize the event images so that they're not far larger than they need to be and make sure they are in the correct aspect ratio so we're displaying the whole image.

- Optimise event images


gemmadallmandfe (approved):
Looks good to me - tested images in the review app compared with the dev environment

PR #2728, opened 2022-08-05
Remove unused partials

ethax-ross:
### Trello card

[Trello-3519](https://trello.com/c/proVQPJy/3519-stop-referencing-images-with-a-digest-directly)

### Context

These partials don't appear to be referenced anywhere and contain images with the asset digest hardcoded (which is why I was looking at them in the first place).

- Remove unused partials


gemmadallmandfe (commented):
Looks fine to me @ethax-ross but have asked @EllieNodder to have a quick look and approve as she is more knowledegable about the nav stuff than I am!

EllieNodder (approved):
looks good to me

PR #2734, opened 2022-08-10
Fix FastImage exception on timeout

ethax-ross:
The timeout will be read from the `.env` file as a String; we are seeing an error in preprod "Can't convert String to time interval", which I think is stemming from the timeout not being cast to a float. This wasn't picked up in development as FastImage loads the images from disk instead of over HTTP.

### DIFF
--- a/lib/image_sizes.rb
+++ b/lib/image_sizes.rb
@@ -25,7 +25,7 @@ class ImageSizes
     end
 
     def cache_image_size(src)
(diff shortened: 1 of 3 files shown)


misaka (approved):
I would've expected the type-casting to happen in `timeout`, presumably that should always return a float. Still good to ship, can be addressed later.

🚢

PR #2744, opened 2022-08-12
Enable image sizing in production

ethax-ross:
Turns on automatic image sizing in production; we've ran this in preprod and it's had a thorough test.


gemmadallmandfe (approved):
Happy for this to be switched on in prod

PR #2748, opened 2022-08-16
reordering events

EllieNodder:
### Trello card

https://trello.com/c/6ePVXIBv/3540-switch-round-order-of-services-on-help-and-advice-page

### Context

Now that TTT events are over until October, it makes sense to make them less prominent on the help and advice page.

I've moved TTAs above events and moved TTT events below the other events.


gemmadallmandfe (approved):
Looks good to me

PR #2749, opened 2022-08-16
redirecting old if youre disabled page

EllieNodder:
### Trello card

https://trello.com/c/2R8JdXpF/3533-unpublish-old-if-youre-disabled-page-and-redirect-to-new-page

### Context

Now that we have a new 'if you're disabled' page under funding and support, we need to unpublish the old one and redirect it.


gemmadallmandfe (approved):
Looks good to me

PR #2753, opened 2022-08-17
Fix TTT events banner not showing

ethax-ross:
We were checking if TTT events exist but not taking into account the dates; as old events are still in the system we need to filter to those that start after the current date/time.


gemmadallmandfe (approved):
Happy to approve!

PR #2768, opened 2022-08-25
Make category cards semantically correct

ethax-ross:
### Trello card

[Trello-3552](https://trello.com/c/NBhH1e8x/3552-investigate-accessibility-issue-with-navigation-cards)

### Context

The category cards are currently a `nav` element with several child anchor links. These should be part of an unordered list to be more semantically correct (according to Silktide).

- Make category cards semantically correct

### Guidance to review

The page shouldn't change visually.


leoapost (approved):

gemmadallmandfe (approved):
Visually this looks the same to me - happy to approve

PR #2770, opened 2022-08-30
Omit page param from redirects

ethax-ross:
### Trello card

[Trello-3565](https://trello.com/c/lAfUB6Cg/3565-remove-query-parameters-from-event-redirects)

### Context

We occasionally see a redirect from an old set of event search results with a particular page. Often the particular page no longer exists due to events expiring over time and the user ends up with the 'no events found' page, which isn't ideal.

Ignore the `page` parameter when redirecting to provide a better user experience.

- Omit page param from redirects

### Guidance to review

If that route gets any more complicated it would be worth splitting it out into a small module, but as it stands I've left it in the routes file for now.

[Example test redirect](https://review-get-into-teaching-app-2770.london.cloudapps.digital/event-categories/school-and-university-events?page=8)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -16,8 +16,9 @@ Rails.application.routes.draw do
   YAML.load_file(Rails.root.join("config/redirects.yml")).fetch("redirects").tap do |redirect_rules|
     redirect_rules.each do |from, to|
       get from => lambda { |env|
+        request = Rack::Request.new(env)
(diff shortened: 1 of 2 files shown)


gemmadallmandfe:
Test redirect works as expected - looks good @ethax-ross 

Will leave approving this PR to tech eyes though :-)

leoapost (approved):

PR #2788, opened 2022-09-02
Change docker caching from gha to registry

RMcVelia:
### Trello card

https://trello.com/c/96RrOmzk/667-change-caching-from-gha-to-registry

### Context

The build workflow is currently using type=gha for docker image caching.
Our preferred method is registry, which gives more control over the caching, and also maintains a common standard between repositories.

Also removing the tar file for the test image, and pushing to ghcr to use caching instead.

Update build-and-deploy workflow
Update build-no-cache workflow

### Guidance to review

Check build logs to make sure it's working as expected
- https://github.com/DFE-Digital/get-into-teaching-app/actions/runs/2977982281 

Check run time to confirm it's not slower than the previous method
Check image tags in ghcr

### DIFF
--- a/.github/workflows/build-no-cache.yml
+++ b/.github/workflows/build-no-cache.yml
@@ -44,12 +44,24 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build docker image
+      - name: Build and push base image
         uses: docker/build-push-action@v3.1.1
         with:
+          context: .
+          tags: ${{ env.DOCKER_REPOSITORY }}:base-master
+          push: true
+          target: base
+          build-args: |
+            SHA=${{ steps.sha.outputs.short }}
+            BUILD_TYPE=release
+
+      - name: Build release image locally
+        uses: docker/build-push-action@v3.1.1
+        with:
+          context: .
           tags: ${{ env.DOCKER_REPOSITORY }}:master
-          load: true
           push: false
+          load: true
           build-args: |
             SHA=${{ steps.sha.outputs.short }}
             BUILD_TYPE=release
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,13 +53,14 @@ jobs:
         run: |
              if [ "${{github.ref}}" == "refs/heads/master" ]
              then
-                echo ::set-output name=DOCKER_IMAGE::${{ env.DOCKER_REPOSITORY }}:sha-${{ steps.sha.outputs.short }}
-                echo ::set-output name=DOCKER_EVENT::${{ env.DOCKER_REPOSITORY }}:master
+                GIT_BRANCH=master
              else
-                echo ::set-output name=DOCKER_SHA::${{ env.DOCKER_REPOSITORY }}:sha-${{ steps.sha.outputs.short }}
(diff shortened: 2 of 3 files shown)


.github/workflows/build-no-cache.yml
```diff
@@ -44,12 +44,24 @@ jobs:
...
         with:
+          context: .
+          tags: ${{ env.DOCKER_REPOSITORY }}:base-master
+          push: true
+          target: base
+          build-args: |
+            SHA=${{ steps.sha.outputs.short }}
+            BUILD_TYPE=release
+
+      - name: Build and push to Docker Hub
```
saliceti, line 58: We use `push: false`. Should the title be something like "Build docker image locally" ?

.github/workflows/build.yml
```diff
@@ -68,18 +70,37 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Build Base
+        uses: docker/build-push-action@v3.1.1
+        with:
+          context: .
+          cache-from: |
+            type=registry,ref=${{ env.DOCKER_REPOSITORY}}:base-PR-${{ github.event.number }}
```
saliceti, line 78: Can't we use the DOCKER_BASE variable?

.github/workflows/build.yml
```diff
@@ -55,10 +55,12 @@ jobs:
              then
                 echo ::set-output name=DOCKER_IMAGE::${{ env.DOCKER_REPOSITORY }}:sha-${{ steps.sha.outputs.short }}
                 echo ::set-output name=DOCKER_EVENT::${{ env.DOCKER_REPOSITORY }}:master
+                echo ::set-output name=DOCKER_BASE::${{ env.DOCKER_REPOSITORY }}:base-master
              else
                 echo ::set-output name=DOCKER_SHA::${{ env.DOCKER_REPOSITORY }}:sha-${{ steps.sha.outputs.short }}
                 echo ::set-output name=DOCKER_IMAGE::${{ env.DOCKER_REPOSITORY }}:review-${{steps.sha.outputs.short }}
                 echo ::set-output name=DOCKER_EVENT::${{ env.DOCKER_REPOSITORY }}:PR-${{ github.event.number }}
+                echo ::set-output name=DOCKER_BASE::${{ env.DOCKER_REPOSITORY }}:base-PR-${{ github.event.number }}
```
saliceti, line 63: We usually use the branch name. Can we use it for consistency with other services?

.github/workflows/build.yml
```diff
@@ -190,11 +227,8 @@ jobs:
       - name: set-up-environment
         uses: DFE-Digital/github-actions/set-up-environment@master
 
-      - name: Download Artifacts
-        uses: actions/download-artifact@v3
-
-      - name: Load Docker Image
-        run: docker load --input app_test/image.tar
+      - name: Pull Docker Image
```
saliceti, line 230: Is this required? The `docker run` task should do the pull automatically

.github/workflows/build-no-cache.yml
RMcVelia: changed title

.github/workflows/build.yml
RMcVelia: changed to use DOCKER_BASE

.github/workflows/build.yml
RMcVelia: changed DOCKER_BASE to be the branch (for PR) or master (for pull)

saliceti (approved):

PR #2790, opened 2022-09-02
Update readme with details on adding YouTube videos

ethax-ross:
Update the readme with details of how to add YouTube videos via frontmatter.


EllieNodder (approved):
Makes sense to me

PR #2794, opened 2022-09-06
Fix focus style on assessment only provider links

ethax-ross:
### Trello card

[Trello-3591](https://trello.com/c/OrNbRpE0/3591-fix-styling-bug-on-assessment-only-page)

### Context

The background color of the focus state for the link was not being applied as the anchor was not a block element.

- Fix focus style on assessment only provider links


gemmadallmandfe (approved):
Looks good to me

PR #2798, opened 2022-09-09
Add emergency banner

ethax-ross:
> ⚠️ **do not merge**


gemmadallmandfe:
Thanks @ethax-ross

NB The URL doesn't work - will need to be a full URL
ie https://www.gov.uk/government/topical-events/her-majesty-queen-elizabeth-ii

ethax-ross:
Closing as the guidance appears to indicate we shouldn't add a banner; we can re-open the PR if that changes.

PR #2804, opened 2022-09-13
replacing train to teach with get into teaching events

EllieNodder:
### Trello card

https://trello.com/c/rrrhDECs/3584-change-references-from-train-to-teach-events-to-get-into-teaching-events

### Context

Now that Train to Teach events are being renamed Get Into Teaching events, we need to change all references to these on the website.

### DIFF
--- a/app/models/event_type.rb
+++ b/app/models/event_type.rb
@@ -1,7 +1,7 @@
 class EventType
   ALL =
     {
-      "Train to Teach event" => 222_750_001,
+      "Get Into Teaching event" => 222_750_001,
       "Question Time" => 222_750_007,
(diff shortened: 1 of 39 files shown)


app/models/event_type.rb
```diff
@@ -1,15 +1,15 @@
 class EventType
   ALL =
     {
-      "Train to Teach event" => 222_750_001,
+      "Get Into Teaching event" => 222_750_001,
```
gemmadallmandfe, line 4: @EllieNodder It is very likely that we will have a new event type so this shouldn't be renamed (along with some of the other event functionality stuff - better to leave to Ross!)

PR #2805, opened 2022-09-14
Add about Get into Teaching events page

ethax-ross:
> ⚠️ WIP


gemmadallmandfe:
@ethax-ross @GDW7 This is the old PR which I think Ross has merged into https://github.com/DFE-Digital/get-into-teaching-app/pull/2809

Can we close this one now so there is no confusion? Or do we need to keep it open still?

PR #2807, opened 2022-09-15
updating train to teach events name

EllieNodder:
### Trello card

https://trello.com/c/rrrhDECs/3584-change-references-from-train-to-teach-events-to-get-into-teaching-events

### Context

Train to Teach events are being renamed Get Into Teaching events so we need to tweak the copy on the website where they are mentioned.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -40,8 +40,8 @@ en:
   find_an_event:
     types:
       222750001: |-
-        <p>Our Train to Teach events will provide you with a wealth of information and help you turn questions to confidence on your journey to the classroom.</p>
-        <p>At some of our Train to Teach events you can meet a whole range of local training providers; at others you’ll have the chance to put your questions to a panel of experts.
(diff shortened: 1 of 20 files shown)


config/locales/en.yml
```diff
@@ -40,8 +40,8 @@ en:
   find_an_event:
     types:
       222750001: |-
-        <p>Our Train to Teach events will provide you with a wealth of information and help you turn questions to confidence on your journey to the classroom.</p>
```
gemmadallmandfe, line 43: @ethax-ross 

I'm presuming that this change should be revoked as:

1. this description is tied to an event type, and Get Into Teaching events have a new event type
2. I don't think this is used anyway - was this to do with the old event design?

ethax-ross: Yes, and yes probably, I'll take a look and remove it in the main PR if its redundant

ethax-ross:
Closing as I've pulled this into my main PR now

PR #2808, opened 2022-09-15
Remove Question Time events

ethax-ross:
### Trello card

[Trello-3649](https://trello.com/c/543WVKAS/3649-update-api-to-support-get-into-teaching-events)

### Context

We are migrating away from the Train to Teach/Question Time events to a new 'Get Into Teaching event' type. The Question Time events have always complicated things because we merge them and treat them as Train to Teach events.

Remove Question Time events to simplify the Train to Teach events before we migrate them over to Get Into Teaching events.

- Remove Question Time events

- Ignore unrecognised query parameters

We currently raise an unhandled exception if an unrecognised query parameter is used on the event search page. As we are removing Question Time events its likely we will get requests that contain the corresponding query parameter still.

Handle unrecognised query parameters more gracefully by ignoring them and returning the search results.

### Guidance to review

Nothing should appear different as far as the user is concerned; we have no Question Time events in production and they were always masked as Train to Teach events anyway.

### DIFF
--- a/app/components/teaching_events/event_component.rb
+++ b/app/components/teaching_events/event_component.rb
@@ -18,7 +18,7 @@ module TeachingEvents
     delegate :provider_event?, to: :type
 
     def train_to_teach?
-      type.train_to_teach_or_question_time_event?
+      type.train_to_teach_event?
     end
 
     def online?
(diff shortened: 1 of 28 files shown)


gemmadallmandfe (approved):
Happy for this to be merged

misaka (approved):

PR #2810, opened 2022-09-15
Change adviser link on food for thought page

gemmadallmandfe:
### Trello card
https://trello.com/c/PwiBz9If

### Context
Updated adviser link to go to Explore teaching advisers page rather than TTA funnel start page

### DIFF
--- a/app/views/content/food-for-thought/_promo.html.erb
+++ b/app/views/content/food-for-thought/_promo.html.erb
@@ -8,6 +8,6 @@
   ) do %>
       <p>Talk to an experienced former teacher who can help you find school experience, learn what it's like to be in the classroom and answer all your questions about getting into teaching.</p>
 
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
@EllieNodder I have changed the adviser link on the food for thought page

We were also going to consider changing the photo in the promo module to the ETA photo - I'm not sure this would work though?

EllieNodder:
> 

If we say 'explore teaching advisers', the pic covers up her eyes. If we change it to 'get an adviser', it's better. 
image

I think saying 'explore teaching adviser' maybe sounds better and less of a commitment for this early audience than 'get an adviser'. But I do prefer the new picture (probably just because I've seen the old one a billion times though!!)

What do we prefer?

gemmadallmandfe:
Let's leave the photo as is - we can change it if we get any other ones

ethax-ross (approved):

PR #2818, opened 2022-09-23
Show all events on the about GiT events page

ethax-ross:
### Trello card

[Trello-3664](https://trello.com/c/KjGDyeg0/3664-investigate-why-december-national-git-event-is-not-appearing-on-about-git-events-page)

### Context

The API caps the events to 10 by default and we have 11 running at the moment; increase the threshold to 20 so that we get all of the current events.

- Show more events on the about GiT events page

- Stop caching the about events page

As we now display teaching events on this page we don't want to have it as part of the static cache (or when the event expires it won't remove until a deploy happens).

Stop caching the about events page.

### DIFF
--- a/app/controllers/teaching_events_controller.rb
+++ b/app/controllers/teaching_events_controller.rb
@@ -3,8 +3,6 @@ class TeachingEventsController < ApplicationController
 
   include CircuitBreaker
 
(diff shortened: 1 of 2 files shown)


gemmadallmandfe (approved):
Looks good to me

PR #2824, opened 2022-09-26
Fix issues with internal events form

ethax-ross:
### Trello card

[Trello-3657](https://trello.com/c/r0jdzIGZ/3657-investigate-issues-with-internal-provider-events-form?filter=member:rossoliver15)

### Context

We have had reports from users of the internal events form that the building selection is not always working and the rate limiting has occasionally been hit when adding new events.

- Fix internal events JS not loading

The internal events Javascript was set to load `async`, which meant it would often execute before the page document was ready and the `js-enabled` class was not present on the document body, resulting in it not running.

Change the JS load method to `defer` so it will always execute after page load and consolidate the logic to add a `js-enabled` class so we're doing it consistently.

- Increase internal events rate limiting thresholds

A provider has notified us that they have hit the 'too many requests' error reasonably often on the internal events form; increasing the rate limit threshold to avoid false positives.

### Guidance to review

You can check it works by navigating to the [internal events portal](https://review-get-into-teaching-app-2824.london.cloudapps.digital/internal/events) and signing in as a publisher to add a new event.

### DIFF
--- a/app/views/layouts/internal.html.erb
+++ b/app/views/layouts/internal.html.erb
@@ -3,7 +3,7 @@
 <% content_for :head do %>
   <%= csrf_meta_tags %>
   <%= stylesheet_pack_tag 'internal', 'data-turbolinks-track': 'reload', media: 'all' %>
-  <%= javascript_pack_tag 'internal', 'data-turbolinks-track': 'reload', async: true %>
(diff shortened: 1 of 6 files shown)


misaka (approved):
🚢

PR #2828, opened 2022-09-26
Fix page titles on error

ethax-ross:
### Trello card

[Trello-3660](https://trello.com/c/BBZSmQaq/3660-fix-title-missing-on-validation-errors)

### Context

When the event sign up form is submitted with errors the subsequent page doesn't render a correct page title. We should also be prefixing page titles with `Error:` to be inline with the GOV.UK guidance.

- Ensure page title is set on validation errors

When the event forum is submitted with validation errors the `before_action` wasn't running to se the page title.

- Prefix page title if the page contains errors

Inline with the GOV.UK guidance we should prefix page titles that contain an error message with `Error: `.

Add middleware to prefix page titles if the GOV.UK error summary element is found within the page.

### Guidance to review

Go to sign up for an event; cause a validation error (leave a field blank, for example) and check the title has `Error: <page title> | Get Into Teaching GOV.UK`.

The other way of doing this would be to modify the `page_title` helper to check for a `422` status code and prefix the title. As Rails doesn't render a `422` by default for validation errors this felt more robust and is also inline with a pattern we have in place for transforming HTML responses.

### DIFF
--- a/lib/middleware/html_response_transformer.rb
+++ b/lib/middleware/html_response_transformer.rb
@@ -3,6 +3,7 @@ require "lazy_load_images"
 require "responsive_images"
 require "external_links"
 require "image_sizes"
+require "error_title"
 
 module Middleware
   class HtmlResponseTransformer
@@ -32,6 +33,7 @@ module Middleware
       response_body = ResponsiveImages.new(response_body).html
       response_body = LazyLoadImages.new(response_body).html
       response_body = ExternalLinks.new(response_body).html
+      response_body = ErrorTitle.new(response_body).html
 
       # <source> has no content so should be self-closing; configuring Nokogiri
       # to remove the closing tags results in breakages elsewhere in the document,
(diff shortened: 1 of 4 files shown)


gemmadallmandfe (approved):
Looks good to me - tested successfully

lib/middleware/html_response_transformer.rb
```diff
@@ -32,6 +33,7 @@ def transform(response_body)
       response_body = ResponsiveImages.new(response_body).html
       response_body = LazyLoadImages.new(response_body).html
       response_body = ExternalLinks.new(response_body).html
+      response_body = ErrorTitle.new(response_body).html
```
misaka, line 36: Just wondering ... is this not starting to get expensive to parse and re-parse the document 6 times with Nokogiri?

misaka: (I feel like I've asked this question on a previous PR ... the one for the LazyLoadImages maybe?)

misaka: Looking into Skylight it looks like this method may be taking up to ~1/3 of the request time ... we agreed to have a look at that as a separate piece of work.

misaka (approved):

PR #2829, opened 2022-09-27
Ensure images are all optimised

ethax-ross:
### Trello card

[Trello-3648](https://trello.com/c/6vu6WDcK/3648-add-a-test-that-fails-if-an-image-size-is-excessive)

### Context

We occasionally find that large/unoptimised images have been accidentally added to the website; this can have a negative impact on the page speed performance and user experience. We want to prevent laege images from being added by automatically flagging it in the build.

- Add unit test to ensure all images are optimised We want to make sure all of our images are reasonably optimised on the
website. We have a pipeline that compresses images before deployment, but it makes sense to ensure we don't add enormous images accidentally to the repository.

Add a unit test that will fail if it finds an image larger than 500kb.

- Optimise large images

These images were identified by the new unit test; optimise them to be under 500kb.


gemmadallmandfe (dismissed):
The image changes all look good to me - I've been through and double-checked where they appear on the site

Will leave the unit test to tech eyes!

spec/requests/image_file_sizes_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "rails_helper"
+
+describe "Image File Sizes", type: :request do
+  include Webpacker::Helper
```
asatwal, line 4: Is this include required here ? Not sure if any webpacker helpers are used in this specific test.

asatwal (dismissed):
Just one minor comment. Otherwise looks good.

gemmadallmandfe (approved):
Happy for this to be merged following the tech review

PR #2830, opened 2022-09-27
Include GiT/Online Q&A events in sitemap

ethax-ross:
### Trello card

[Trello-3706](https://trello.com/c/7Qp3pHO9/3706-add-git-events-and-online-qas-to-sitemap)

### Context

We want to include GiT events and Online Q&A events in the sitemap.xml so that Google hopefully does a better job of keeping our indexed pages up to date.

- Include GiT/Online Q&A events in sitemap

Add Get into Teaching events and Online Q&A events to the `sitemap.xml`.

We only include future events and limit to 100 (which should be plenty to cover all active events at any given time).

### Guidance to review

The sitemap can be viewed [here](https://review-get-into-teaching-app-2830.london.cloudapps.digital/sitemap.xml).

It uses the global default `lastmod` for the entries; we don't expose a 'created at' date from the API and from what I've ready Google ignores the `lastmod` anyway so its probably not worth doing so.

### DIFF
--- a/app/controllers/sitemap_controller.rb
+++ b/app/controllers/sitemap_controller.rb
@@ -1,5 +1,6 @@
 class SitemapController < ApplicationController
   DEFAULT_LASTMOD = "2021-03-01".freeze
+  FUTURE_EVENTS_LIMIT = 100
 
   ALIASES = {
     "/home" => "/",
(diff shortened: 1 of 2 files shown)


gemmadallmandfe (dismissed):
Looks good to me

app/controllers/sitemap_controller.rb
```diff
@@ -45,6 +45,22 @@ def build
...
+  end
+
+  def event_pages
+    events.map { |e| event_path(e.readable_id) }.index_with({})
+  end
+
+  def events
+    GetIntoTeachingApiClient::TeachingEventsApi.new.search_teaching_events(
+      start_after: Time.zone.now,
+      quantity: 100,
```
asatwal, line 58: I would prefer a constant (defined in this controller like DEFAULT_LASTMOD), called something like FUTURE_EVENTS_LIMIT rather than a hardcoded number, but that's just my preference.

asatwal (dismissed):
Looks Good - Just  minor comment.

gemmadallmandfe (approved):
Looks good to me

PR #2837, opened 2022-09-29
Adding scholarships/bursaries holding message

EllieNodder:
### Trello card

https://trello.com/c/7kZJqURg/3708-add-holding-message-regarding-bursaries-and-scholarships-ahead-of-announcement

### Context

The Financial incentives team have suggested that we add some kind of holding message on GIT before Find opens to make it clear that the bursary and scholarship amounts we are displaying relate to the previous recruitment cycle (ie trainees starting now).

### DIFF
--- a/app/views/content/subjects/engineers-teach-physics/_article.html.erb
+++ b/app/views/content/subjects/engineers-teach-physics/_article.html.erb
@@ -118,7 +118,7 @@
      Scholarships of £26,000 or bursaries of £24,000 are available for trainee physics teachers starting their teacher training course in 2022.
      </p>
      <p>
-     Financial support for courses starting in 2023 will be announced by the end of the year.
+     Financial support for courses starting in 2023 will be announced in autumn 2022.
      </p>
 
      <p>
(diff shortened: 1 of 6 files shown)


app/views/content/subjects/engineers-teach-physics/_article.html.erb
```diff
@@ -115,14 +115,17 @@
     <h2 class="heading-s">Funding for teaching physics</h2>
 
     <p>
-     Scholarships of £26,000 or bursaries of £24,000 are available for trainee physics teachers. This is money that you do not have to pay back.
+     Scholarships of £26,000 or bursaries of £24,000 are available for trainee physics teachers starting their teacher training course in 2022.
+     </p>
+     <p>
+     Financial support for courses starting in 2023 will be announced by the end of the year.
```
EllieNodder, line 121: @gemmadallmandfe the message the FI team gave was that the amounts would be released 'soon', but I don't think that's particularly useful for users. I've put instead 'by the end of the year' - can we double check if this is accurate/if we have a more accurate timing for this?

gemmadallmandfe: @EllieNodder the announcement is likely to be 'soon' but we don't have a clear date yet!

gemmadallmandfe (approved):
Looks good to me

PR #2839, opened 2022-09-30
Whitelist assets host in CSP

ethax-ross:
### Trello card

[Trello-3710](https://trello.com/c/aBBbyJUA/3710-update-csp-to-unblock-welcome-guide-video)

### Context

We have noticed that a video hosted on our assets server is being blocked by our Content Security Policy. We're not sure why/when this started happening and other directives appear to allow the Rails `asset_host` (images are fine, for example).

- Whitelist asset host in CSP

Explicitly whitelist the assets host if set to ensure videos are displayed; we will investigate why it's only effecting the `media_src` directive.

### DIFF
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -51,6 +51,13 @@ SecureHeaders::Configuration.default do |config|
   data                 = ["data:"]
   blob                 = ["blob:"]
 
+  # We're not sure why yet but the asset host needs to be
+  # explicitly whitelisted in the media_src directive or the CSP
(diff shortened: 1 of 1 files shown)


gemmadallmandfe (approved):
Looks good to me - video is now working on Welcome page and other videos on the site are unaffected

asatwal (approved):
:+1 LGTM.

ethax-ross:
> Looks good to me - video is now working on Welcome page and other videos on the site are unaffected

So the review app hosts images on the same instance and it will always work there; we're going to have to ship this to see if it fixes it as we only have an external asset host configured on test/prod

PR #2852, opened 2022-10-03
Move event videos into the hero section

ethax-ross:
### Trello card

[Trello-3693](https://trello.com/c/glJcgHDP/3693-consider-adding-video-to-individual-event-pages)

### Context

The events team have requested videos to be added to the Get Into Teaching events at thee top of the page. We want to render the event images as a thumbnail preview for the videos and then play the video on click.

- Move event videos into hero section

We are assigning YouTube videos to the Get Into Teaching events on the website; currently they display under the event summary information quite a way down the page. Instead, we want to indicate that the event image in the hero section is clickable and, when clicked, will transition to play the YouTube video.

### Guidance to review

A limitation with modern browsers is that they will not let videos auto-play with sound enabled, this means the only way (short of using something other than youtube video embedding) we can play the videos on clicking the preview image is if it is muted.

We should check the other YouTube videos on the website still look fine after this change.

[Example event with a video/preview](https://review-get-into-teaching-app-2852.london.cloudapps.digital/events/221101-get-into-teaching-south-west-event)

### DIFF
--- a/app/components/content/youtube_video_component.html.erb
+++ b/app/components/content/youtube_video_component.html.erb
@@ -1,8 +1,4 @@
-<%= tag.iframe(
-  title: title,
-  loading: "lazy",
-  src: src,
-  frameborder: 0,
-  allow: "autoplay; encrypted-media",
-  allowfullscreen: true
-) %>
+<div class="youtube-video" data-controller="youtube-video">
+  <%= preview %>
+  <%= video %>
+</div>
--- a/app/components/content/youtube_video_component.rb
+++ b/app/components/content/youtube_video_component.rb
@@ -1,19 +1,63 @@
 module Content
   class YoutubeVideoComponent < ViewComponent::Base
-    attr_reader :id, :title
+    attr_reader :id, :title, :preview_image
 
-    def initialize(id:, title:)
+    def initialize(id:, title:, preview_image: nil)
       super
 
       @id = id
       @title = title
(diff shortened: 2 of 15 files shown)


gazcarr:
Is there anyway the white heading background can flow with the copy? Or we stop the overlap of it above the video and let it sit in the column with the rest of the content?

gemmadallmandfe:
Looks good to me @ethax-ross - I've tested on iPhone 11 as well

I've tested the videos on the following pages and they work fine still:
* https://review-get-into-teaching-app-2852.london.cloudapps.digital/returning-to-teaching
* https://review-get-into-teaching-app-2852.london.cloudapps.digital/teacher-training-advisers

(NB these videos are not muted - presumably because of the way they are embedded.

I've also tried using a long event name and looked at other kinds of event pages too (online Q&As and provider events) - all looks good

Happy for this to be merged when:
a) videos are in production (or before if there is not a dependency on this!)
b) you've had a tech review

NB just to confirm that these videos are only being added to the in-person GIT events and not the online national GIT events.

asatwal (approved):
👍 Technically looks good to me. I've checked the review App also. Might be worth getting a UX Person (or equivalent) to check it complies with the GDS accessibility requirements.

gemmadallmandfe:
Closing this PR for now as per Trello ticket - we may revisit once we have revised videos, but will need to review for accessibility

PR #2854, opened 2022-10-04
Remove unused code from the website

ethax-ross:
### Trello card

[Trello-3714](https://trello.com/c/lGem9aar/3714-remove-legacy-videocontrollerjs-and-associated-partials-css)

### Context

We want to tidy up the website layouts and align them better with the designs. Ahead of doing this it makes sense to purge the website of any unused partials, view components, Stimulus controllers or CSS.

- Remove legacy video player

We had a video player on the website that was a popup modal; we have since replaced it with the `YoutubeVideoComponent` and can remove the legacy component/Stimulus controller.

- Remove unused CardComponent

The `CardComponent` was used in the old My Story into Teaching section that has since been removed. It also had two component subclasses that were no longer referenced anywhere.

- Remove unused view partials

As far as I can tell these are not referenced anywhere. If we delete a partial that is referenced somewhere the content tests should pick up on it. 

- Remove unused view components

These no longer appear to be referenced anywhere. I've also expanded the
test cases on the `CallToActionComponentInjector` to cover additional view
components I found.

- Rename view components for consistency

These view components should be suffixed with `Component` to be consistent with all our other view components.

- Remove eligibility checker

We have never enabled this in production and its been disabled completely for many months; if we decide we want to bring it back at some point we can always resurrect it from GitHub.

- Remove unused CSS

This was just a quick pass through the CSS to identify any main classes that aren't being referenced any longer.

- Remove unused Markdown file

### Guidance to review

Best reviewed by-commit.

The tests _should_ catch any major issues with this change, but its worth eyeballing the website in general to look for anything untoward (most likely broken CSS as its sometimes hard to figure out if a selector is actually in use or not).


asatwal (approved):
Looks Good as far as I can tell.

gemmadallmandfe (approved):
Happy for this to be merged given that it has had a tech review

PR #2858, opened 2022-10-04
removing (honours) from copy

EllieNodder:
### Context

Tom Crookes Smith pointed out that you don't need a bachelor's degree **(honours)** to be a teacher - confirmed by Phoebe in policy. So we're removing the **(honours)** from the copy.


gemmadallmandfe (approved):
Looks good to me

PR #2859, opened 2022-10-04
Make closed events viewable but not open for registration

ethax-ross:
### Trello card
https://trello.com/c/dW8y2mPP

We currently don't have consistent logic around when an event should be viewable but not open for registration. Going forward we are going to allow closed events to be viewable (as we already display them in search results) but _only_ allow registration to open events.

Also updates the events helper to determine if an event can be signed up to so that it is consistent with the `EventStatus` module.

We're going to take a closer look at the different scenarios in general to clean some of this up (in terms of what to do when an event is open/closed/past/future/has web feed id/etc in a follow-up ticket/PR.

### DIFF
--- a/app/controllers/event_steps_controller.rb
+++ b/app/controllers/event_steps_controller.rb
@@ -23,10 +23,10 @@ protected
 private
 
   def restrict_sign_ups
-    event_is_viewable = EventStatus.new(@event).viewable?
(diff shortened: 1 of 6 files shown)


gemmadallmandfe (approved):
Looks good to me from the front end

Happy for you to wait for a tech review before merging

asatwal (approved):
👍 LGTM

PR #2861, opened 2022-10-05
Clean up layouts

ethax-ross:
### Trello card

[Trello-3723](https://trello.com/c/LB5fkUfy/3723-clean-up-existing-layouts)

### Context

Ahead of aligning the design layouts with our application layouts it makes sense to clean up what we have and try and consolidate some of the logic between the layout files in order to make them simpler to understand.

- Clean up generic body tag

We have an `analytics_body_tag` that used to be used to add a GTM controller. As we no longer use a GTM controller it doesn't make sense to have this helper; instead it can be re-purposed into a generic `body_tag` helper that adds global Stimulus controllers and data attributes/classes.

- Encapsulate gtm_enabled? check in partial

To avoid calling it everywhere we can move this check into the partial.

- Clean up main tag in layouts
 
Make the main tags consistent and bring it together in a helper so its easier to change in the future.

- Rename main-body to main-section

Each page has a `main-body` at the moment, but some pages (such as the category page) have multiple. The naming is confusing as there can only ever be a single HTML body tag; renaming to `main-section` to make it clearer that there can be several on a page.

- Move GTM fallback script into HeaderComponent

We include a `HeaderComponent` in all layouts (apart from internal, but we don't want GTM there anyway) so we can shift
the GTM fallback partial rendering into there to DRY it up. 

- Rename campaign header component to hero

We have a `Campaign::HeaderComponent` we use on some layouts, however we have another `HeaderComponent` that is on every page and this makes the naming confusing as the campaign pages have two 'headers'. Renaming to `CampaignHeroComponent` clears up the confusion and brings it inline with the existing `HeroComponent` naming.

### Guidance to review

Best reviewed by-commit.


gemmadallmandfe:
@ethax-ross I've had a good click around the site both on desktop and mobile (iPhone 11) and all looks OK to me

I've signed up for mailing list and and event, looked at the Welcome Guide, looked at subject-specific template, checked funding widget, done event searches, etc. etc.

Presume you will need tech eyes on this so haven't approved - but happy for this to go out next week once it has had a tech review

asatwal (commented):
👍  Nice. Like the commit by commit changes and clear comments.

Needs a rebase though to resolve conflicts.

gemmadallmandfe (approved):
Happy for this to be merged now that it has had a tech review

PR #2862, opened 2022-10-05
Redirect /home to root page

ethax-ross:
### Trello card

[Trello-3721](https://trello.com/c/0gWcggxe/3721-fix-canonical-issue-with-home-page)

### Context

Google picked up on an issue where our home page is served on `/home` (the name of the markdown file) but its canonical should be `/` and its also set to `/home`.

To get around this we should just redirect `/home` to `/` so that its not accessible under the `/home` path at all.

- Redirect /home to the root page


gemmadallmandfe (approved):
Looks good to me

PR #2882, opened 2022-10-13
Clean up subject specific layouts

ethax-ross:
### Trello card

[Trello-3678](https://trello.com/c/RErN2o9v/3678-simplify-subject-specific-template)

### Context

 The subject specific set of CSS classses has since been used in other areas of the website (other than the subject specific pages) such as the "About Get Into Teaching Events" page. It was also hacked at a fair amount to get the about events page shipped in time for the new event cycle.

Instead, we want to rename these set of classes to 'campaign' as they tend to be paired with the campaign hero image and used for campaign-related content.

As part of this refactor I've also attempted to standardise the layout widths we use across the website; centralising them in the breakpoints SCSS file:

```
$page-max-width: 2000px;

$content-max-width: 1200px;

$column-845: 845px;
$column-720: 720px;
$column-570: 570px;
```

Part of what made this difficult is that we have a generic `container` (1200px) that, when used with a `markdown` sub-element it will indent content but not headings. This worked on all the content pages but ended up getting messy/custom solutions in other layouts. To try and replicate this consistently I've added an `inset` qualifier to the `container` class which can be used to replicate the indentation on a more granular level:

```
<section class="container">
  <h1>Heading</h1>

  <div class="inset">
    Some indented content">
  </div>
</section>
```

I've also pulled out the 'What they're saying` CSS from the teaching events section so it can be used more generically. It would be nice to have a component for this in the future.

- Make subject specific layout generic/tidy

### Guidance to review

**The main pages this change effects are subject specific, welcome guide, about GiT events and other event pages.**

I originally tried to structure the commits but it proved too difficult as I was continuously chasing threads when I changed a bit of CSS.

The headings on the welcome page/subject specific pages remain a little inconsistent (some indent but not all), but I think its OK for now (and no worse off than it is in prod).

I've not got all the layouts 100% consistent but its an improvement and I'll follow up in another PR to to ensure spacing/content widths are consistent everywhere (we should be using `$indent-amount` everywhere, but we have custom margins/padding in some places). I'd also like to pull the various container widths out into view components to give it a bit more structure/clarity.


gemmadallmandfe:
@ethax-ross I've had a look at some of the pages (About GIT events, subjects/maths) and there are a few strange things going on eg padding above/below paras, image sizes etc.

Ideally I'd really like @gazcarr to look through them but this might not be possible before his leave!

Otherwise it may be easier to walkthrough over screenshare and discuss - I appreciate there may be things that you are planning to clear up later

app/views/content/subjects/maths/_article.html.erb
```diff
@@ -127,7 +127,7 @@
     </p>
   </section>
 
-  <%= render SubjectSpecific::TextBlockWithOneImageComponent.new(
+  <%= render Campaign::TextBlockWithOneImageComponent.new(
     heading: "Getting into teacher training",
     colour: "yellow",
     image_path: "media/images/content/hero-images/0009.jpg",
```
gemmadallmandfe, line 133: This image is quite small on the `/subjects/maths` page compared with current prod, both on mobile and desktop

It's not a huge showstopper - but the image definitely loses a bit of impact

image

app/views/content/subjects/maths/_article.html.erb
```diff
@@ -36,7 +36,7 @@
     </p>
   </section>
 
-  <%= render SubjectSpecific::TextBlockWithTwoImagesComponent.new(
+  <%= render Campaign::TextBlockWithTwoImagesComponent.new(
```
gemmadallmandfe, line 39: These images look much better on desktop now - still a little small on mobile compared with prod, so they lose a bit of impact I think?

app/views/content/explore-teaching-advisers/_article.html.erb
```diff
@@ -30,7 +30,7 @@
     </p>
   </section>
 
-  <%= render SubjectSpecific::TextBlockWithTwoImagesComponent.new(
+  <%= render Campaign::TextBlockWithTwoImagesComponent.new(
     heading: "What kind of teacher could you be?",
```
gemmadallmandfe, line 34: On prod, there is a bit more white space above this section / below the body copy which would be good to adjust if possible - gives a bit more breathing space to the page

(this is just one example - this is generally true throughout these pages)

image

image

app/webpacker/styles/campaign/_videos.scss
```diff
@@ -0,0 +1,37 @@
+.videos {
```
gemmadallmandfe, line 1: As mentioned yesterday, the videos are slightly smaller (but proportionally) than those on prod

eg on `/events/about-get-into-teaching-events`

I don't think this is an issue - as agreed, I'll raise a separate ticket on videos for a later sprint

gemmadallmandfe (commented):
@ethax-ross 

I've reviewed the following so far on laptop and mobile (iPhone 11):

* `/subjects/maths`
* `/subjects/engineers-teach-physics`
* `/explore-teaching-advisers`
* `/landing/food-for-thought`

* `/events`
* `/events/about-get-into-teaching-events`

I've left a few minor comments but looking much better

I'll review the Welcome pages next but thought you might want to look at these in the meantime!

app/views/content/subjects/maths/_article.html.erb
ethax-ross: I'll make these a bit bigger but IMO they're too big in prod and almost hit the edge of the page for mobile; they should have our standard margin I think to be inline with the text below (although they'll be slightly off that due to the rotation!)

app/webpacker/styles/campaign/_videos.scss
ethax-ross: I'll make them a bit bigger for now and I've got another ticket for looking at videos as a whole later on

app/views/content/explore-teaching-advisers/_article.html.erb
gemmadallmandfe: looks much better now

app/webpacker/styles/welcome-guide.scss
```diff
@@ -751,7 +752,7 @@ $z-indices: ("bottom": 5, "low": 10, "medium": 15, "high": 20, "top": 25);
 
       .text {
```
gemmadallmandfe, line 753: not sure if this is the right bit of css - but the 'You're not the only one with questions is slightly out of kilter with the other headings on the Welcome guide page as it is full width on mobile and none of the other headings are

however, this is actually like this in prod so happy to leave as is if necessary

image

gemmadallmandfe (commented):
I've had a look at the welcome guide @ethax-ross - only a couple of very minor comments

Updates since my last comments look good!

asatwal (approved):
I haven't checked the layouts of the affected pages as this has been done already, but technically it looks like a good cleanup.

Just a few minor comments

app/views/teaching_events/about_git_events.html.erb
```diff
@@ -35,70 +35,67 @@
...
-
-    <section class="fullwidth grey-bg no-padding no-margin">
-      <section class="container no-margin">
-        <section class="category__cards">
-          <%= tag.nav(class: "category__nav-cards") do %>
-            <ul class="single-column">
-              <%= render(Categories::CardComponent.with_collection(categorise_events(@git_events))) %>
-            </ul>
-          <% end %>
+    <section class="overlapping">
```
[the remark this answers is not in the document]
ethax-ross: These need to be separate unfortunately; the `overlapping` has to encompass the the two child `section` elements and the `grey` only applies to the first child. The `container` constrains the width so if we combined `grey container` the grey background doesn't go all the way to the edge of the browser window.

PR #2892, opened 2022-10-14
Update homepage to promote Get Into Teaching events

ethax-ross:
### Trello card

[Trello-3747](https://trello.com/c/kNAJmWA8/3747-support-event-promo-launch?filter=member:rossoliver15)

### Context

There is a marketing programme for the Get Into Teaching events between 18th Oct and 30th Nov; we want to support the effort during this period by featuring Get Into Teaching events more prominently on the homepage.

- Rename badges to vertical-tags

We are going to have an alert style 'badge' component in the header going forward and to me these aren't really badges, so renaming to 'vertical tags' to avoid confusion.

- Add Stimulus controller for badges

A simple Stimulus controller that will make an ajax request for some JSON data and display the value of a given key in a 'badge' component. If the count is 0 the badge will not display.

- Show badge of open GiT events in header

We want to show a badge in the header next to a new 'Get Into Teaching events' link that indicates how many open GiT events there currently are.

- Update home page to focus on events

Update the hero section to prompt users to sign up for a Get Into Teaching event. Replace the event CTA further down the page with a mailing list CTA.

### Guidance to review

Getting the badge to appear consistently and without flickering on page changes was a bit tricky; I tried using `turbolinks-permanent` initially but it didn't yield good results, so instead I've opted to maintain a cache of the badge counts in the global `window` object. This way we can restore the badge immediately on Turbolinks page changes.

The initial page load (and full page reloads thereafter) may get a short delay before the badge appears. We can't get around this due to our static page cache (we could fix it on dynamic pages but it would introduce extra complexity and isn't worth it IMO).

The new content on the home page is much longer than we had previously and it causes the images to be really tall; we may want to do something about that.

I thought about writing a dedicated statistics endpoint on the API so we don't have to pull all the GiT events just to get the count, but in reality the events endpoint gets cached so we'll only be hitting the API once every ~10 mins and the number of GiT events is low and not expected to increase (around 10) so I think its fine for now.


app/views/content/home.md
```diff
@@ -5,10 +5,10 @@
     Read official Department for Education information to help you become a teacher. Learn about funding, teacher training, and steps to become a teacher. 
   date: "2021-06-16"
   fullwidth: true
-  subtitle: "Get information and support to help you become a teacher."
-  subtitle_button: "Register for updates"
-  subtitle_link: "/mailinglist/signup/name"
-  image: "media/images/content/hero-images/0012.jpg"
+  subtitle: "Whether you're ready to start your career in teaching or just curious, we can answer your questions at a Get Into Teaching event."
+  subtitle_button: "Get Into Teaching events"
+  subtitle_link: "/events?teaching_events_search[type][]=git"
```
gemmadallmandfe, line 10: Can this go to `/events/about-get-into-teaching-events` instead

gemmadallmandfe:
Generally looks good!

@ethax-ross 
Please can you change the two event links (button in hero and extra nav bar link) to go to `/events/about-get-into-teaching-events` rather than pre-filtered results on the `/events` page? We want to showcase the GIT events on the landing page

@gazcarr @EllieNodder 
Any comments? Hero image looks a bit poor quality on desktop for me?

app/views/content/home/_calls-to-action.html.erb
```diff
@@ -1,11 +1,11 @@
 <div class="homepage-feature blocks">
   <%= render CallsToAction::HomepageComponent.new(
     icon: "icon-calendar",
-    title: "Attend an event",
-    text: "Talk to course providers, teachers and expert advisers to get your questions answered in person or online at an event.",
-    link_text: "Find an event",
-    link_target: "/events",
-    image: "media/images/content/homepage/here-to-help.jpg")
+    title: "Get personalised guidance to your inbox",
+    text: "Answer a few questions about your current situation and we'll send you helpful advice and insights relevant to you.",
```
gemmadallmandfe, line 5:
```suggestion
    text: "Answer just a few questions and we'll send you helpful advice and insights relevant to you.",
```

I think talking about 'your current situation' sounds a bit offputting @EllieNodder ?

app/views/content/home/_calls-to-action.html.erb
```diff
@@ -1,11 +1,11 @@
...
+    link_text: "Register for updates",
```
gemmadallmandfe, line 6: Is updates the right term here @EllieNodder ? Might be worth trying out something else as we know that people were confused by this in UR - and it doesn't seem to fit with the content before it

app/components/header/extra_navigation_component.html.erb
```diff
@@ -1,10 +1,13 @@
 <%= tag.div(class: classes, data: { controller: "searchbox", "searchbox-search-input-id-value" => search_input_id }, role: "navigation") do %>
   <ul class="extra-navigation__list">
-    <li class="extra-navigation__link extra-navigation__mail">
-      <%= link_to("/mailinglist/signup/name", class: "link--black") do %>
-      Register your interest
-      <span class="icon icon-mail" aria-hidden="true"></span>
-      <span class="icon icon-mail-white" aria-hidden="true"></span>
+    <li class="extra-navigation__link">
+      <%= link_to(events_path({ teaching_events_search: { type: ["git"] }}), class: "link--black") do %>
```
gemmadallmandfe, line 4: @ethax-ross Can you change this link to https://getintoteaching.education.gov.uk/events/about-get-into-teaching-events please - we want people to go to the showcase page, not a filtered list on the events listing page

(does this mess up your event count?

app/components/header/extra_navigation_component.html.erb
```diff
@@ -1,10 +1,13 @@
...
+      <i
+        class="badge"
+        data-controller="badge"
+        data-badge-path-value="<%= git_statistics_events_path %>"
+        data-badge-json-key-value="open_events_count"></i>
+      Get Into Teaching events
```
gemmadallmandfe, line 10: @ethax-ross If this is the text used in the link in the extra nav bar, there is an random space at the beginning that needs removing:

image

app/views/content/home.md
gemmadallmandfe, line 10: @ethax-ross Can you change this link to https://getintoteaching.education.gov.uk/events/about-get-into-teaching-events please - we want people to go to the showcase page not a filtered list on events listing page

gemmadallmandfe:
@ethax-ross You said that the mailing list and TTA blocks had really tall images which we might want to address

@EllieNodder has cut this down a bit for the mailing list block since then - is this still an issue?

gemmadallmandfe:
Happy with the page (links all look good now @ethax-ross!) apart from the image quality - anything we can do about this @gazcarr ?

gazcarr:
@ethax-ross 

When we created the Hero variables there was an option to change the widths to allow for more copy on the left and reduce the size of the image. This should fix the quality issue but I am not sure how it will look in this scenario but think this should be our first option.

gemmadallmandfe:
This looks a bit odd on mobile now - image is poor and extends below yellow banner which I don't think it is meant to?

image

gemmadallmandfe (approved):
Looks good to me - happy for this to be deployed and then the hero image changed to an events one in a separate PR if possible

PR #2897, opened 2022-10-17
Unwind homepage changes supporting events comms burst

ethax-ross:
> :warning: Not to be merged until November.

### Trello card

[Trello-782](https://trello.com/c/RkUrPiwZ/782-unwind-changes-supporting-event-comms-burst-on-home-page)

### Context

We changed the home page to support the GiT events comms burst; once that has ended we want to re-instate the original home page design with some minor tweaks to the events CTA.

- Revert to register for interest header link

We want to restore the mailing list sign up link in the header; events will be called out further down in a CTA.

- Update homepage component for badge/caption

We want to add a purple caption and badge (containing text) to the homepage component. Update component with attributes and tweak CSS of badge to support fixed-width for multi-line text content.

- Update homepage CTA for events

Update the left homepage CTA to call out events.

### DIFF
--- a/app/views/content/home.md
+++ b/app/views/content/home.md
@@ -5,8 +5,8 @@
     Read official Department for Education information to help you become a teacher. Learn about funding, teacher training, and steps to become a teacher. 
   date: "2021-06-16"
   fullwidth: true
-  subtitle: "Get information and support to help you become a teacher."
-  subtitle_button: "Register for updates"
+  subtitle: "Whether you're just thinking about teaching or you're ready to apply, get tailored advice and insights sent straight to your inbox."
+  subtitle_button: "Get advice in your inbox"
   subtitle_link: "/mailinglist/signup/name"
   image: "media/images/content/hero-images/0012.jpg"
   navigation_title: "Home"
--- a/app/views/content/home.md
+++ b/app/views/content/home.md
@@ -5,7 +5,7 @@
     Read official Department for Education information to help you become a teacher. Learn about funding, teacher training, and steps to become a teacher. 
   date: "2021-06-16"
   fullwidth: true
-  subtitle: "Whether you're just thinking about teaching or you're ready to apply, get tailored advice and insights sent straight to your inbox."
+  subtitle: "Whether you're just thinking about teaching or you're ready to apply, get tailored advice sent straight to your email."
   subtitle_button: "Get advice in your inbox"
   subtitle_link: "/mailinglist/signup/name"
   image: "media/images/content/hero-images/0012.jpg"
--- a/app/views/content/home.md
+++ b/app/views/content/home.md
@@ -5,10 +5,10 @@
     Read official Department for Education information to help you become a teacher. Learn about funding, teacher training, and steps to become a teacher. 
   date: "2021-06-16"
   fullwidth: true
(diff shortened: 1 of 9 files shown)


gemmadallmandfe:
@EllieNodder @philspray Could one of you possibly have a look at this before Thursday and tweak the copy?

* **Hero** - do we need to change the description / CTA ahead of the promotional activity next week or leave as is?
* **Events block** - we definitely need to change the copy on this block - some of it is dummy from the design (eg new caption and badge, which we can use if we want but don't have to, plus link text)

In addition, the original idea was to point the events CTA to the About GIT events showcase page; however, this may not be appropriate now that we are halfway through the programme?

EllieNodder:
> @EllieNodder @philspray Could one of you possibly have a look at this before Thursday and tweak the copy?
> 
> * **Hero** - do we need to change the description / CTA ahead of the promotional activity next week or leave as is?
> * **Events block** - we definitely need to change the copy on this block - some of it is dummy from the design (eg new caption and badge, which we can use if we want but don't have to, plus link text)
> 
> In addition, the original idea was to point the events CTA to the About GIT events showcase page; however, this may not be appropriate now that we are halfway through the programme?

@gemmadallmandfe have tweaked the mailing list hero - feels like a good time to test some new copy? (We've already tweaked the mailing list copy from what it was in the hero when we moved it down the page so have kept that new CTA and then tweaked the copy some more now it's moving back up.) Maybe add a ticket to the backlog to review how the new copy is performing vs the old?

And have edited the events block copy and the TTA block copy to avoid repetition of the whole 'whether you're just thinking about it or... etc'. This felt better placed at the top as a better introduction to the site to give more context.

For the events block, I removed the caption (didn't feel like it added much) and the badge. Given that when we publish these changes, there won't be very many GIT events left, felt like it might be best to get rid. For this reason, I also changed the link from the about git events page to the regular events page.

On the badge - atm, it says there are 4 events in the badge at the top, but when I go to the about git events page, there are 6?

app/views/content/home.md
```diff
@@ -5,10 +5,10 @@
     Read official Department for Education information to help you become a teacher. Learn about funding, teacher training, and steps to become a teacher. 
   date: "2021-06-16"
   fullwidth: true
-  subtitle: "Whether you're ready to start your career in teaching or just curious, we can answer your questions at a Get Into Teaching event this November."
-  subtitle_button: "Get Into Teaching events"
-  subtitle_link: "/events/about-get-into-teaching-events"
-  image: "media/images/content/hero-images/event-image.jpg"
+  subtitle: "Whether you're just thinking about teaching or you're ready to apply, get tailored advice and insights sent straight to your inbox."
```
gemmadallmandfe, line 8: Is "insights" the right word here - doesn't sound quite as appealing as it perhaps could? Or should we just remove it - this is quite a long sentence compared to what we used to have?

(but this is very minor and definitely shouldn't hold up merging tomorrow!)

gemmadallmandfe (commented):
@EllieNodder Thanks for making these changes - some very minor comments but will be good to go tomorrow as planned!

I agree with linking to /events and tweaking the mailing list hero (we have a card on the backlog re testing out different CTAs)

gemmadallmandfe (approved):
Looks good to me

PR #2898, opened 2022-10-17
Editing salary info on international page

EllieNodder:
### Trello card

https://trello.com/c/YaKtj975/3713-update-salary-information

### Context

The international page says what the minimum salary is for a teacher, but doesn't address that this will be different in London.

Adding copy to make that clearer.


gemmadallmandfe (approved):
Looks good to me

PR #2912, opened 2022-10-21
Allow searching with distance but no postcode

ethax-ross:
### Trello card

[Trello-3784](https://trello.com/c/pzop8GSN/3784-address-500-errors-served-on-event-search)

### Context

We think users are entering a postcode then searching by distance and finding 0 results, prompting them to remove their postcode (instead of changing the distance to nationwide).

Allow searching by distance without a postcode; treating it as if they have selected nationwide, which is likely their intent.

- Allow searching with distance but no postcode

### DIFF
--- a/app/controllers/teaching_events_controller.rb
+++ b/app/controllers/teaching_events_controller.rb
@@ -111,7 +111,7 @@ private
 
   def setup_results
     @event_search = TeachingEvents::Search.new_decrypt(search_params)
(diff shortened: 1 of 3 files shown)


gemmadallmandfe (approved):
Looks good to me - tested:

* searching with search radius but no postcode
* searching with invalid postcode

asatwal (commented):
👍

PR #2914, opened 2022-10-21
Pull in 24 months worth of future events

ethax-ross:
### Trello card

[Trello-3786](https://trello.com/c/q5BBQJJT/3786-change-search-to-load-in-24-months-of-results-1000-type)

### Context

When we changed from the old events design to the new one we inadvertently changed the number of events we were pulling in from 24 months (1000/type) to 6 months (100/type).

Revert back to the original behaviour.

- Pull in 24 months worth of future events

### DIFF
--- a/app/models/teaching_events/search.rb
+++ b/app/models/teaching_events/search.rb
@@ -6,7 +6,7 @@ module TeachingEvents
     include EncryptedAttributes
     encrypt_attributes :postcode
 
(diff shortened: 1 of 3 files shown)


gemmadallmandfe (approved):
I can't test this due to the lack of events in dev CRM but happy to approve this on reviewing the file changes given that @ethax-ross has tested locally against prod

PR #2916, opened 2022-10-24
Changing 'teacher training advisers' blog tag to 'advisers'

EllieNodder:
### Trello card

https://trello.com/c/EDtUz2sF/3775-change-blog-tag-from-teacher-training-advisers-to-advisers

### Context

We currently have a blog tag of 'teacher training advisers'. We often use this in blogs when we're actually referring to return to teaching advisers or explore teaching advisers, so we should really change this to 'advisers' for accuracy (and to avoid duplicate link text issues).


ethax-ross (approved):

gemmadallmandfe (approved):
Looks good to me

PR #2918, opened 2022-10-26
Switch to new events search endpoint

ethax-ross:
### Trello card

[Trello-3787](https://trello.com/c/4mGKt24o/3787-switch-event-search-to-use-ungrouped-api-endpoint)

### Context

The old endpoint returned events grouped by type (which was how we used to display them) - as we display all events in the same list its easier to use the `search_teaching_events` method.

- Switch to new events search endpoint

- Switch internal events to search_events endpoint

The internal event was using the legacy `teaching_teaching_events_grouped_by_type` endpoint; switching to the new
`search_teaching_events` endpoint makes the logic cleaner and lets us remove the redundant presenter. We can also remove an unused `SearchComponent`.

### Guidance to review

As far as the user is concerned nothing should have changed. I tested against production events locally.


gemmadallmandfe (approved):
I've had a look at the front end and it looks fine to me (although obviously the number of events is limited in the dev CRM). Happy for this to be merged once you've had a tech review

I'll check with the Events team as to the overall cap required but this doesn't need to hold up the PR.

asatwal (approved):
👍 LGTM

gemmadallmandfe:
Events team are happy with an overall cap of 1000 across all event types - they will let us know if ever they get an influx of provider events in future that would mean we have more than 1000 active events in total (but this is unlikely at the moment)

PR #2919, opened 2022-10-26
Remove remaining sidebars from site

gemmadallmandfe:
### Trello card
https://trello.com/c/2CLtnTAI

### Context
As part of the navigation work, we have been replacing sidebars with Related content sections. There are a final few references to sidebars which need clearing up for consistency and also because sidebars were flagged as an issue in our DAC accessibility audit regarding confusing focus order.


gemmadallmandfe:
@ethax-ross Couple of questions for you!

1) I've removed the `right_column` reference from the frontmatter on two international pages - presume that it right

2) In the review app, the get an RTTA box is now immediately below the RTTA video - the code looks like there is a gap in between but is there something else I need to do?

image

gemmadallmandfe:
Thanks @ethax-ross - I've had a look at all the videos around the site after you removed the redundant JS / fixed the padding and they all look fine to me

ethax-ross (approved):

EllieNodder (approved):
Looks good to me

PR #2923, opened 2022-10-26
Increase app instance count

ethax-ross:
We are seeing high CPU warnings in production due to an events marketing campaign driving up traffic to the website.

Increase instance count to alleviate the extra load.


ethax-ross:
The traffic has tailed off so I think we're over the worst of it.

ethax-ross:
Re-opening as we're having a burst of comms today and expect high load on the website.

saliceti (approved):

PR #2924, opened 2022-10-27
Rework integration tests to improve success rate

ethax-ross:
### Trello card

[Trello-3819](https://trello.com/c/eBafzMCd/3819-fix-the-flakey-integration-test)

### Context

Our events integration tests are really flakey and I haven't been able to figure out why; they sometimes fail because they can't find the event and other times they can't find the sign up button, which is odd because the mailing list integration tests have been very reliable.

The purpose of the integration tests is to sanity check the end-to-end connection with the CRM; we don't need to perform mailing list and event sign ups to do that. Pairing back the integration tests to just the mailing list sign up journey should fix the flakiness without losing much in the way of test coverage.

- Rework integration tests to improve success rate


asatwal (approved):
👍

PR #2925, opened 2022-10-27
Link to Find has been tweaked to prevent confusion

Dom-Collins:
### Trello card

https://trello.com/c/ZbPPJxxO/3798-tweak-wording-of-training-courses-for-graduates-link-on-if-you-want-to-change-career

### Context

Background:
In Ways to Train Round 2, users misunderstood relevance of 'training courses for graduates' and they thought this was for recent graduates only.

Action:
Tweak wording of this link to make it clearer this is for anyone who already has a degree...

### DIFF
--- a/app/views/content/train-to-be-a-teacher/if-you-want-to-change-career.md
+++ b/app/views/content/train-to-be-a-teacher/if-you-want-to-change-career.md
@@ -22,7 +22,7 @@ You’ll need a degree and [qualified teacher status](/what-is-qts) (QTS) to wor
 
 [Now Teach](https://nowteach.org.uk/) is a service that specifically helps people get into teaching who are changing career.
 
-If you already have an undergraduate degree, you can [search for postgraduate teacher training courses](https://www.find-postgraduate-teacher-training.service.gov.uk/). There are specific training courses you can choose from that match your particular subject knowledge and experience.
+If you already have an undergraduate degree, you can [search for postgraduate teacher training courses](https://www.find-postgraduate-teacher-training.service.gov.uk/) to get QTS.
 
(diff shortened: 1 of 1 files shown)


app/views/content/train-to-be-a-teacher/if-you-want-to-change-career.md
```diff
@@ -22,7 +22,7 @@ You’ll need a degree and [qualified teacher status](/what-is-qts) (QTS) to wor
 
 [Now Teach](https://nowteach.org.uk/) is a service that specifically helps people get into teaching who are changing career.
 
-[Teacher training courses for graduates](https://www.find-postgraduate-teacher-training.service.gov.uk/) may also work for you. If you have knowledge and experience in a particular subject area there are specific training courses you can choose.
+If you already have an undergraduate degree, you can [search for postgraduate teacher training courses](https://www.find-postgraduate-teacher-training.service.gov.uk/). There are specific training courses you can choose from that match your particular subject knowledge and experience.
```
gemmadallmandfe, line 25: @EllieNodder First sentence makes much more sense to me but the second feels a bit misleading (I think the original is also a bit misleading - makes it sound like we are guaranteeing courses matching career changer experience?

EllieNodder:
```suggestion
If you already have an undergraduate degree, you can [search for postgraduate teacher training courses](https://www.find-postgraduate-teacher-training.service.gov.uk/) to get QTS.
```
agreed - do we even need that second sentence?

gemmadallmandfe (approved):
Looks good to me

PR #2926, opened 2022-10-27
Optimize HTML response transformer middleware

ethax-ross:
### Trello card

[Trello-3821](https://trello.com/c/neJ3KmBj/3821-optimise-htmlresponsetransformer)

### Context

We have middleware that performs various optimisations to our HTML responses (such as injecting responsive images, adding image sizes, etc). This is proving to be a bottleneck in our dynamic page responses. The cached, static pages take a hit on the first load but subsequent requests are very fast.

I expected the main time-sink to be Nokogiri parsing/generating HTML in each transformation library (of which we have 6). To make this more performant I've refactored it so that the document is parsed only once across all transformation libraries. I did a contrived benchmark locally and the runtime went down around 50-60%.

- Optimize HTML response transformer middleware

### DIFF
--- a/lib/error_title.rb
+++ b/lib/error_title.rb
@@ -1,15 +1,16 @@
 class ErrorTitle
-  def initialize(page)
-    @page = page
+  attr_reader :doc
+
+  def initialize(doc)
+    @doc = doc
   end
 
-  def html
-    doc = Nokogiri::HTML(@page)
+  def process
     error = doc.at_css(".govuk-error-summary")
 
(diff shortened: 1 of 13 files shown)


gemmadallmandfe (approved):
This looks OK from a front end perspective - I've had a clickaround the events section and run different search combinations

Happy for this to be merged once it has had tech eyes on it

asatwal (approved):
👍  LGTM

PR #2942, opened 2022-11-02
Adding Find link to engineers teach physics courses

EllieNodder:
### Trello card

https://trello.com/c/6co6d0bm/3777-add-link-to-find-courses-on-engineers-teach-physics-page

### Context

The new engineers teach physics courses are now up on Find so we need to add a link to our engineers teach physics page.

### DIFF
--- a/app/views/content/subjects/engineers-teach-physics/_article.html.erb
+++ b/app/views/content/subjects/engineers-teach-physics/_article.html.erb
@@ -64,7 +64,7 @@
      <p>You'll also be invited to an annual community day and have the opportunity to network with existing physics teachers.
      </p>
 
-    <p>You can <a href="https://www.find-postgraduate-teacher-training.service.gov.uk/">find teacher training courses</a> and filter by 'physics' and 'engineers teach physics'.
+    <p>You can <a href="https://www.find-postgraduate-teacher-training.service.gov.uk/">find teacher training courses</a> and filter by 'physics' then 'engineers teach physics'.
      </p>
   </section>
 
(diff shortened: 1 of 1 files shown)


app/views/content/subjects/engineers-teach-physics/_article.html.erb
```diff
@@ -64,7 +64,7 @@
      <p>You'll also be invited to an annual community day and have the opportunity to network with existing physics teachers.
      </p>
 
-    <p>Courses starting in the 2022/23 academic year are now closed. You’ll be able to find courses starting in the 2023/24 academic year in autumn 2022.
+    <p>You can <a href="https://www.find-postgraduate-teacher-training.service.gov.uk/">find teacher training courses</a> and filter by 'physics' then 'engineers teach physics'.
```
gemmadallmandfe, line 67:
```suggestion
    <p>You can <a href="https://www.find-postgraduate-teacher-training.service.gov.uk/">find teacher training courses</a> and filter by 'physics' then 'engineers teach physics'.
```
```suggestion
    <p>You can <a href="https://www.find-postgraduate-teacher-training.service.gov.uk/">search for Physics teacher training courses</a> and then filter by 'Engineers teach physics'.
```
Is this a bit easier? But happy for you to disagree @EllieNodder 

I've capitalised 'Engineers' as that is how it is written on Find - again, please disagree if necessary!

image

EllieNodder: just didn't know if we should put 'physics' in the link text, just because we're not linking them to **physics** courses but all courses - but I guess we're saying 'search for', so maybe it's okay?

gemmadallmandfe: Yes, I was a bit in two minds about that - which is why I thought search might be better than find in this case. Happy to go with whichever you think best!

gemmadallmandfe (approved):
Looks good to me

PR #2951, opened 2022-11-08
Update dfe-analytics gem

ethax-ross:
### Trello card

[Trello-3550](https://trello.com/c/zWz5rZU8/3550-remove-decorator-id-attribute-for-dfe-analytics)

### Context

Update to the latest version and remove the now-redundant `id` attribute on `User` as we can tell dfe-analytics which attribute to use (in our case `username`).

- Update dfe-analytics gem

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,7 +46,7 @@ gem "faraday_middleware"
 
 gem "fastimage"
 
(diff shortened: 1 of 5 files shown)


asatwal (approved):
LGTM 👍

PR #2954, opened 2022-11-09
Updating events promos

EllieNodder:
### Trello card

https://trello.com/c/3WP4dxDN/3863-change-promo-modules-towards-the-end-of-the-autumn-events-programme

### Context

Now that the autumn programme of GIT events is coming to an end, we should consider changing some of the events promos on our pages.

### DIFF
--- a/app/views/content/train-to-be-a-teacher/promos/_events-promo-apply.html.erb
+++ b/app/views/content/train-to-be-a-teacher/promos/_events-promo-apply.html.erb
@@ -1,8 +1,6 @@
 
 <%= render(CallsToAction::Promo::PromoComponent.new) do |promo| %>
-  <% promo.left_section(
-      heading: "Get Into Teaching events") do %>
-  <% end %>
+  <% promo.left_section(classes: %w[events-background]) %>
   <% promo.right_section(
         caption: "Online and in-person events",
         heading: "Get tips on your application") do %>
--- a/app/views/content/train-to-be-a-teacher/promos/_find-your-undergraduate-course.html.erb
+++ b/app/views/content/train-to-be-a-teacher/promos/_find-your-undergraduate-course.html.erb
@@ -6,7 +6,7 @@
   </div>
 
   <div class="right">
-    <h2 class="heading-m">Find your teacher training course</h2>
+    <h2 class="heading-m heading--margin-top-0">Find your teacher training course</h2>
     <p>
       You can search for undergraduate courses that will start you on your teaching career.
     </p>
--- a/app/views/content/train-to-be-a-teacher/promos/_find-your-undergraduate-course.html.erb
+++ b/app/views/content/train-to-be-a-teacher/promos/_find-your-undergraduate-course.html.erb
@@ -6,7 +6,7 @@
   </div>
 
   <div class="right">
-    <span class="caption-l caption--bold caption--purple">Find your course</span>
+    <h2 class="heading-m">Find your teacher training course</h2>
     <p>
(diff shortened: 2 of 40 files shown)


EllieNodder:
@gemmadallmandfe do we want to use this PR to correct all promos on site (removing headings from images, and changing captions to span instead of h2) or cover that on a new one?

app/views/content/train-to-be-a-teacher/promos/_events-promo-apply.html.erb
```diff
@@ -1,8 +1,6 @@
 
 <%= render(CallsToAction::Promo::PromoComponent.new) do |promo| %>
-  <% promo.left_section(
-      heading: "Get Into Teaching events") do %>
-  <% end %>
+  <% promo.left_section(classes: %w[events-background]) %>
   <% promo.right_section(
         caption: "Online and in-person events",
```
EllieNodder, line 5: @ethax-ross according to trello ticket 3870, we need to change the captions in these event promos to a 'span' - can't quite work out how! 🤔 mind taking a look?

app/views/content/train-to-be-a-teacher/promos/_find-your-undergraduate-course.html.erb
```diff
@@ -6,7 +6,7 @@
   </div>
 
   <div class="right">
-    <span class="caption-l caption--bold caption--purple">Find your course</span>
+    <h2 class="heading-m">Find your teacher training course</h2>
```
EllieNodder, line 9: @ethax-ross I've changed the captions here to headings (for accessibility reasons) but now all the text has shifted down and is no longer aligned with the top of the icon on the left - would you be able to take a look pls?

ethax-ross: I've updated these

app/views/content/train-to-be-a-teacher/promos/_events-promo-apply.html.erb
ethax-ross: I'll pick this change up separately as we need to change captions everywhere not just in the promo modules so its better in its own PR

EllieNodder:
@gemmadallmandfe in this PR, I've switched out the events promos to ttas/mailing list. I've also removed all the headings overlaying the images and changed the captions in the 'find your course' components to headings.

Ross will change the captions in the relevant events promos to headings in another PR ☝🏻 (see comment above)

gemmadallmandfe:
Looks good to me @EllieNodder - will approve and merge tomorrow

The only page I can think of that isn't included in this PR is `/non-uk-teachers/international-qualified-teacher-status` - not sure if you want to change this one as well as part of this overall PR?

EllieNodder:
> Looks good to me @EllieNodder - will approve and merge tomorrow
> 
> The only page I can think of that isn't included in this PR is `/non-uk-teachers/international-qualified-teacher-status` - not sure if you want to change this one as well as part of this overall PR?

My understanding of the promo changes was that we were leaving in headings/CTAs over images when they operated as actual CTAs separate to the ones on the right-hand side of the promo (like in this case) rather than just unnecessary headings like on the other pages - but I may have misunderstood! Happy to change if so

gemmadallmandfe:
> 

I think we just need to sort out the caption/heading h2 on the iQTS promo at some point

However, there is already a ticket in the backlog to change this specific promo when the Dec event has taken place - let's pick up then

gemmadallmandfe (approved):
Looks good to me

PR #2957, opened 2022-11-10
Updating orphan page titles

EllieNodder:
### Trello card

https://trello.com/c/Ax9uKyXr/3872-revise-page-titles-for-organic-search-orphan-pages

### Context

Now we've added 'GOV.UK' to all our page titles, we should review them to make sure they're not too many characters.

This PR changes the page titles for:

- Support for ECTs
- Get qualified teacher status through an 'assessment only' programme
- Get into teaching physics
- What qualified teacher status means

### DIFF
--- a/app/views/content/assessment-only-providers.md
+++ b/app/views/content/assessment-only-providers.md
@@ -1,5 +1,5 @@
 ---
-title: Get qualified teacher status through an 'assessment only' programme
+title: QTS assessment only providers
 description: |-
   Find accredited providers who offer the assessment only teacher training route to qualified teacher status (QTS) for experienced teachers with a degree.
 date: "2021-06-08"
--- a/app/views/content/assessment-only-providers.md
+++ b/app/views/content/assessment-only-providers.md
@@ -1,5 +1,5 @@
 ---
-title: QTS assessment only providers
+title: Assessment only QTS providers
 description: |-
   Find accredited providers who offer the assessment only teacher training route to qualified teacher status (QTS) for experienced teachers with a degree.
 date: "2021-06-08"
--- a/app/views/content/subjects/engineers-teach-physics.md
+++ b/app/views/content/subjects/engineers-teach-physics.md
@@ -1,5 +1,5 @@
 ---
-title: Get into teaching physics
+title: Engineers teach physics
 description: |-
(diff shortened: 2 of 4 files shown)


gemmadallmandfe:
@EllieNodder The `title` changes are fine in terms of what will be (hopefully) displayed in search results

However, the h1 has also changed (`heading` or just defaulted from `title` in some cases) on some of the pages - just wanted to make sure this was intentional before I merge this!

EllieNodder:
> @EllieNodder The `title` changes are fine in terms of what will be (hopefully) displayed in search results
> 
> However, the h1 has also changed (`heading` or just defaulted from `title` in some cases) on some of the pages - just wanted to make sure this was intentional before I merge this!

Intentional! In some cases, the URL/page title/h1 were all different - I think these should be the same where possible (both for SEO and for the user), unless there's a benefit to putting extra info in the h1.

app/views/content/assessment-only-providers.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Get qualified teacher status through an 'assessment only' programme
+title: QTS assessment only providers
```
gemmadallmandfe, line 2: Just wondering whether this quite makes sense @EllieNodder ?

I know we are going to look at whether to consolidate the two AO pages but I'm not sure this h1/title is quite right in the meantime on reflection

EllieNodder: My thinking was that the /train-to-be-a-teacher/if-you-have-experience page is talking about how and why you can do assessment only, whereas this page is basically just listing the different providers, with a bit of preamble at the top. They go to the first page to find out what, then this page to find out where.

The url is also just /assessment-only-providers, so this would make the page title/h1 more in line with that.

What were you thinking?

gemmadallmandfe: Sorry - my comment wasn't very helpful!

I was really meaning should it be 'Assessment only QTS providers' rather than QTS at the front

(NB most visitors land on the page rather than come from the other one first, looking at GA. Something to consider when we look at consolidation)

EllieNodder: ohhh, I see!! yes I think you're right - shall change

PR #2963, opened 2022-11-14
Updating train to be a teacher page titles

EllieNodder:
### Trello card

https://trello.com/c/Zdn9jsbU/3878-revise-page-titles-for-organic-search-train-to-be-a-teacher

### Context

Now we've added 'GOV.UK' to all our page titles, we should review them to make sure they're not too many characters.

This PR changes the page titles for:

- ways to train if you have a degree
- ways to train if you don't have a degree
- if you've worked in a classroom
- if you want to change career
- initial teacher training (ITT)
- improve your subject knowledge

### DIFF
--- a/app/views/content/train-to-be-a-teacher/if-you-dont-have-a-degree.md
+++ b/app/views/content/train-to-be-a-teacher/if-you-dont-have-a-degree.md
@@ -1,6 +1,6 @@
 ---
-title: "Ways to train if you don't have a degree"
-heading: "Train to be a teacher if you don't have a degree"
+title: "Train to teach if you do not have a degree"
+heading: "Train to teach if you do not have a degree"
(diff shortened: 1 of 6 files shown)


EllieNodder:
> @EllieNodder I haven't looked at this properly yet but just wanted to check that if the `title` and `heading` are the same, do we need them both? Is there any SEO benefit? Or can we just have `title` which will then be used for the `h1` as well?

not sure about any SEO benefit 🤔 I just left it in there in case we wanted to change it in future. I don't mind leaving in/taking out?

philspray:
Not sure if this helps but on previous sites I've worked on, heading had little or no SEO impact, but title did. They all look fine to me, but I wonder if the subject enhancement change might make it harder to show up in search?

PR #2965, opened 2022-11-14
remove incorrect loan repayments info

philspray:
### Trello card

https://trello.com/c/V9hCpQQ4 

### Context

We need to remove a line stating that student loan repayments will not increase if you take additional funding for teacher training as the terms of student loans have changed and this is no longer true.


gemmadallmandfe (approved):
Looks good to me!

EllieNodder (approved):
👍🏻🤩🎉

PR #2967, opened 2022-11-16
Clean up campaign layout

ethax-ross:
> ⚠️ WIP


gazcarr:
I have reviewed and all looks ok to me. I have one minor issue that could do with being sorted, but it might be more of a site wide change for headings (potentially) , so I have created a ticket in Trello:
https://trello.com/c/jCvvFk7y/3900-headings-background-to-wrap-with-text

gemmadallmandfe:
@ethax-ross I've had a look at the pages listed on desktop, on mobile in the Chrome emulator and on my iPhone 11 and generally they look fine to me - certainly good enough to merge this week once you've had a tech review

My only comments would be around:

1. **Padding** - there is a bit less breathing space around some components compared with prod. @gazcarr will look at this in more detail though and I'm very happy for this to be addressed on a separate PR!

2. **H2s/cards** - there is a bit of inconsistency on mobile as to whether h2s and cards are full width or not eg 

image

image

I think the help and support ones probably need amending?

app/components/text_block_with_one_image_component.rb
```diff
@@ -0,0 +1,16 @@
+class TextBlockWithOneImageComponent < ViewComponent::Base
+  attr_reader :heading, :colour
+
+  def initialize(heading:, image_path:, image_alt:, colour: "pink")
+    super
+
+    @heading = heading
+    @colour = colour
+    @image_path = image_path
```
asatwal, line 9: I expect `image_path` and `image_alt` are for internal to the component class and therefore don't require getters ?

ethax-ross: Yeah it looks like they're only used in the `image` method below so don't need getters (for some reason GitHub is showing this as a new file in the diff but its actually all existing code and I just moved it)

app/views/content/explore-teaching-advisers/_article.html.erb
```diff
@@ -1,5 +1,5 @@
-<article>
-  <section class="statement">
+<div class="row inset">
```
asatwal, line 1: I expect these classes such as `row`, `inset`, `col` and `col-720` etc will eventually be added to css ?

asatwal (approved):
👍  LGTM - Just a few very minor questions.

app/views/content/explore-teaching-advisers/_article.html.erb
ethax-ross: They're already in the layout CSS (may have been added in one of the later commits) https://github.com/DFE-Digital/get-into-teaching-app/pull/2967/files#diff-2fc2fec0b84bdf7c923b7bbf3015aee64123c49bb9dc0518c830a6dbf6ec65f0R170-R180

PR #2968, opened 2022-11-17
Changing who do you want to teach page titles

EllieNodder:
### Trello card

https://trello.com/c/AhAzZtSl/3879-revise-page-titles-for-organic-search-who-do-you-want-to-teach

### Context

Now that we've added 'GOV.UK' to our page titles, we should review them to try and keep them short enough to not be truncated by Google.

This PR covers the change of:

- become a further education teacher
- teaching disabled pupils and pupils with special educational needs

### DIFF
--- a/app/views/content/teach-disabled-pupils-and-pupils-with-special-educational-needs.md
+++ b/app/views/content/teach-disabled-pupils-and-pupils-with-special-educational-needs.md
@@ -1,5 +1,5 @@
 ---
-title: "Teach disabled pupils and pupils with special educational needs"
+title: "Teach pupils with special educational needs and disabilities"
 description: |-
   As a qualified teacher you’ll have the opportunity to work with pupils with complex needs, no matter what setting you work in.
 date: "2021-05-07"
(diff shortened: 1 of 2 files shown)


app/views/content/teach-disabled-pupils-and-pupils-with-special-educational-needs.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Teach disabled pupils and pupils with special educational needs"
+title: "Teach pupils with special educational needs and disabilities"
```
gemmadallmandfe, line 2: @EllieNodder 

Is it worth putting `(SEND)` at the end of the h1 at least (ie by having a separate `heading`), now that we have changed the wording to the more common phrase?
eg https://www.gov.uk/children-with-special-educational-needs

Looking in GSC, SEND does appear in search terms - not sure what you think about adding it to `title` as well as the h1 - I appreciate this will make the title even longer!

EllieNodder: good shout - shall we just add it to the h1?

gemmadallmandfe (approved):
Looks good to me

PR #2997, opened 2022-11-25
Change site name in structured data

ethax-ross:
### Trello card

[Trello-3959](https://trello.com/c/B1iLhK9S/3959-amend-structured-data-site-name-in-an-attempt-to-get-google-to-display-it)

### Context

Google is not registering out site name from the structure data. As far as we can tell its in the correct format, so we think Google may be ignoring our site name because it things we're trying to impersonate a government body (which is one of the reasons they list).

Remove `GOV.UK` from the site name to see if Google registers it.

- Change site name in structured data

### DIFF
--- a/app/helpers/structured_data_helper.rb
+++ b/app/helpers/structured_data_helper.rb
@@ -72,7 +72,7 @@ module StructuredDataHelper
 
   def home_structured_data
     data = {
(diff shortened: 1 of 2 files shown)


gemmadallmandfe (approved):
Looks good to me

PR #3020, opened 2022-12-03
Fix unused promo section

ethax-ross:
### Trello card

[Trello-3976](https://trello.com/c/z4gfDc9s/3976-fix-snyk-security-scanner-error)

### Context

We are getting a false positive error from Snyk/brakeman scanning this file as there is a syntax error on line 5. It's
not causing issues to users/our tests because the partial is unused.

Fix syntax error (we want to use this partial again the future).

- Fix events promo section

### DIFF
--- a/app/views/content/train-to-be-a-teacher/promos/_events-promo-apply.html.erb
+++ b/app/views/content/train-to-be-a-teacher/promos/_events-promo-apply.html.erb
@@ -1,8 +1,7 @@
-
 <%= render(CallsToAction::Promo::PromoComponent.new) do |promo| %>
   <% promo.left_section(classes: %w[events-background]) %>
   <% promo.right_section(
-        : "Online and in-person events",
+        caption: "Online and in-person events",
         heading: "Get tips on your application") do %>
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
@EllieNodder Tagging you in on this - would you rather delete this promo entirely (as @ethax-ross is doing in this PR) or fix the erroneous line (looks like there are two lines with headings but only one is defined as such) because you are likely to want to use it in future?

```
  <% promo.right_section(
        **: "Online and in-person events",**
        heading: "Get tips on your application") do %>
      <p>Talk to teacher training providers, expert advisers and teachers to get help writing your teacher training application.</p>
```

EllieNodder:
> @EllieNodder Tagging you in on this - would you rather delete this promo entirely (as @ethax-ross is doing in this PR) or fix the erroneous line (looks like there are two lines with headings but only one is defined as such) because you are likely to want to use it in future?
> 
> <% promo.right_section( **: "Online and in-person events",** heading: "Get tips on your application") do %>

@gemmadallmandfe @ethax-ross we're probs likely to use this promo again when events start up again - could we fix instead of removing if poss?

EllieNodder (approved):
looks good to me

PR #3033, opened 2022-12-12
Improving TTA promo page

EllieNodder:
### Trello card

https://trello.com/c/YjpbDshJ/4000-improve-teacher-training-advisers-page

### Context

The TTA service is experiencing a reduction in sign-ups compared with last year, which in turn was worse than the year before.

Following user research insight, we should update the Teacher training advisers page.

In this PR I've:

- added a section explaining how to contact an adviser
- restructured the page by consideration stage
- tweaked the quotes
- removed predicted grades for students
- removed the videos (these had barely any clicks in GA)
- added blog links similar to in is teaching right for me at the bottom of the page
- moved the 'Not a final year student or graduate?' section - having this before the blog links made it look to me like the blogs were about ETAs. I thought it might make more sense to make that section into a h3 under 'Can I get an adviser?' and move the CTA down (you can use the funnel to sign up for both TTAs and ETAs anyway).

Gary's going to tweak the blog link component for this page (and others) - I just added the current version in here for now to give a sense of how this might look. This is by no means what would actually go out (there are some issues with image sizing currently).

**Update: Having discussed the blog links with Phil, we think we should aim to publish the updates to the rest of the page without these and revisit in the new year. Naeem said he'll start to recruit for some more case studies in the new year so might be a good idea to put a pin in this for now so we can consider these, too? Now that we've added more quotes throughout the page, having the extra quotes in the blog links at the bottom might actually be overkill - to discuss in UCD.**

What I've not done in this PR:

- updated the image (but will probably get Ross to do this in this PR?)


app/views/content/teacher-training-advisers.md
```diff
@@ -49,68 +43,88 @@ youtube_video:
...
+You’ll be matched with your own dedicated adviser. You’ll have the same adviser throughout your journey into teaching.
+How you talk to them is up to you – you can chat through:
 
-* arranging school experience
-* seeking out financial support
-* writing a personal statement
-* preparing for interviews
+* email
+* phone
+* text
```
gemmadallmandfe, line 62: Don't they use WhatsApp as well @EllieNodder ? One to check with Andy

app/views/content/teacher-training-advisers.md
```diff
@@ -49,68 +43,88 @@ youtube_video:
...
 
-## Meet Emma, a teacher training adviser
+You can talk to them as little or as often as you need. This could be every week for a year, or just a few times before making an application.
```
gemmadallmandfe, line 64: People can use advisers **after** they make an application too - eg for interview prep, or if they get rejected first time round and want to apply again. Does this need rewording slightly @EllieNodder ?

app/views/content/teacher-training-advisers.md
```diff
@@ -49,68 +43,88 @@ youtube_video:
...
-## What an adviser can do for you
+They have years of teaching experience and can answer all your questions about life in the classroom, as well as how you can get there.
 
 $q-tta-one$
 
-An adviser can give you as much or as little help as you need. 
+## How does it work?
 
-Other trainee teachers say they are particularly helpful with: 
+You’ll be matched with your own dedicated adviser. You’ll have the same adviser throughout your journey into teaching.
```
gemmadallmandfe, line 57: Bit repetitive to say 'journey into teaching' again here @EllieNodder ?

app/views/content/teacher-training-advisers.md
gemmadallmandfe, line 64: Does this imply that a year is the maximum @EllieNodder ? I don't think that is true so this might need rewording?

app/views/content/teacher-training-advisers.md
```diff
@@ -49,68 +43,88 @@ youtube_video:
...
-## Meet Emma, a teacher training adviser
+You can talk to them as little or as often as you need. This could be every week for a year, or just a few times before making an application.
 
-$about-teacher-training-advisers-video$
+## How your adviser can help you
 
-## Helping you decide if teaching is right for you
+### If teaching is just an idea
 
 $q-tta-two$
```
gemmadallmandfe, line 70: I wonder if the quotes would be better at the end of each section rather than under the h3 @EllieNodder? It makes it very disjointed on mobile as it currently is
image

app/views/content/teacher-training-advisers.md
```diff
@@ -49,68 +43,88 @@ youtube_video:
...
-## Meet Simon, a teacher training adviser
+* write your personal statement
+* understand your different training options
+* prepare for interviews
 
-$about-teacher-training-advisers-experience-video$
+### If you've applied and been unsuccessful
 
-## Already applied and been unsuccessful?
+$q-tta-three$
```
gemmadallmandfe, line 92: Is this quote in the right place? Feels more like it belongs in the section before?

app/views/content/teacher-training-advisers/_blog_links.html.erb
```diff
@@ -0,0 +1,27 @@
+  <section>
```
gemmadallmandfe, line 1: I agree that these blocks work as is @EllieNodder - would suggest taking them out for now until @gazcarr has thought about a new component

gemmadallmandfe (commented):
Looks much better @EllieNodder! Some initial thoughts from me before you send to Andy for review

EllieNodder:
@gazcarr @ethax-ross the new image kinda cuts off the top half of their faces - is there a way to fix this?

app/views/content/teacher-training-advisers.md
```diff
@@ -29,88 +16,115 @@ keywords:
...
+
 $get-a-tta$
 
+### If you're a veteran
+
+If you're a veteran, you do not need to have or be studying for a degree to get support from an adviser.
+
+Call us on <a href="tel:08003892500">0800 389 2500</a> to get an adviser.
+
+Our telephone service is available Monday to Friday between 8:30am and 5:30pm.
```
gemmadallmandfe, line 122:
```suggestion
Our telephone service is available Monday to Friday between 8:30am and 5:30pm, except on bank holidays.
```
(@EllieNodder needs link adding as well to gov.uk page as per footer component)

app/views/content/teacher-training-advisers.md
```diff
@@ -29,88 +16,115 @@ keywords:
...
-* preparing for interviews
+How often you talk to them is up to you – this could be every week for more than a year, or just a few times for some one-to-one support.
+
+You can chat by phone, or by:
+
+* email
+* text
+* WhatsApp
+
+## How your adviser can help you
```
gemmadallmandfe, line 58: This is the only h2 on the page that isn't a question - should it be reworded to a question for consistency @EllieNodder?

gemmadallmandfe (approved):
Looks good to me

PR #3037, opened 2022-12-12
Editing link text to make telephone links clearer

philspray:
### Trello card

https://trello.com/c/eUf7bdzY

### Context

Some links were not descriptive enough for screen reader users when navigating out of the context.

Added more context to link text around telephone numbers


app/views/content/teacher-training-advisers.md
```diff
@@ -99,7 +99,7 @@ To sign up, you’ll need (one of the following):
 * to be a final year student studying for a bachelor’s degree with honours, with a predicted grade of 2:2 or higher
 * an overseas qualification equivalent to a bachelor’s degree with honours, grade 2:2 or higher
 
-If you're applying from overseas, you can check your qualifications by calling us on <a href="tel:08003892500">0800 389 2500</a>.
+If you're applying from overseas, you can check your qualifications by <a href="tel:08003892500">calling us on 0800 389 2500</a>.
```
gemmadallmandfe, line 102: @philspray This page has changed since you created the PR (sorry!) so needs rebasing

There are now two phone number links on the page to change (international qualifications and veterans)

app/components/content/talk_to_us_component.html.erb
```diff
@@ -7,7 +7,7 @@
       <div>
         <h3 class="heading-l">Call us</h3>
         <p>Our phone line gives you a quick and easy way to get one-off support.</p>
-        <p><a href="tel:08003892500" class="telephone">0800 389 2500</a></p>
+        <p><a href="tel:08003892500" class="telephone">Call us on 0800 389 2500</a></p>
```
gemmadallmandfe, line 10: @philspray What's your view on this change? It feels quite repetitive under the `Call us` h3 - I wonder if we are OK to not make this change here since the page is called 'Get help and support' and we have an h2/h3 as well that indicate what this link is?

gemmadallmandfe (commented):
@philspray Thanks for this - couple of minor comments for you - plus I think there are some other pages that need similar changes:

* /funding-and-support/if-youre-a-veteran
(NB this link doesn't look like it is styled properly as a phone number so this needs changing too)

* /mailinglist/signup/completed

app/components/content/talk_to_us_component.html.erb
philspray: I agree, I think it's too repetitive with anything else. I've tweaked it.

app/views/content/non-uk-teachers/teach-in-england-if-you-trained-overseas.md
```diff
@@ -158,6 +158,12 @@ Most teaching vacancies are advertised in March, April or the first half of May,
 
 Set up a Teaching Vacancies job alert to be updated about opportunities throughout the year.
 
+You can search for teaching jobs by region, subject specialism and age group.
```
gemmadallmandfe, line 161: @philspray Something odd is going on here - this PR has a LOT of commits and this content has been duplicated on this page 

image

Firstly, this change needs removing - and then I think it might be worth just flagging with Ross to make sure nothing else weird is going on!

app/views/content/teacher-training-advisers.md
```diff
@@ -109,15 +109,15 @@ To sign up, you’ll need (one of the following):
 
 If you’re a returning teacher who already has qualified teacher status (QTS), you'll need to be returning to teach at a secondary school in England to get an adviser.
 
-If you completed your degree overseas, you can check your qualifications by calling us on <a href="tel:08003892500">0800 389 2500</a>.
+If you completed your degree overseas, you can check your qualifications by <a href="tel:08003892500">calling us on 0800 389 2500</a>.
```
gemmadallmandfe, line 112: @philspray I think this is slightly odd as link text ie `calling` rather than `call` but I can't think of an alternative unless you say `call us on xxxx to check your qualifications` which might be a bit longwinded?

app/views/mailing_list/steps/completed.html.erb
```diff
@@ -19,8 +19,7 @@
     <div class="mailing-list__actions">
       <%= link_to("Book a callback", callbacks_steps_path, class: "button") %>
       <div>
-        <small>or call us now:</small>
-        <p class="telephone">0800 389 2500</p>
+        <p class="telephone">or call us now on 0800 389 2500</p>
```
gemmadallmandfe, line 22: @philspray This isn't actually a link (it is not in prod either) so is not clickable on mobile - so I'm not sure the change is worth it as is (and actually it makes the formatting a bit funny)

Should we change it to be a link? Or just keep the original with the two lines?

image

PR #3044, opened 2022-12-14
Changing international funding card description

EllieNodder:
### Trello card

https://trello.com/c/X34kwyg6/3986-rename-international-card-on-funding-and-support-category-page

### Context

In UR, we found that users were confused by this card on the fund your training page - the heading and the description suggest different things.

We should change the description to make the content clearer.


philspray (approved):
Looks good to me too!

PR #3049, opened 2022-12-16
Update teach-in-england-if-you-trained-overseas.md

laurahermionetennant92:
Removed mention of Get an adviser


philspray (approved):
Looks good to me.

PR #3053, opened 2022-12-19
Fix colour contrast warning on WAVE tool

ethax-ross:
### Trello card

[Trello-4035](https://trello.com/c/0MQH4IOi/4035-investigate-colour-contrast-failure-flagged-by-webaim)

### Context

The WAVE chrome extension reports a colour contrast issue on the steps heading/text; it incorrectly thinks we are rendering white text on a white background.

It also picks up on a contrast issue with the skip to main content link, but as this is hidden until you focus on it it's a non-issue (after focusing there is no colour contrast problem).

We can silence this warning by giving the elements an explicit purple background.

- Fix colour contrast warning on WAVE tool

### Guidance to review

There's one remaining colour contrast issue on the skip to main content l

### DIFF
--- a/app/webpacker/styles/header/skiplink.scss
+++ b/app/webpacker/styles/header/skiplink.scss
@@ -1,6 +1,7 @@
 .skiplink {
   position: absolute;
   left: -9999em;
+  color: $white;
 
   &:focus {
     left: 0;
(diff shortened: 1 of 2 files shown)


gemmadallmandfe:
Looks good to me @ethax-ross - tested successfully using WebAim WAVE

I haven't ticked it yet as it can't go out until the freeze is over but can be merged first thing in January

(As an aside, is it actually worth fixing, given that it is not actually something that causes a problem for users, just for automated testing results?)

gemmadallmandfe (approved):
Happy for this to be deployed to prod @ethax-ross

PR #3054, opened 2022-12-20
Forward query params to cookie preferences

ethax-ross:
### Trello card

[Trello-4052](https://trello.com/c/F4KZFcqG/4052-investigate-advertising-tagging-issues-for-users-who-opt-out-of-cookies)

### Context

Users will often arrive on the website with tracking query parameters in the URL. As no tags fire until they accept cookies we find that if they go to cookie preferences and then accept the original tracking query parameters are lost.

Forward query parameters to the cookie preferences page, so that if the user then accepts cookies we get the tracking information firing.

- Forward query params to cookie preferences

### DIFF
--- a/app/components/footer/cookie_acceptance_component.html.erb
+++ b/app/components/footer/cookie_acceptance_component.html.erb
@@ -28,9 +28,10 @@
         <a tabindex="-1" href="#" class="button call-to-action-button" data-action="click->cookie-acceptance#accept" data-cookie-acceptance-target="agree" id="biscuits-agree">
           <span>Accept all cookies</span>
         </a>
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
@ethax-ross I've tried testing this following the steps outlined in Omnigov's email

However, I can't see the gclid being saved if you go via cookie preferences in the review app. This might be because GA is set up different in this environment?

If you agree, I'm happy for this to be deployed to prod when you return in Jan and we can test it properly there (we should coordinate this with Omnigov).

gemmadallmandfe (approved):
Happy for this to be deployed to production this morning (if you are) for Omnigov to test @ethax-ross

PR #3070, opened 2023-01-09
Prevent cookie acceptance on privacy page with query param

ethax-ross:
On the privacy pages (cookie, cookie acceptance, privacy policy) we should never show the cookie acceptance dialog as it obscures the content. This usually is the case, however, when there is a query parameter in the URL the logic fails to match as we are matching a path against a path with a query string.

Match on path without query string.


gemmadallmandfe (approved):
I've tested this in the review app and it looks good to me

PR #3071, opened 2023-01-10
Fix retaining query params on cookie prefs link

ethax-ross:
### Trello card

[Trello-4052](https://trello.com/c/F4KZFcqG/4052-investigate-advertising-tagging-issues-for-users-who-opt-out-of-cookies)

### Context

We were previously passing the query parameters through as part of the template rendering, but our static page cache prevents this from updating properly.

Instead, we need to update the path in Javascript when we sho the dialog.

- Fix retaining query params on cookie prefs link


asatwal (approved):
👍 LGTM

PR #3098, opened 2023-01-30
Fixing disability rights uk link on if youre disabled page

EllieNodder:
### Trello card

https://trello.com/c/KbkBfObZ/1113-fix-broken-link-on-if-youre-disabled-page

### Context

The link for disability rights UK on our if you're disabled page is currently broken so we should replace.


gemmadallmandfe (approved):
Looks good to me

PR #3117, opened 2023-02-09
Removing how to fund your training blog

EllieNodder:
### Trello card

https://trello.com/c/oUXB0ZyR/4213-update-how-to-fund-your-training-blog

### Context

Our how to fund your training blog is currently very out of date and is not a high traffic page. We should delete the blog and redirect to our funding page.


gemmadallmandfe:
Redirect tested successfully :-)

Do we want to clear up the blog tags at the same time @EllieNodder ?

It looks to me like none of these are used on any other blog
image

Should we remove them from the blog tag config file as good housekeeping practice?

gemmadallmandfe (approved):
Looks good to me

PR #3138, opened 2023-02-22
Update integration tests for new ML journey

ethax-ross:
The mailing list now has an additional step we need to account for in the integration tests.


EllieNodder (approved):
looks good to me

PR #3139, opened 2023-02-22
Disable integration tests

ethax-ross:
These have started failing as the test can't find the new radio button on the returning teacher exclusion step; I can't immediately see why and we need to ship this change so I'll revisit this.


gemmadallmandfe (approved):
Happy for this to be merged and then investigated

PR #3143, opened 2023-02-23
Downgrade Knapsack

ethax-ross:
### Trello card

[Trello-4261](https://trello.com/c/O6i5vjyl/4261-remove-knapsack-pro)

### Context

We were using a free tier of Knapsack Pro but they've recently asked us to move onto a paid plan. Either their policy has changed or we have exhausted the free tier limits.

Downgrade to regular Knapsack, which still provides parallelisation, just not quite as optimised as the pro version.

- Downgrade Knapsack


asatwal (approved):
👍 LGTM

PR #3156, opened 2023-03-01
Prevent caching of JS/CSS assets in development

ethax-ross:
### Trello card

[Trello-4299](https://trello.com/c/rzc4Pc09/4299-fix-shakapacker-not-reflecting-updated-css-js-in-development)

### Context

By default Shakapacker generates applicaton.js/css without any digest. This causes the browser to serve the cached version of the asset even after the contents have changed.

- Prevent caching of JS/CSS assets in development

Change webpack to always generate JS/CSS asset filenames with a digest.

Also update webpack overlay to only show errors as we have deprecation warnings in dependencies we can't do anything about.

### DIFF
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -20,4 +20,15 @@ const options = {
   },
 }
 
+// add hashing of generated JS and CSS files to fix caching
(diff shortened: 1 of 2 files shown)


asatwal (approved):
👍

PR #3157, opened 2023-03-02
Support CORS preflight requests

ethax-ross:
### Trello card

[Trello-4264](https://trello.com/c/IBdTsDhA/4264-investigate-404-landing-pages-for-adviser-service)

### Context

A CORS preflight request is a CORS request that checks to see if the CORS protocol is understood and a server is aware using specific methods and headers.

It is an OPTIONS request, using three HTTP request headers: Access-Control-Request-Method, Access-Control-Request-Headers, and the Origin header.

We currently respond to all OPTIONS requests with a 404; instead, we can include the `rack-cors` gem to correctly service these requests.

- Support CORS preflight requests

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -97,6 +97,8 @@ gem "git_wizard", github: "DFE-Digital/get-into-teaching-wizard"
 
 gem "rack-host-redirect"
 
+gem "rack-cors"
+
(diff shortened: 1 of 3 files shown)


asatwal (approved):
👍

PR #3159, opened 2023-03-02
Remove cache read metric

ethax-ross:
### Trello card

[Trello-4256](https://trello.com/c/JE8s3iFB/4256-look-to-see-if-we-can-remove-metrics-to-alleviate-the-load-on-prometheus)

### Context

We are overloading our InfluxDB instance due to the amount of metrics we are processing. We don't tend to look at the cache read metric and its likely one of the busier ones so removing it to alleviate some of the load.

- Remove cache read metric


asatwal (approved):
👍

PR #3162, opened 2023-03-07
Update copy on mailing list postcode step

ethax-ross:
### Trello card

[Trello-4276](https://trello.com/c/8Cf2f1J7/4276-update-postcode-step-to-avoid-confusion-for-international-users)

### Context

The postcode step confuses international candidates as we only accept UK postcodes, but this isn't clear at the moment.

- Update copy on mailing list postcode step

Update the title to include 'UK' and add a helper section for international candidates.


EllieNodder (approved):
Looks good to me! 🤩

PR #3163, opened 2023-03-07
Updating SKE card description and CTA

EllieNodder:
### Trello card

https://trello.com/c/kx3XWkhz/4307-reword-ske-card-description

### Context

The current SKE card description doesn't actually mention the term 'subject knowledge enhancement course' when we know from keyword research and UR that users are often familiar with it.

This PR adds the term to the card description and tweaks the CTA.


gemmadallmandfe (approved):
Looks good to me!

PR #3164, opened 2023-03-07
Fix mailing list integration tests

ethax-ross:
[Trello-4258](https://trello.com/c/5DcZLZxH/4258-fix-mailing-list-integration-tests)

For some reason the 'No' label/radio isn't found by Capybara using `choose "No"`. If we find the label explicitly and click it then it appears to work.


asatwal (approved):
👍

PR #3167, opened 2023-03-08
Fix link color for accessibility

ethax-ross:
### Trello card

[Trello-4179](https://trello.com/c/uXHvcGZj/4179-fix-colour-contrast-issue-on-help-and-support-page)

### Context

There is not enough contrast in this link with the background; using our standard dark link color should fix that.

- Fix link color for accessibility


gemmadallmandfe (approved):
Looks good to me

PR #3168, opened 2023-03-08
No-index blog tag pages

ethax-ross:
### Trello card

[Trello-4308](https://trello.com/c/auA2NsRE/4308-add-noindex-to-the-blog-tag-pages)

### Context

The blog tag pages have the same intro paragraph and could be considered duplicate content according to Google Search Console. We have also not reviewed the tags yet so there are likely improvements that need to be made in this area. In the meantime we don't want to index these pages with search engines.

- No-index blog tag pages


gemmadallmandfe (approved):
Looks good to me

PR #3170, opened 2023-03-08
Enable Google Optimize on homepage

ethax-ross:
We want to run an experiment on the homepage; enable Google Optimize so the script is included.

### DIFF
--- a/config/google_optimize.yml
+++ b/config/google_optimize.yml
@@ -7,6 +7,7 @@
 #   - /
 
 paths:
+  - /
   - /salaries-and-benefits-search
   - /salaries-and-benefits-social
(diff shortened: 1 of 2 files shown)


gemmadallmandfe:
@ethax-ross Can we close this PR now?

ethax-ross:
@gemmadallmandfe Ramone was just asking me about this and it sounded like we're going ahead with the landing home page test, so will we not want this still?

gemmadallmandfe:
@ethax-ross was thinking this was the blur-specific version but of course this PR is just adding Optimize to the page

Yes, we are still going ahead with a home page test (just a different one than before!)

ethax-ross:
Closing this in favour of #3224

PR #3173, opened 2023-03-09
tweaking non-UK meta descriptions

philspray:
### Trello card

https://trello.com/c/TwtllO6w

### Context

We need to update the meta descriptions of non-UK pages to make sure it is clear this is for those outside of the UK. We have also updated to use non-UK rather than international or overseas.

- making it clear these pages are for non-UK users
- changing international or overseas to non-UK


laurahermionetennant92:
Happy with this, thanks @philspray

laurahermionetennant92 (approved):

PR #3177, opened 2023-03-14
Creating 'If you have a degree' landing page

EllieNodder:
### Trello card

https://trello.com/c/4thyXIQG/4267-create-and-test-a-new-if-you-have-a-degree-landing-page-for-paid-traffic-test-4

### DIFF
--- a/app/views/content/landing/if-you-have-a-degree.md
+++ b/app/views/content/landing/if-you-have-a-degree.md
@@ -1,0 +1,15 @@
+---
+title: "Train to be a teacher if you have a degree"
+description: find out how to train to be a teacher if you have a degree, including what qualifications you can get and how to fund your training.
+content:
+    - content/landing/if-you-have-a-degree/header
+    - content/landing/if-you-have-a-degree/collage
+    - content/landing/if-you-have-a-degree/mailing_list
+    - content/landing/if-you-have-a-degree/content
+    - content/landing/if-you-have-a-degree/promo
+image: "static/content/hero-images/M_DFE_Southfeilds_Room_A360_10445.jpg"
+colour: "yellow"
+layout: "layouts/minimal"
+talk_to_us: false
+noindex: true
+---
--- a/app/views/content/landing/if-you-have-a-degree/_collage.html.erb
+++ b/app/views/content/landing/if-you-have-a-degree/_collage.html.erb
@@ -1,0 +1,25 @@
+<div class="row">
+  <div class="grey col">
+    <section class="col col-full-content">
+      <div class="inset">
+        <p class="col-720">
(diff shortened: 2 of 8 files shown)


app/views/content/landing/if-you-have-a-degree.md
```diff
@@ -0,0 +1,15 @@
+---
+title: "Train to be a teacher if you have a degree"
```
gemmadallmandfe, line 2: this looks slightly overbearing on so many lines @EllieNodder !

Would it work to just call it 'Train to be a teacher' or 'Train to teach'?

It is only a landing page (so we are directing traffic to it) so h1 less important for search. It is also not a category card page on Train to be a teacher so doesn't need to have different h1s/titles from that perspective?

app/views/content/landing/if-you-have-a-degree.md
```diff
@@ -0,0 +1,15 @@
+---
```
gemmadallmandfe, line 1: should we include `train-to-be-teacher` in the URL @EllieNodder ?

I'm conscious that we are building out a lot of landing pages and we may need to make sure it is really clear which is which - let's discuss?

app/views/content/landing/if-you-have-a-degree/_collage.html.erb
```diff
@@ -0,0 +1,25 @@
...
+        If you have a degree, you can do postgraduate teacher training to get qualified teacher status (QTS). Postgraduate teacher training:
+        </p>
+
+        <%= render Content::ChecklistCollageComponent.new(
+          checklist: [
+            "takes 9 months full time, or 18 to 24 months part time",
+            "typically includes 120 days of classroom experience",
+            "involves both practical and theoretical learning"
+          ],
+          image_paths: [
```
gemmadallmandfe, line 15: are we going to remove the images @EllieNodder?

EllieNodder:
@gemmadallmandfe @philspray do we think we might need a link to the 'if you do not have a degree' page somewhere? Given that the users clicking this link won't be seeing the 'if you have a degree' card on the category page and may have just searched a generic 'how to train to teach', but this page is only talking about postgrad?

(also, I may have broken the PR by trying to remove the images ...)

gemmadallmandfe:
> @gemmadallmandfe @philspray do we think we might need a link to the 'if you do not have a degree' page somewhere? Given that the users clicking this link won't be seeing the 'if you have a degree' card on the category page and may have just searched a generic 'how to train to teach', but this page is only talking about postgrad?
> 
> (also, I may have broken the PR by trying to remove the images ... have now called Ross in for back up)

Do we really want lots of links though? Are we better to link people to the Train to teach category page rather than the If you have a degree page?

I think it is probably worth having a conversation with Karen/Michael too?

PR #3190, opened 2023-03-22
Updates for A/B tests

ethax-ross:
- Add Google Optimize to mailing list page

We need Optimize running on the mailing list start page for the salaries and bursaries landing page test.

- Create copy of if you have a degree page for testing

We need a copy of the if you have a degree page that we can use for targeting in an A/B test.

Add page, noindex it and add Google Optimize.

[Link to page](https://review-get-into-teaching-app-3190.london.cloudapps.digital/train-to-be-a-teacher/if-you-have-a-degree-search)


gemmadallmandfe (approved):
Looks good to me

PR #3192, opened 2023-03-23
Irp academic year tweak

laurahermionetennant92:


laurahermionetennant92:
Please review and deploy @philspray :)

laurahermionetennant92:
Yes it is – putting the CTA there is designed to capture the people who can't be bothered to read the eligibility criteria (loads, probably) – for us it is still valuable them signing up even if we get some that aren't eligible. But the people who do read the criteria also need the info.

philspray (approved):
Looks ok to me

PR #3197, opened 2023-03-27
removing talk to us from pages

philspray:
### Trello card

https://trello.com/c/jw1vklmt

### Context

User feedback highlighted the Talk to us component is misleading on the Early years and Further education pages.

Removing the Talk to us component on the Early years and Further Education page


gemmadallmandfe:
@philspray I can see you've made changes but this isn't reflected on the front end!

Suspect there is more to it - and that Ross might be needed (not tagging him yet as he is busy with Apply!)?

philspray:
Yes, I've messaged Ross for advice and he's going to take a look when he gets a chance.

gazcarr:
@ethax-ross didn't we add a grey border to the top of the feedback form at some point? 

@gemmadallmandfe if no one is using the feedback form on these pages then probably safe to remove as we have done so on other pages

EllieNodder:
@philspray I'd say probs remove the feedback bar from these pages as it does look a bit odd?

gemmadallmandfe:
@gazcarr It was through the feedback form that people told us there was a problem with the Talk to us component :-D

However, there is also a feedback link in the footer on the Early years and FE pages (overlooked this in my earlier comment!), so I think it is fine to remove the Feedback bar on these two pages @philspray; people will still be able to leave feedback if needs be

gemmadallmandfe (approved):
Looks good to me

PR #3199, opened 2023-03-28
Updating steps page title and description

EllieNodder:
### Trello card

https://trello.com/c/v6DPjwpe/4365-change-title-and-meta-description-for-steps-page-for-organic-search

### Context

How to become a teacher is a much higher search term than steps to become a teacher - while we do the bigger steps work in the background, it's a good idea to update the page title and meta description in the mean time to make sure we're capturing this term to try and pick up more organic traffic.


philspray (approved):
Looks good to me :)

gemmadallmandfe (approved):
Looks good to me

PR #3201, opened 2023-03-29
Updating mailing list step in events sign up

EllieNodder:
### Trello card

https://trello.com/c/PuQSyO66/4375-improve-postcode-content-and-error-message-in-events-funnel-to-avoid-confusion-for-international-users

### Context

Updating the postcode copy to make it clearer for international users and also improving the content.

### DIFF
--- a/app/helpers/option_set_helper.rb
+++ b/app/helpers/option_set_helper.rb
@@ -1,0 +1,6 @@
+module OptionSetHelper
+  def degree_status_text(option)
+    key = option.value.parameterize(separator: "_")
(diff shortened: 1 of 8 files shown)


philspray (approved):
A big improvement! Looks good to me.

gemmadallmandfe:
@EllieNodder @ethax-ross Presumably changing the dropdown options for the Do you have a degree question doesn't impact the values sent to CRM - ie we don't need to do any additional testing with the CRM team?

If there are no changes to what is sent via the API, happy for this to be merged @EllieNodder (if you can switch the order of the error messages first, even better!)

gemmadallmandfe (approved):
Looks good to me

PR #3202, opened 2023-03-29
Updating postcode error message

EllieNodder:
### Trello card

https://trello.com/c/xfqyBTaJ/4414-improve-postcode-error-message-on-mailing-list-funnel

### Context

We should update our postcode error message so that it specifies it must be a 'UK' postcode.


gemmadallmandfe (approved):
Looks good to me

PR #3205, opened 2023-03-30
Send request to dfe analytics in middleware for cached pages

asatwal:
### Trello card

[Trello-1058](https://trello.com/c/6b49g4yk/1058-bug-git-website-home-page-views-in-ga-much-greater-than-in-events-table-in-bq) 

### Context

We have a big inconsistency tracking homepage visits on the GiT website through BigQuery.

BQ Count: 1637 vs GA Count: 65,263. Please see ticket for details.

After some troubleshooting we found that there is  page caching  of static pages, the homepage being a static page.

GiT website uses [rack-page_caching](https://github.com/pkorenev/rack-page_caching) GEM, that implements caching through rack middleware.

The DfE Analytics syncs web request events with a controller after action callback.

With the rack middleware being low level intercept and the controller action higher level, there is no simple solution that comes to mind.

One solution would be to have some  JS that gets called on Page load (of the cached page) and calls an API.

However, this would be a very manual non-trivial solution and goes against the grain of the DfE Analytics GEM (ie install, configure and forget)

We have added `DfeAnalytics` Middleware that intercepts before the `ActionDispatch:Static` (This middleware returns a cached page if there is one and returns from the middleware stack early).

The `DfeAnalytics` middleware checks if there is a cached page in the file system. If there is a cached page, then the correct type of request/response objects are instantiated and the web request event is sent to Big Query.

### Guidance to review

`DfeAnalytics` middleware is only added if page caching is enabled.

The file handling logic in `DfeAnalytics` is copied from the `ActionDispatch:Static` class. Please see [here](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/static.rb) 

Tests are still outstanding. Suggestions are most welcome.

### DIFF
--- a/config/initializers/middleware.rb
+++ b/config/initializers/middleware.rb
@@ -1,4 +1,5 @@
 require "middleware/html_response_transformer"
+require "middleware/dfe_analytics"
 
 # Page caching middleware
 Rails.application.config.middleware.use Rack::PageCaching,
@@ -14,7 +15,13 @@ Rails.application.config.middleware.use(Middleware::HtmlResponseTransformer)
 if Rails.application.config.action_controller.perform_caching
   # Serve the static page cache
   Rails.application.config.action_controller.page_cache_directory = Rails.root.join("public/cached_pages")
+
   Rails.application.config.middleware.insert_before \
     ActionDispatch::Static, ActionDispatch::Static, File.join(Rails.application.config.root, "public", "cached_pages"),
+    headers: { "Cache-Control" => "max-age=#{5.minutes.to_i}, public, immutable" }
+
+  # Send request event to dfe analytics if static page is served
+  Rails.application.config.middleware.insert_before \
+    ActionDispatch::Static, Middleware::DfeAnalytics, File.join(Rails.application.config.root, "public", "cached_pages"),
     headers: { "Cache-Control" => "max-age=#{5.minutes.to_i}, public, immutable" }
 end
--- a/lib/middleware/dfe_analytics.rb
+++ b/lib/middleware/dfe_analytics.rb
@@ -1,0 +1,33 @@
+module Middleware
+  class DfeAnalytics
+    def initialize(app, path, index: "index", headers: {})
+      @app = app
+      @file_handler = ActionDispatch::FileHandler.new(path, index: index, headers: headers)
+    end
+
+    def call(env)
+      # Detect if page is Cached and send request event accordingly
+      send_request_event(env) if @file_handler.attempt(env)
+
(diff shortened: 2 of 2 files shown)


config/initializers/middleware.rb
```diff
@@ -14,7 +15,13 @@
...
   # Serve the static page cache
   Rails.application.config.action_controller.page_cache_directory = Rails.root.join("public/cached_pages")
+
   Rails.application.config.middleware.insert_before \
     ActionDispatch::Static, ActionDispatch::Static, File.join(Rails.application.config.root, "public", "cached_pages"),
     headers: { "Cache-Control" => "max-age=#{5.minutes.to_i}, public, immutable" }
+
+  # Send request event to dfe analytics if static page is served
+  Rails.application.config.middleware.insert_before \
+    ActionDispatch::Static, Middleware::DfeAnalytics, File.join(Rails.application.config.root, "public", "cached_pages"),
```
ethax-ross, line 25: Not sure the headers are needed here; from memory we're using the headers in the page cache middleware to tell it to serve content with a cache-control of 5 minutes. I'm _fairly_ certain that under the hood it caches the page indefinitely (until a deployment essentially clears the cache/files).

ethax-ross:
I think this looks good generally - I wonder if we should tie it into the gem we're using instead of checking the file existence manually. Something like this may work:

```
response = Rack::PageCaching::Response.new(rack_response, env)
track_analytics_event if response.cacheable?
```

In terms of testing the only thing I could come up with is a bit contrived but does test the middleware logic:

```
require "rails_helper"
require "middleware/dfe_analytics"

describe Middleware::DfeAnalytics, type: :request do
  let(:url) { "http://example.com/test" }
  let(:app) { ->(env) { [200, env, "app"] } }

  it "does not send a web_request event for dynamics pages" do
    Dir.mktmpdir do |dir|
      middleware = described_class.new(app, dir)
      expect { middleware.call(env_for(url)) }
        .not_to have_sent_analytics_event_types(:web_request)
    end
  end

  it "sends a web_request event for cached pages" do
    Dir.mktmpdir do |dir|
      path = "#{dir}/test.html"
      File.write(path, "test page")
      middleware = described_class.new(app, dir)
      expect { middleware.call(env_for(url)) }
        .to have_sent_analytics_event_types(:web_request)
      File.delete(path)
    end
  end

  def env_for(url, opts={})
    Rack::MockRequest.env_for(url, opts)
  end
end
```

asatwal:
Change is been implemented in DfE Analytics GEM. No longer required in the GIT App.

PR #3208, opened 2023-03-31
Update git_wizard

ethax-ross:
[Trello-4372](https://trello.com/c/iDqBrbp4/4372-clean-up-existingcandidaterequest-model)

The update removes first/last name from the matchback requests; the API no longer uses these fields and instead matches back only on email.


ethax-ross:
Closing this; it turns out we still need first name for the email to go out successfully.

PR #3218, opened 2023-04-03
Send request events for cached pages

asatwal:
### Context
We have a big inconsistency tracking homepage visits on the GiT website through BigQuery.

BQ Count: 1637 vs GA Count: 65,263. Please see ticket for details.

After some troubleshooting we found that there is page caching of static pages, the homepage being a static page.

GiT website uses [rack-page_caching](https://github.com/pkorenev/rack-page_caching) GEM, that implements caching through rack middleware.

The DfE Analytics syncs web request events with a controller after action callback.

Cached pages are served by rack middleware, returning early and therefore do not execute any actions in the controller. This means that any cached page visits handled by rack middleware do NOT result in a web request event being sent to Big Query.

Add a custom  proc which will be called with the rack env, that returns a boolean indicating whether the page is cached and will be served by rack middleware.

### Guidance to review

This changes use a version of the dfe-analytics GEM from a [PR](https://github.com/DFE-Digital/dfe-analytics/pull/67) that is still under development, for testing purposes only.

Once the dfe-analytics GEM has been merged and released, we will then update this PR to to use the released version.

This change has been testing manually only. It is difficult to test in `rspec` as the caching is set through `Rails.application.config.action_controller.perform_caching` at startup time. I haven't been able to alter the value of `Rails.application.config.action_controller.perform_caching`  through rspec tests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,7 +45,7 @@ gem "faraday_middleware"
 
 gem "fastimage"
 
-gem "dfe-analytics", github: "DFE-Digital/dfe-analytics", tag: "3a20e6038e18439834c9f94c78bb24046f985492"
+gem "dfe-analytics", github: "DFE-Digital/dfe-analytics", tag: "v1.8.0"
 
 gem "hashids"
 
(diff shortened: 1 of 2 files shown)


asatwal:
> The only thing I noted was the proc will evaluate to the result of `attempt` when the page is cached which is an array rather than boolean. Is it worth casting to a boolean here or are we happy the calling code will treat that as `true`?

Thanks for picking this up @ethax-ross. I think it's best to return a boolean here, so I've updated.

ethax-ross (approved):

PR #3219, opened 2023-04-03
Enable Google Optimize on variant test pages

ethax-ross:
### Context

Currently we only fire Google Optimize on the original page and not the variant (as the traffic is all sent to the original page and redirected to the variant from there).

We're seeing oddness around far more traffic going to the original than the variant (although GA shows more of a split) so we're assuming the Optimize snippet also needs to execute on the variant page in order for it to register correctly.

- Enable Google Optimize on variant test pages


gemmadallmandfe (approved):
Looks good to me

PR #3222, opened 2023-04-04
Remove blur from Optimize experiment

ethax-ross:
### Trello card

[Trello-4434](https://trello.com/c/WRldjmDA/4434-remove-blur-effect-from-optimize-experiments)

### Context

Currently, when a user visits a page running an A/B test if they have not accepted cookies the content behind the cookie dialog is blurred. This is to ensure they can't see the original page if they then get redirected to the variant page on accepting cookies/opting-in to the experiment.

We think this isn't a great user experience and - given most A/B tests are the same or similar for above the fold content - we would rather show the original briefly and switch out to the variant without the blurring effect.

- Remove blur from Optimize experiment

### Guidance to review

Visit a page running optimize [like this one](https://review-get-into-teaching-app-3222.london.cloudapps.digital/steps-to-become-a-teacher-search) without accepting cookies and check the page isn't blurred.

### DIFF
--- a/app/webpacker/javascript/google_optimize.js
+++ b/app/webpacker/javascript/google_optimize.js
@@ -15,18 +15,9 @@ export default class GoogleOptimize {
 
     this.listenForTurbolinksBeforeVisit();
     this.listenForConsentChange();
-
-    if (!this.seenCookieDialog && this.isExperimentPath()) {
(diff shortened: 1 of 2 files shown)


gemmadallmandfe (approved):
Looks good to me

PR #3224, opened 2023-04-04
Enable Google Optimize on home/landing

ethax-ross:
We are running an A/B test on the home page with a home landing page; this enables Optimize on the
respective paths.


gemmadallmandfe (approved):
Looks good to me

PR #3226, opened 2023-04-05
Automatically exclude pages from the cache

ethax-ross:
### Trello card

[Trello-4436](https://trello.com/c/OAqeDuJi/4436-improve-page-cache-exclusion)

### Context

Currently we maintain a manual array of page paths that we want to exclude from the static page cache. These tend to be pages that contain a form with a post request method (if we cache these pages the CSRF token is also cached and becomes invalid, causing the form submission to fail). Going forward we want to automate this process as its currently prone to human error.

- Automatically exclude pages from the cache

Add middleware to check if a response contains a form and - if it does - remove the cached version of the page that is created earlier up the middleware stack.

### Guidance to review

Visit a static page with a POST form, for example [this one](https://review-get-into-teaching-app-3226.london.cloudapps.digital/landing/how-much-do-teachers-get-paid) and check if submits correctly. It should also return `cache-control` headers that don't allow caching. If you visit a page that does allow caching, like [this one](https://review-get-into-teaching-app-3226.london.cloudapps.digital/) it should return with `cache-control` headers using a 5m TTL.

### DIFF
--- a/lib/middleware/page_cache_exclusion.rb
+++ b/lib/middleware/page_cache_exclusion.rb
@@ -1,0 +1,27 @@
+module Middleware
+  class PageCacheExclusion
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      status, headers, response = @app.call(env)
+
+      if dynamic_page?(body(response))
+        Rack::PageCaching::Cache.delete("#{env['PATH_INFO']}.html")
+      end
+
+      [status, headers, response]
+    end
+
+  private
+
+    def body(response)
+      response.is_a?(ActionDispatch::Response::RackBody) ? response.body : response.first
+    end
+
+    def dynamic_page?(body)
+      body&.match(/method="post"/i)
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


lib/middleware/page_cache_exclusion.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+  private
+
+    def body(response)
+      response.is_a?(ActionDispatch::Response::RackBody) ? response.body : response.first
+    end
+
+    def dynamic_page?(body)
+      body&.match(/method="post"/i)
```
asatwal, line 24: Could the method possibly be a `patch` or `put` also ?

asatwal (dismissed):
Looks good. Just 1 comment

lib/middleware/page_cache_exclusion.rb
ethax-ross: Ah yes good point it could, I'll update it

asatwal (approved):
👍

PR #3235, opened 2023-04-17
Add legend to event format filters

ethax-ross:
### Trello card

[Trello-4468](https://trello.com/c/GEFL0EWH/4468-add-fieldset-legend-to-event-format-section-on-events-listing-page)

### Context

Silktide picked up that we don't have a legend for the event
format checkboxes; we should add one for improved accessibility of these fields.

- Add legend to event format filters


gemmadallmandfe (approved):
Looks good to me - tested on iPhone 11 using VoiceOver

PR #3240, opened 2023-04-19
Update home page with variant

ethax-ross:
### Trello card

[Trello-4450](https://trello.com/c/xyeVnPld/4450-unwind-home-page-service-proposition-test)

### Context

We want to unwind the home page A/B test and update the home page to match the variant (which performed better in the test).

- Update home page with variant


gemmadallmandfe:
Looks good to me @ethax-ross - design elements are all there and I've tested the redirect and the mailing list embed

I've just noticed that the **salary banner** was deleted - I've checked with @philspray and @EllieNodder and we think it should still be there below the events/adviser blocks and above the directory (none of us noticed during the split test :-) )

Is it easy to add the salaries banner back in on this PR? It's not critical in terms of getting this change out - can always be done retrospectively if easier

image


image

gemmadallmandfe (approved):
Looks good to me - thanks for re-introducing the salaries bar @ethax-ross !

@gazcarr has also reviewed the page and is happy with it

Happy for this to be merged

PR #3244, opened 2023-04-20
Fix cache deletion for home page

ethax-ross:
### Trello card

[Trello-4485](https://trello.com/c/rxPZXfap/4485-fix-homepage-cache-exclusion-issue)

### Context

We auto-delete the page cache if a form is detected. When doing this we use the request path to find the cached filename. The homepage is unique because the path is not the same as the cached page filename (the path is `/` and the filename is `/index.html`). We need to account for this explicitly to ensure the cached file is removed when the homepage contains a form.

- Fix cache deletion for home page

### Guidance to review

I've added a [Trello ticket](https://trello.com/c/nV6cjqlo/4486-improve-page-path-matching-logic-for-page-cache-exclusion-middleware) to see if we can improve this, but we need to get this fix out asap.

### DIFF
--- a/lib/middleware/page_cache_exclusion.rb
+++ b/lib/middleware/page_cache_exclusion.rb
@@ -8,13 +8,19 @@ module Middleware
       status, headers, response = @app.call(env)
 
       if dynamic_page?(body(response))
-        Rack::PageCaching::Cache.delete("#{env['PATH_INFO']}.html")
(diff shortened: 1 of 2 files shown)


gemmadallmandfe (approved):
Happy for this to be merged

asatwal (approved):
👍

PR #3247, opened 2023-04-21
fixing full stops

philspray:
### Context

A small change to fix some full stops in the wrong place

Moved two full stops outside of links


gemmadallmandfe (approved):
Looks good to me

PR #3248, opened 2023-04-21
Add common build errors to content doc

ethax-ross:
### Trello card

[Trello-4490](https://trello.com/c/mYfxCHuY/4490-add-common-build-errors-to-the-content-docs)

### Context

We want to track the common build errors in the content document so content editors can more easily resolve these issues.

- Add common build errors to content doc


philspray (approved):
Makes sense to me

PR #3252, opened 2023-04-24
Remove noindex from homepage

ethax-ross:
This was accidentally copied form the landing page variant we went ahead with for our home page after the test completed.


gemmadallmandfe (approved):
Looks good to me

PR #3274, opened 2023-05-03
Ensure content filename consistency

ethax-ross:
### Trello card

[Trello-4494](https://trello.com/c/LBL4YK3n/4494-add-test-to-prevent-capital-letters-in-path-names)

### Context

We sometimes have markdown files added with incorrect casing or using underscores instead of hyphens.

Add a test to ensure all content filenames are lower case letters, numbers and underscores.

- Ensure content filename consistency


ethax-ross:
@gemmadallmandfe I only needed to add the redirect for that one file as it got in before the test was added; it won't be possible to add them in the first place once the test is in. This also only tests the static markdown pages so won't apply to the adviser sign up (or any other dynamic page).

PR #3275, opened 2023-05-03
Update guidance for markdown filenames

ethax-ross:
### Trello card

[Trello-4495](https://trello.com/c/QqxxaIjM/4495-update-content-docs-with-additional-build-errors-guidance)

### Context

We occasionally have files added that contain an upper case character or other symbol. This can cause issues as URLs are case sensitive and we should be consistent throughout the website.

We recently added a test to enforce filename consistency; this adds relevant guidance to the content documentation.

- Update guidance for markdown filenames

### DIFF
--- a/docs/content.md
+++ b/docs/content.md
@@ -410,5 +410,20 @@ This error means there is a hyperlink in your content that is linking to a headi
 
 #### Resolution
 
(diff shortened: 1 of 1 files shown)


philspray (approved):
Makes sense to me

PR #3282, opened 2023-05-09
Improve page cache exclusion logic

ethax-ross:
### Trello card

[Trello-4486](https://trello.com/c/nV6cjqlo/4486-improve-page-path-matching-logic-for-page-cache-exclusion-middleware)

### Context

We had an issue in production whereby the home page was being cached after we added a form to it, resulting in the CSRF token becoming invalid.

To avoid this we added middleware to prevent caching of any pages containing a form, but this contained a workaround for the root path, which is a special case whereby the path does not match the cached filename (`/` resolves to `/index`).

We added a quick fix for this but it wasn't very robust; instead, this commit queries the page caching gem to obtain the cached page filename for the given request path.

- Improve page cache exclusion logic

### DIFF
--- a/lib/middleware/page_cache_exclusion.rb
+++ b/lib/middleware/page_cache_exclusion.rb
@@ -5,29 +5,32 @@ module Middleware
     end
 
     def call(env)
-      status, headers, response = @app.call(env)
+      rack_response = @app.call(env)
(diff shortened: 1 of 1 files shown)


asatwal (approved):
👍

PR #3292, opened 2023-05-12
Fix Turbolinks page change/scroll issue

ethax-ross:
### Trello card

[Trello-4566](https://trello.com/c/VqOAqE3d/4566-fix-issue-with-some-links-opening-halfway-down-page)

### Context

On some pages if you scroll down and click a link you will be taken to the page and the scroll position will be maintained (instead of scrolling back to the top of the new page). This bug was introduced when we migrated from Webpacker to Shakapacker.

The cause appears to be Turbolinks reloading the
`application.js` asset on these page changes. It only happens when the destination page uses a different layout containing different Javascript assets; in our case going from a normal content page to a page containing a form is the main culprit as we pull in an additional
`govuk_frontend` pack on these pages. Its not clear why Turbolinks then reloads the `application` pack as well.

We likely want to migrate away from Turbolinks to Turbo in the future, but we've tried that in the past and it causes a separate set of headaches with our analytics. A fix for now is to ensure all pages load the same set of Javascript packs.

- Fix Turbolinks page change/scroll issue

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,7 +17,6 @@ class ApplicationController < ActionController::Base
   before_action :record_utm_codes
   before_action :add_home_breadcrumb
   before_action :declare_frontmatter
-  before_action :declare_js_packs
 
 protected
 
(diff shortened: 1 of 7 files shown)


gemmadallmandfe (approved):
Looks good to me @ethax-ross - I have tested on desktop and on my iPhone, going between pages with different layouts (including the mailing list start page). All links tested open up pages at the top and do not scroll to the previous scroll position

PR #3301, opened 2023-05-17
Adding indented text for internationals to parents/carers and disabled page

EllieNodder:
### Trello card

https://trello.com/c/Zp3EFYvI/4440-add-international-inset-text-to-if-youre-a-parent-or-carer-if-youre-disabled-and-if-youre-a-veteran-pages

### Context

Most international users won't be eligible for the support detailed on the 'if you're a parent or carer' and 'if you're disabled' pages.

We need to highlight this and effectively signpost them on to other relevant support.

### DIFF
--- a/app/components/funding_widget_component.html.erb
+++ b/app/components/funding_widget_component.html.erb
@@ -30,7 +30,7 @@
             <h3><%= sub_head %></h3>
             <div>
               <p>
-                A scholarship or bursary isn't available for your chosen subject, but you can still get a <a href="/funding-and-support/tuition-fee-and-maintenance-loans">tuition fee and maintenance loan</a> if you're eligible (non-UK citizens without settled status are unlikely to be eligible).
+                A scholarship or bursary isn't available for your chosen subject, but you can still get a <a href="/funding-and-support/tuition-fee-and-maintenance-loans">tuition fee and maintenance loan</a> if you're eligible (non-UK citizens without <a href="https://www.gov.uk/guidance/indefinite-leave-to-remain-in-the-uk">indefinite leave to remain in the UK</a> are unlikely to be eligible).
               </p>
 
               <p>
--- a/app/components/funding_widget_component.html.erb
+++ b/app/components/funding_widget_component.html.erb
@@ -30,7 +30,7 @@
             <h3><%= sub_head %></h3>
             <div>
               <p>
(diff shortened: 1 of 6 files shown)


EllieNodder:
Have discussed the wording with Laura - she's just double checking with Jane to approve. Laura has confirmed we don't need this indented text for the veterans page as the info there can apply to them @philspray @gemmadallmandfe

gemmadallmandfe:
@EllieNodder My concern with this is around the positioning - being right at the top of the page. Could this be a bit offputting for people who are actually eligible, who may not scroll down? It is a big block of text on mobile in particular

Is there any reason why it couldn't go beneath the intro paras, above the h2 (as we do on steps)? We don't have feedback (as far as I'm aware) that lots of international candidates are applying for this funding

cc @philspray

philspray (commented):
Just a suggestion on the wording and an extra full stop. Otherwise looks good to me.

gemmadallmandfe (dismissed):
Looks good to me - happy for this to be merged first thing tomorrow

EllieNodder:
Have tweaked after feedback from Jayne @gemmadallmandfe @philspray

philspray (approved):
Looks good to me.

app/components/funding_widget_component.html.erb
```diff
@@ -30,7 +30,7 @@
             <h3><%= sub_head %></h3>
             <div>
               <p>
-                A scholarship or bursary isn't available for your chosen subject, but you can still get a <a href="/funding-and-support/tuition-fee-and-maintenance-loans">tuition fee and maintenance loan</a> if you're eligible (non-UK citizens without settled status are unlikely to be eligible).
+                A scholarship or bursary isn't available for your chosen subject, but you can still get a <a href="/funding-and-support/tuition-fee-and-maintenance-loans">tuition fee and maintenance loan</a> if you're eligible (non-UK citizens without <a href="https://www.gov.uk/guidance/indefinite-leave-to-remain-in-the-uk">indefinite leave to remain in the UK</a> are unlikely to be eligible).
```
gemmadallmandfe, line 33: @EllieNodder Why is there a link in this default widget option but not in the other subject-specific ones?

Unless you feel it is critical here, I would suggest removing the link as a) there are already lots of links in the widget text so it is distracting and b) we've already linked to this in the intro section above the widget so we are essentially repeating the link twice with the same link text which is not great in terms of accessibility best practice

gemmadallmandfe (approved):
Looks good to me

PR #3304, opened 2023-05-19
changing info box link

philspray:
### Context

International team have been in contact to ask that the link is changed to point to the fees and funding page for trainees

Changed the link in the info bar to point to the fees and funding for trainees page


gemmadallmandfe (approved):
Looks good to me

PR #3309, opened 2023-05-22
Consolidate TeachingSubject presentation

ethax-ross:
### Trello card

[Trello-4591](https://trello.com/c/UQdgvC0I/4591-consolidate-and-streamline-teaching-subjects-on-git-funnels)

### Context

We currently retrieve teaching subjects from the API in a multitude of different ways. We want to be consistent in how we retrieve subjects and also in how we filter/display them.

- Consolidate TeachingSubject presentation

Centralise the teaching subject queries to the `Crm::TeachingSubject` model and update all references to go through this.

### DIFF
--- a/app/models/crm/teaching_subject.rb
+++ b/app/models/crm/teaching_subject.rb
@@ -51,7 +51,7 @@ class Crm::TeachingSubject
     end
 
     def all
-      GetIntoTeachingApiClient::LookupItemsApi.new.get_teaching_subjects.reject { |subject| IGNORED.value?(subject.id) }.each { |subject| if subject && RENAMED[subject.value].present? then subject.value = RENAMED[subject.value]; end; }
+      GetIntoTeachingApiClient::LookupItemsApi.new.get_teaching_subjects.reject { |subject| IGNORED.value?(subject.id) }.each { |subject| if subject && RENAMED[subject.value].present? then subject.value = RENAMED[subject.value]; end; }.sort_by { |f| [f.value] }
     end
   end
 end
--- a/app/models/crm/teaching_subject.rb
+++ b/app/models/crm/teaching_subject.rb
@@ -1,47 +1,19 @@
 class Crm::TeachingSubject
-  ALL =
-    {
-      "Art" => "6b793433-cd1f-e911-a979-000d3a20838a",
-      "Art and design" => "7e2655a1-2afa-e811-a981-000d3a276620",
-      "Biology" => "802655a1-2afa-e811-a981-000d3a276620",
-      "Business studies" => "822655a1-2afa-e811-a981-000d3a276620",
-      "Chemistry" => "842655a1-2afa-e811-a981-000d3a276620",
-      "Citizenship" => "862655a1-2afa-e811-a981-000d3a276620",
-      "Classics" => "882655a1-2afa-e811-a981-000d3a276620",
-      "Computing" => "8a2655a1-2afa-e811-a981-000d3a276620",
-      "Dance" => "8c2655a1-2afa-e811-a981-000d3a276620",
-      "Design and technology" => "8e2655a1-2afa-e811-a981-000d3a276620",
-      "Drama" => "902655a1-2afa-e811-a981-000d3a276620",
-      "Economics" => "922655a1-2afa-e811-a981-000d3a276620",
-      "English" => "942655a1-2afa-e811-a981-000d3a276620",
-      "French" => "962655a1-2afa-e811-a981-000d3a276620",
-      "General science" => "982655a1-2afa-e811-a981-000d3a276620",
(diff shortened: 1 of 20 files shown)


spec/helpers/welcome_helper_spec.rb
```diff
@@ -89,7 +89,7 @@
 
     context "when stored in the mailinglist session store" do
       context "when the subject name is a common noun" do
-        %i[art_and_design biology chemistry general_science languages_other maths physics_with_maths physics].each do |subject_key|
```
ethax-ross, line 92: Why are we removing these?

app/models/crm/teaching_subject.rb
```diff
@@ -51,27 +23,35 @@ def lookup_by_keys(*keys)
...
+    def all_hash
+      all.to_h { |item| [item.value, item.id] }
+    end
+
+    def all_without_primary
+      all_hash.reject { |_h, s| s == PRIMARY }
+    end
+
+    def all
+      GetIntoTeachingApiClient::LookupItemsApi.new.get_teaching_subjects.reject { |subject| IGNORED.value?(subject.id) }.each { |subject| if subject && RENAMED[subject.value].present? then subject.value = RENAMED[subject.value]; end; }.sort_by { |f| [f.value] }
```
ethax-ross, line 54: I think this needs to be simplified its quite hard to follow. We could condense it to something along the lines of:

```
def all
  GetIntoTeachingApiClient::LookupItemsApi.new.get_teaching_subjects
    .sort_by(&:value)
    .reject(&method(:ignored?))
    .each(&method(:rename))
end

private

def rename(subject)
  subject.value = RENAMED[subject.value] || subject.value
end

def ignored?(subject)
  IGNORED.value?(subject.id)
end
```

As a follow-up PR I think we should look at removing the `all_hash` from this class, but that would have many knock-on changes so it can be done separately

ethax-ross (commented):
Looking good, just a couple of questions and a suggestion

spec/features/event_wizard_spec.rb
```diff
@@ -336,7 +336,7 @@ def personalised_attributes
     {
       degree_status_id: 222_750_000,
       consideration_journey_stage_id: 222_750_000,
-      preferred_teaching_subject_id: Crm::TeachingSubject.lookup_by_key(:art),
+      preferred_teaching_subject_id: Crm::TeachingSubject.lookup_by_key(:art_and_design),
```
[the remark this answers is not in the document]
plisovin: ok I will check, I think I've made wrong assumption that this is just an example subject used in this test and as the subject got removed then a next one could be picked up from the list. I will investigate it further

spec/helpers/welcome_helper_spec.rb
plisovin: Same reason as above.

app/models/crm/teaching_subject.rb
plisovin: yeah, good idea, I just didn't look through as just fixed it so rubocop doesn't complain

plisovin (approved):

gemmadallmandfe:
Looks good @plisovin - just a couple of minor things:

* Can you please add `General science` back in again but rename it as `Science` (my fault - sorry!)

* Can the subjects be displayed alphabetically? `Communication and media studies` is appearing where `Media studies` previously did but this is no longer in alphabetical order
NB this will also be the case when `General science` is renamed as `Science`

image

* I'm presuming that only the adviser funnel on GIT will be updated with the modified subject list, and not the original adviser funnel? I don't think that is an issue particularly - just want to be clear


I will need to let the adviser team know of the changes so please don't merge this in the meantime :-)

spec/helpers/welcome_helper_spec.rb
```diff
@@ -89,7 +89,7 @@
...
+        %i[art_and_design biology chemistry languages_other maths physics].each do |subject_key|
```
gemmadallmandfe, line 92: @plisovin Do we need to add `science` in here now that we've added it back in as a subject?

gemmadallmandfe (approved):
Looks good to me - happy for this to be merged tomorrow @plisovin

One thing I've noticed is that a couple of questions in the adviser funnel have an additional option of 'Other'

'/teacher-training-adviser/sign_up/subject_taught'
Which main subject did you previously teach?

`/teacher-training-adviser/sign_up/subject_like_to_teach`
Which subject would you like to teach if you return to teaching?

'/teacher-training-adviser/sign_up/what_subject_degree`
What subject is your degree?

We may need to introduce this 'Other' option back in again for these specific questions - however, I will raise a separate ticket for this rather than hold up this PR

We then have work planned in later this summer to address degree subject properly!

PR #3312, opened 2023-05-23
Update focus state for accessibility

ethax-ross:
Fixes and re-introduces https://github.com/DFE-Digital/get-into-teaching-app/pull/3308


gemmadallmandfe (approved):
Looks good to me

PR #3320, opened 2023-05-30
Changes ahead of TTA sign up rollout

ethax-ross:
### Trello card

[Trello-4582](https://trello.com/c/bNfnapLz/4582-launch-and-run-adviser-funnel-split-test)

### Context

Ahead of rolling out the TTA sign up we want to allow access to the form but only if you have the URL. We don't want to allow Google to index it yet or redirect traffic from the website to the new form until the feature flag is turned on.

- Allow TTA sign up access in production

- Enable Optimize on TTA sign up start page

We are going to use Optimize to redirect a portion of traffic to the existing Get an adviser service.

Enable Optimize on the start page of the new TTA sign up form.

- Disable TTA feedback form in production

When we enable the TTA sign up form we don't want to enable the feedback form; we will be re-working this into a generic GiT feedback form first.


plisovin (dismissed):
LGTM

PR #3324, opened 2023-05-31
Updating adviser start page for internationals

EllieNodder:
### Trello card

https://trello.com/c/LXOcE9Rx/4627-clarify-guidance-for-international-trainees-and-teachers-on-git-adviser-funnel-start-page

### Context

In moving the adviser funnel to GIT, we need to make sure that we are still being clear that international teachers need QTS to be eligible for an adviser.

We are currently implying that both international trainees and teachers should ring the contact centre to check their qualifications. We ideally want trainees to ring about their degree, and want to direct teachers towards the non uk teachers page, so we should tweak the order of the content on the start page to make this clearer.

### DIFF
--- a/app/views/teacher_training_adviser/steps/_identity.html.erb
+++ b/app/views/teacher_training_adviser/steps/_identity.html.erb
@@ -14,12 +14,12 @@
 </p>
 
 <p>
-  If you're already a qualified teacher, you'll need English qualified teacher status (QTS) and want to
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
Looks good @EllieNodder 

My only (very minor) suggestion would be to remove 'more' from the inset component - stops it wrapping onto another line on (some) mobiles so you get to the form a bit quicker?

image


but happy to tick now if you disagree :-)

PR #3352, opened 2023-06-14
Updating feedback form link in the footer

EllieNodder:
### Trello card

https://trello.com/c/sVwbuzMx/4668-update-feedback-form-link-in-the-footer

### Context

We changed our feedback form from Google to MS as part of the G-suite migration, but the feedback form link in the footer still points to the old Google form.

We need to update this so it now points to the MS form.


gemmadallmandfe (approved):
Looks good to me

PR #3357, opened 2023-06-20
How to become a teacher tweaks

EllieNodder:
### Context

Kate from Find has reviewed the new How to become a teacher content and suggested a small tweak to some link text.


gemmadallmandfe (approved):
Looks good to me

PR #3366, opened 2023-06-28
Set version of shakapacker to match its version in package.json

plisovin:
This should fix issue running the app in codespaces


asatwal (approved):
👍 LGTM

PR #3385, opened 2023-07-04
Add a template page for dev design components

plisovin:


gemmadallmandfe:
@garethwilliamsdesign Are you happy for us to close this PR now? I think this was created when we were looking at whether we could have a page with the design components on it; that may still be a good idea but something that we are not actively working on at the moment?

cc @GayeCord

garethwilliamsdesign:
Yes, I am happy for this PR to be closed.

PR #3388, opened 2023-07-05
Update how to attend content

philspray:
### Trello card

https://trello.com/c/x8rsMyqS

### Context

Revisiting the 'How to attend' content on events to try to encourage further registrations.

Adding a benefit into the 'How to attend' content.

### DIFF
--- a/app/views/teaching_events/show/_how-to-attend.html.erb
+++ b/app/views/teaching_events/show/_how-to-attend.html.erb
@@ -13,13 +13,15 @@
 
       <% if @event.in_person? %>
       <p>
-        Register for this event and we can offer further tailored support and advice on your journey into teaching.
+        Register for this event so you can check-in faster on arrival using a QR code.
+        
+        When you register, you can also sign-up for free email reminders about events near you.
       </p>
       <% end %>
 
       <% if @event.online? %>
       <p>
-          Once registered, you will receive log-in information and joining
+          Register for this event and you will receive log-in information and joining
           instructions via email.
       </p>
 
--- a/app/views/teaching_events/show/_how-to-attend.html.erb
+++ b/app/views/teaching_events/show/_how-to-attend.html.erb
@@ -13,19 +13,18 @@
 
       <% if @event.in_person? %>
       <p>
-        Register for this event so you can check-in faster on arrival using a QR code, and sign-up for free email reminders about events near you.
+        Register for this event to secure your place and check-in faster on arrival using a QR code.
       </p>
       <% end %>
 
       <% if @event.online? %>
       <p>
-          Register for this event and you will receive log-in information and joining
+          Register for this event and you'll receive log-in information and joining
           instructions via email.
       </p>
 
       <p>
-          To access this event you will need a laptop or desktop PC
-          and be using Google Chrome as your browser.
+          For the best experience when accessing this event, we reccomend using Google Chrome on a laptop or desktop PC.
       </p>
       <% end %>
 
--- a/app/views/teaching_events/show/_how-to-attend.html.erb
+++ b/app/views/teaching_events/show/_how-to-attend.html.erb
@@ -14,7 +14,7 @@
       <% if @event.in_person? %>
       <p>
         Register for this event so you can check-in faster on arrival using a QR code.
-        
+      
         When you register, you can also sign-up for free email reminders about events near you.
       </p>
       <% end %>
--- a/app/views/teaching_events/show/_how-to-attend.html.erb
+++ b/app/views/teaching_events/show/_how-to-attend.html.erb
@@ -11,7 +11,7 @@
 
     <% if can_sign_up_online?(@event) %>
       <p>
-        Register to attend this event and you can also quickly and easily to sign-up for tailored email guidance to help you get into teaching.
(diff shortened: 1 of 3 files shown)


app/views/teaching_events/show/_how-to-attend.html.erb
```diff
@@ -11,13 +11,16 @@
 
     <% if can_sign_up_online?(@event) %>
       <p>
-        You must register for a place to attend this event.
-
+        Register for this event and we can offer further tailored support and advice on your journey into teaching.
```
EllieNodder, line 14: I'm wondering if we might want to be more specific here - to me, this kinda sounds like if you sign up for an event, you're essentially joining our mailing list?

Instead, for in-person events, could we say something like 'Register for this event and you'll receive a QR code to use for a quicker check-in process when you arrive, as well as reminder emails about events near you' or something along those lines?

(Sorry to complicate things!!)

philspray: Thanks, I see your point and I did consider this but I don't think it's currently structured to show different content for in-person/online events. Is that possible @plisovin?

app/views/teaching_events/show/_how-to-attend.html.erb
```diff
@@ -10,17 +10,24 @@
     <h2>How to attend</h2>
 
     <% if can_sign_up_online?(@event) %>
+
+      <% if @event.in_person? %>
       <p>
-        You must register for a place to attend this event.
+        Register for this event so you can check-in faster on arrival using a QR code, and sign-up for free email reminders about events near you.
```
EllieNodder, line 16:
```suggestion
        Register for this event and we'll send you a QR code for a faster check-in when you arrive and free email reminders about events near you.
```
(sorry to be a pain) I'm just wondering if by saying 'sign up for free email reminders', it kind of sounds like we're implying that that's a separate process? i.e. register for this event, and then sign up for emails? (when I'm assuming what we mean is that if they sign up for the event, we'll automatically send them email reminders? they don't need to sign up again to something else?)

feel free to ignore if you disagree or if I've got this totally wrong!

app/views/teaching_events/show/_how-to-attend.html.erb
```diff
@@ -10,17 +10,24 @@
...
+      <% if @event.online? %>
+      <p>
+          Register for this event and you will receive log-in information and joining
           instructions via email.
+      </p>
 
-          Please note: to access this event you will require a laptop/desktop PC
+      <p>
+          To access this event you will need a laptop or desktop PC
           and be using Google Chrome as your browser.
```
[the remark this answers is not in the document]
gemmadallmandfe: Pete Y and I spoke to the Events team previously about this (will send info)  - I think we should say something like "For the best experience, we recommend using Google Chrome on a laptop or desktop PC"

app/views/teaching_events/show/_how-to-attend.html.erb
gemmadallmandfe: Should we say something like 'Register for this event **to secure your place**.....' - events do sometimes close early when full?

Agree with Ellie - not sure we need to talk about email reminders here - they may well get them already anyway through other sign-ups

app/views/teaching_events/show/_how-to-attend.html.erb
```diff
@@ -10,17 +10,23 @@
...
-        You must register for a place to attend this event.
+        Register for this event to secure your place and check-in faster on arrival using a QR code.
+      </p>
+      <% end %>
 
-        <% if @event.online? %>
-          Once registered, you will receive log-in information and joining
+      <% if @event.online? %>
+      <p>
+          Register for this event and you'll receive log-in information and joining
```
gemmadallmandfe, line 22:
```suggestion
          Register for this event and you'll receive and joining
```
think log-in information and joining instructions are the same thing?

app/views/teaching_events/show/_how-to-attend.html.erb
```diff
@@ -10,17 +10,23 @@
     <h2>How to attend</h2>
 
     <% if can_sign_up_online?(@event) %>
+
+      <% if @event.in_person? %>
       <p>
-        You must register for a place to attend this event.
+        Register for this event to secure your place and check-in faster on arrival using a QR code.
```
gemmadallmandfe, line 16:
```suggestion
        Register for this event to secure your place and check in faster on arrival using a QR code.
```
Hyphen not needed for a verb?

spec/features/teaching_events/viewing_spec.rb
```diff
@@ -103,15 +103,14 @@
...
-      it { is_expected.to have_content("You must register for a place to attend this event") }
       it { is_expected.to have_link(register_link_text) }
     end
 
     context "when the event is online" do
       let(:event) { build(:event_api, :online) }
 
       let(:expected) do
-        "Once registered, you will receive log-in information and joining instructions via email."
+        "Register for this event and you'll receive log-in information and joining instructions via email."
```
gemmadallmandfe, line 113:
```suggestion
        "Register for this event and you'll receive joining instructions via email."
```

gemmadallmandfe (approved):
Looks good to me

PR #3397, opened 2023-07-11
Adding context to salaries pages re maintained schools

EllieNodder:
### Trello card

https://trello.com/c/mI6qifQQ/4773-add-content-around-maintained-schools-on-salaries-pages

### Context

We don't highlight on our salaries pages that the figures apply to maintained schools - schools such as academies can set their own pay scales. We should make this clear to users.

### DIFF
--- a/app/views/content/landing/how-much-do-teachers-get-paid/_collage.html.erb
+++ b/app/views/content/landing/how-much-do-teachers-get-paid/_collage.html.erb
@@ -9,7 +9,7 @@
 
         <%= render Content::ChecklistCollageComponent.new(
           checklist: [
-            "You'll earn at least £28k (or more in London) if you're just starting as a teacher in a maintained primary, secondary or special school in England (schools funded by local authorities).",
+            "You'll earn at least £28k (or more in London) if you're just starting as a teacher.",
             "The average salary for a classroom teacher in England is £39,500 a year."
           ],
           image_paths: [
--- a/app/views/content/landing/how-much-do-teachers-get-paid/_collage.html.erb
+++ b/app/views/content/landing/how-much-do-teachers-get-paid/_collage.html.erb
@@ -9,7 +9,7 @@
 
         <%= render Content::ChecklistCollageComponent.new(
           checklist: [
-            "You will earn at least £28k (or more in London) if you're just starting as a teacher.",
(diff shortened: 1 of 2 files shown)


app/views/content/salaries-and-benefits.md
```diff
@@ -52,11 +52,13 @@ keywords:
 
 For helping to shape the next generation, you’re entitled to a competitive salary, generous holidays, and a substantial pension.
 
-Figures are for the 2022/23 academic year.
+Figures are for the 2022/23 academic year and apply to maintained primary, secondary and special schools in England (schools funded by local authorities).
```
gemmadallmandfe, line 55: @EllieNodder Is it worth bolding `Figures are for the 2022/23 academic year` so it doesn't get lost? But if this is a content no-no, happy as is!

app/views/content/landing/how-much-do-teachers-get-paid/_collage.html.erb
```diff
@@ -9,7 +9,7 @@
 
         <%= render Content::ChecklistCollageComponent.new(
           checklist: [
-            "You will earn at least £28k (or more in London) if you're just starting as a teacher.",
+            "You'll earn at least £28k (or more in London) if you're just starting as a teacher in a maintained primary, secondary or special school in England (schools funded by local authorities).",
```
gemmadallmandfe, line 12: @EllieNodder I wonder if this is a bit overkill for the landing page? The advertising team don't include this caveat in their ads

We link through to the salaries and benefits page which has this detail so we are not hiding it

EllieNodder: worth leaving the landing page as is then?

app/views/content/salaries-and-benefits.md
EllieNodder: hmm we don't really use bold to emphasise text in regular content pages (and the gov.uk style guide says to use it sparingly) so would be inclined not to do so here

philspray: Is it worth splitting into 2 sentences? 

```suggestion
Figures are for the 2022/23 academic year. 

These salaries apply at maintained primary, secondary and special schools in England (schools funded by local authorities).
```

philspray (commented):
Just one suggestion, otherwise looks good to me.

philspray (approved):
Looks good to me

PR #3399, opened 2023-07-11
updating no events message

philspray:
### Trello card

https://trello.com/c/IxPbkiS5

### Context

An automated message is displayed when there are no Get Into Teaching events because the programme is closed. This message currently refers to the autumn programme so needs to be updated.

Amended the message to make it generic to avoid the need for a manual update in future.

### DIFF
--- a/app/views/teaching_events/about_git_events.html.erb
+++ b/app/views/teaching_events/about_git_events.html.erb
@@ -61,7 +61,7 @@
         </section>
       <% else %>
         <section class="col col-720 col-space-m">
-          <p><strong>Our Get Into Teaching events are now closed. Details of our next events will be published on this page soon.</strong></p>
+          <p><strong>We do not have any Get Into Teaching events open for registration right now. We'll update this page with new Get Into Teaching events soon.</strong></p>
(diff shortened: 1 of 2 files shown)


app/views/teaching_events/about_git_events.html.erb
```diff
@@ -61,7 +61,7 @@
         </section>
       <% else %>
         <section class="col col-720 col-space-m">
-          <p><strong>Our autumn Get Into Teaching events are now closed. Details of our spring events will be published on this page in early January 2023.</strong></p>
+          <p><strong>Our Get Into Teaching events are now closed. Details of our next events will be published on this page soon.</strong></p>
```
EllieNodder, line 64:
```suggestion
          <p><strong>We do not have any Get Into Teaching events open for registration right now. We'll update this page with new Get Into Teaching events soon.</strong></p>
```
I'm wondering if the first sentence almost implies that they've stopped completely? would it be better to go with something a bit like this?

gemmadallmandfe (approved):
Looks good to me

PR #3412, opened 2023-07-13
Updating salaries in line with the new pay award

EllieNodder:
### Trello card

https://trello.com/c/DP4nUHdx/4779-update-salaries-information-for-new-pay-award

### Context

We need to update the salary information around our site in line with the new pay award (which is subject to a statutory consultation process and cannot be confirmed until the final School Teachers’ Pay and Conditions Document is published.)


gemmadallmandfe (approved):
Looks good to me

PR #3422, opened 2023-07-18
Remove providers groups as it's been removed from the page

plisovin:
### Trello card

[Trello-1549](https://trello.com/c/9mYNj2T9/1549-delete-teaching-internships-providers-file)

### Context

We are getting 404 errors to do with internship providers (see screenshot) even though they are no longer listed on the front end: https://getintoteaching.education.gov.uk/teaching-internship-providers

The providers list probably needs removing from the front matter on the internships page (the page itself should not be deleted).


gemmadallmandfe (approved):
Looks good to me

PR #3425, opened 2023-07-18
Fix vwo as it has to be exact code even with linting errors

plisovin:
### Trello card

[Trello-4484](https://trello.com/c/h7ZMKKQd/4484-set-up-vwo-account)

### Context

Need the exact code snippet to be able to work properly


asatwal (approved):
👍  Looks OK to me.

PR #3428, opened 2023-07-19
improve adviser completion page for international callbacks

philspray:
### Trello card

https://trello.com/c/B29cZRt3

### Context

We need to improve the content of the adviser sign up completion page to make it clearer that international candidates need to wait for a call back. 

- moving header into conditional sections


EllieNodder (approved):
looks good to me!

gemmadallmandfe (approved):
Looks good to me

PR #3429, opened 2023-07-20
Re-order international pages

philspray:
### Trello card

https://trello.com/c/RlVuD63s

### Context

Two of the category card pages had the same navigation setting. We need to update them to the correct order on the page.

- updated navigation settings


gemmadallmandfe:
Not sure this solves the problem @philspray

eg /teach-in-england-if-you-trained-overseas has a nav order of 20.2 so we are now just duplicating that?

I think we need to look at all the international pages and make sure there are no duplicates across all of them?

(Also - don't see why we can't start at 20.10?)

EllieNodder (approved):
Looks good to me

gemmadallmandfe (approved):
Looks good to me - happy for you to merge this first thing tomorrow @philspray

PR #3432, opened 2023-07-24
Creating on campus landing page

philspray:
### Trello card

https://trello.com/c/oNDULNHT

### Context

The on-campus team will be running activity from early September. We want to create a landing page to encourage mailing list sign-ups.

- created page


gemmadallmandfe:
Looks good @philspray !

Just to put a slight spanner in the works, I wonder if the URL needs changing (again!)?

The on campus team talked about pointing people to mailing list, events and adviser signups so I could see a scenario whereby we have separate landing pages for all three, so perhaps it might make more sense to name them along the lines of `campus-mailing-list`, `campus-events` and `campus-adviser`?

I also suspect the on campus team might want a pink h1 background rather than purple but I'll leave you to discuss that with them!

PR #3434, opened 2023-07-26
delete financial support for international page

philspray:
### Trello card

https://trello.com/c/DzldAmz2

### Context

We were using this page to redirect to guidance on gov.uk which has now been withdrawn so we need to delete the page and redirect to our content page.

- created redirect
- deleted page


gemmadallmandfe (dismissed):
Looks good to me @philspray 

Do we have to delete the app/views/content/guidance as well? Might be a question for Pavel?

philspray:
I wasn't sure about deleting app/views/content/guidance so I'll check with Pavel when he's back.

PR #3445, opened 2023-08-03
Fixing error on salaries page

EllieNodder:
### Context

Noticed an error on the salaries page where a sentence is in a bullet point but should be a stand alone sentence.


philspray (approved):
Looks good to me.

PR #3459, opened 2023-08-10
Adding information about what you can teach

EllieNodder:
### Trello card

https://trello.com/c/fcvSpS2Z/4893-update-qualifications-page-about-what-you-can-teach

### Context

Advisers pointed out at the conference that if you train to teach a specific subject, that doesn’t mean that’s the only thing you can teach (and rarely ever is).

We should consider adding guidance around this to our qualifications page, and anywhere else relevant on the site.


EllieNodder:
Policy approval given by Gaby

gemmadallmandfe (approved):
Looks good to me

PR #3461, opened 2023-08-10
Adding guidance on completing personal statement

EllieNodder:
### Trello card

https://trello.com/c/h82SvoLt/4894-consider-if-we-should-encourage-candidates-to-spend-more-time-on-their-personal-statement

### Context

Data from Apply analysis shows that unsuccessful Non-EEA candidates in particular are spending a lot less time on their personal statements.

We should add some guidance around this to our personal statements page.


gemmadallmandfe (approved):
Looks good to me

PR #3468, opened 2023-08-14
Editing salaries and benefits page

EllieNodder:
### Trello card

https://trello.com/c/2a9GVAZB/4862-consider-editing-salaries-page-content

### Context

On updating the salaries page with the new pay award, the teacher pay team highlighted that we're not being as representative of an average teacher's salary as we could be.

We should restructure this page to better represent teacher pay and to meet user needs.

### DIFF
--- a/app/views/content/salaries-and-benefits.md
+++ b/app/views/content/salaries-and-benefits.md
@@ -1,6 +1,6 @@
 ---
-title: "Teacher pay"
-heading: "Teacher pay"
+title: "Teacher pay and benefits"
+heading: "Teacher pay and benefits"
 description: |-
   All qualified teachers will have a starting salary of at least £30,000 (or higher in London). Find out about teacher pay ranges and more benefits of teaching.
 date: "2021-06-24"
@@ -10,7 +10,7 @@ navigation: 40
 promo_content:
     - content/train-to-be-a-teacher/promos/mailing-list-promo-salaries
 related_content:
-    Steps to become a teacher : "/steps-to-become-a-teacher"
+    How to become a teacher : "/steps-to-become-a-teacher"
     Train to be a teacher if you have or are studying for a degree : "/train-to-be-a-teacher/if-you-have-a-degree"
     Scholarships and bursaries for training to teach : "/funding-and-support/scholarships-and-bursaries"
     Why I became a teacher : "/blog/the-head-of-science-inspiring-students-in-blackpool"
@@ -58,9 +58,9 @@ $new-salaries$
 
 If you’re a primary or secondary teacher in England, your salary will depend on the type of school you work in, where the school is, and the pay range you’re on. 
 
-These pay ranges apply to maintained schools (schools funded by local authorities). 
+The pay ranges below apply to maintained schools (schools funded by local authorities). 
 
-Non-maintained schools (for example, academies, independent schools and private schools) can set their own pay, but often follow the pay ranges below. 
+Non-maintained schools (for example, academies, independent schools and private schools) can set their own pay, but often follow these pay ranges. 
 
 Figures apply from 1 September 2023.
 
@@ -101,7 +101,7 @@ What this will involve will depend on your school, but could include:
 
 * coaching and mentoring other teachers, including trainees and early career teachers
 * undertaking a national professional qualification (NPQ) 
-* supporting your school to produce schemes of work to reduce workload
+* supporting your school with plans to reduce workload
 
 | Area                                     | Minimum | Maximum |
 | -------                                  | -----   | -----   |


app/views/content/salaries-and-benefits.md
```diff
@@ -56,57 +56,52 @@ keywords:
...
+You’ll start on at least £30,000 (or more in London) as a primary or secondary school teacher in England. 
 
-### Qualified teacher salary
+Your salary will be reviewed every year, with most teachers moving up the pay range annually (this will depend on your school’s performance management arrangements).    
 
-Your school will have their own pay scales for qualified teachers. Pay increases will always be linked to performance, not length of service, and will be reviewed every year.
+A typical teacher with 5 years’ experience will earn at least £41,333 (or more in London).
 
-The teacher pay scales for qualified teachers are:
+### Qualified teacher salary
```
EllieNodder, line 75: @plisovin is there a way we can make this into a caption for the below table, like how it describes in the [WCAG guidelines?](https://www.w3.org/WAI/tutorials/tables/caption-summary/)

EllieNodder:
Approval given by Dominic from the Teacher Pay team

app/views/content/salaries-and-benefits.md
EllieNodder, line 1: @gemmadallmandfe @philspray feedback from Dominic is that we can't use 'teacher pay scales' even though that is the highest search volume term :(  so I've done some more keyword research to make sure we're targetting the next highest terms - am slightly concerned this might have a negative impact on search performance though

> Unfortunately we are going to have to insist that you use range and not scale.  Although they sound similar they actually mean two different things in the pay context.  We used scales prior to 2013 because every pay band had increments and the scale refers to the incremental part.  In 2013 we abolished mandatory spine points (we only have advisory spine points now and only for some of the pay bands – leading practitioners and leadership don’t even have advisory spine points) and changed the legal definition to ranges.  So to be accurate we do have to use range.

app/views/content/salaries-and-benefits.md
EllieNodder, line 75: @gemmadallmandfe @philspray this table now combines the main pay range and upper pay range after feedback from the teacher pay team that most teachers are on these 2 scales and the median teacher is on the upper pay range. Thought by combining them, we're giving users a more realistic view of their likely min/max salaries. I don't use the technical lingo (main pay range, upper pay range) as I'm not sure this is necessary and we can perhaps be clearer without it (keyword research supports this as searches for these terms are low)?

app/views/content/salaries-and-benefits.md
```diff
@@ -56,57 +56,52 @@ keywords:
...
-| Inner London                             | £36,745 | £47,666 |
-
-Most established teachers will earn more than this maximum by progressing onto the upper pay range for teachers, or by becoming a leading practitioner.
+| England (excluding London)     | £30,000 | £46,525 |
+| London fringe                            | £31,350 | £47,839 |
+| Outer London                             | £34,514 | £51,179 |
+| Inner London                             | £36,745 | £56,959 |
 
-If you’re not sure whether you'll be teaching in inner, outer or the fringe of London, talk to your school.
+### Additional payments
```
EllieNodder, line 84: @gemmadallmandfe @philspray I've not included the actual TLR/SEN pay tables here and would be interested in your thoughts on this! I thought it might be enough to highlight the fact that they could be eligible for additional payments without going into the specifics - feel like if we started talking about TLRs 1,2 and 3, we'd have to explain the differences between them which could get more complicated. But equally happy to add them back in if you think worth it!

app/views/content/salaries-and-benefits.md
```diff
@@ -56,57 +56,52 @@ keywords:
...
-The upper pay ranges for teachers are:
+Depending on your responsibilities, you could earn up to an extra £15,690. Your school will determine how much you receive. 
 
-| Area                                     | Minimum | Maximum |
-| -------                                  | -----   | -----   |
-| England (excluding London)     | £43,266 | £46,525 |
-| London fringe                            | £44,579 | £47,839 |
-| Outer London                             | £47,592 | £51,179 |
-| Inner London                             | £52,526 | £56,959 |
+## Career progression
```
EllieNodder, line 94: @gemmadallmandfe @philspray created a new section on career progression to try and further reiterate that the salaries already mentioned are what most teachers can expect to be on, and the salaries in this section (leading practitioner, headteachers) are for if you move further up the ranks (as the teacher pay team have said not many teachers are on the leading practitioner salary)

app/views/content/salaries-and-benefits.md
```diff
@@ -130,29 +125,17 @@ The salary ranges for headteachers are:
 | Outer London                             | £57,124 | £134,765 |
 | Inner London                             | £62,304 | £139,891 |
 
-### Other payments
+## If you do not have qualified teacher status (QTS)
```
EllieNodder, line 128: @gemmadallmandfe @philspray created a new h2 for this section as it just seemed to sit really oddly alongside the other bits

app/views/content/salaries-and-benefits.md
```diff
@@ -182,15 +161,3 @@ It is a 'defined benefit' pension and is:
 * flexible and allows you to take some of it as a tax-free lump sum
 
 You also get other insurance benefits too. Find out more about the [teachers' pension scheme](/what-pension-does-a-teacher-get).
-
-## Support for early career teachers
```
EllieNodder, line 186: @gemmadallmandfe @philspray I removed this section for multiple reasons:

a) the page is quite lengthy so already a lot for users to sift through
b) I feel like we lumped this in here when we created the new support for ECT page but that it doesn't really make sense on a page where we're talking about pay? and given the navigation work, we should be surfacing the support for ECT page in a much better way, so this feels a bit unnecessary now
c) looking in GA, VERY few users are actually clicking through to the ECT page link any way (0.15% of users in the last 6 months)

But again, happy to add back in if you think otherwise!

app/views/content/salaries-and-benefits.md
gemmadallmandfe: @EllieNodder @plisovin We can't have captions on the site yet unfortunately as there is an iOS/VoiceOver bug which renders them meaningless (see https://trello.com/c/mjikZkhz) for iPhone/VoiceOver users - VoiceOver reads out a long string of numbers as part of the caption

app/views/content/salaries-and-benefits.md
gemmadallmandfe: If the data shows that very few users are clicking through, I think that's fine - however, we do need to make sure we are surfacing this page effectively elsewhere (which we should do as part of the nav work)

gemmadallmandfe (commented):
@EllieNodder I've left a few comments but broadly the changes make sense

Let's discuss when you are back!

app/views/content/salaries-and-benefits.md
```diff
@@ -1,8 +1,8 @@
 ---
-title: "Teacher salaries and pay scales"
-heading: "Salaries and benefits"
+title: "Teacher pay"
```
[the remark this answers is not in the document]
EllieNodder: So 'teacher pay' has an avg monthly search volume of 9,900, higher than 'teacher salary' which is 8,100. 'Teacher salary' does have a higher CTR/clicks, though it would be interesting to see why. Is it because the current title/meta is optimised for 'salary' and so uses the same language that users are searching for, making them more likely to click on it? If we changed to 'pay', would this lead to a higher CTR for that term, too? Looking at the SERPs, both terms seem to generate largely the same results.

I think looking at UR is hard because when we've done it in the past, the page has always been called 'salaries' so likely that users will repeat the same language back when talking about it?

The current page title (Teacher salaries and pay scales) doesn't have 'benefits' for space reasons, but if we have to drop 'pay scales' now, we can probs add it back in.

Think it could be an interesting test to see how CTR/clicks are affected, and could keep our eye on and change back if we see no added benefit in terms of clicks/CTR for that term. But equally happy to change back to salary! Thoughts? @gemmadallmandfe @philspray

app/views/content/salaries-and-benefits.md
```diff
@@ -56,57 +56,52 @@ keywords:
...
+These pay ranges apply to maintained schools (schools funded by local authorities). 
 
-Other types of schools (for example, academies) can set their own pay scales.
+Non-maintained schools (for example, academies, independent schools and private schools) can set their own pay. 
 
-Figures apply from September 2023.
+Figures apply from 1 September 2023.
 
-## Primary and secondary teaching salaries
+## Primary and secondary teacher salary
```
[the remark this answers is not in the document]
EllieNodder: yep, just trying to match search terms more closely

app/views/content/salaries-and-benefits.md
EllieNodder: :( sadness

app/views/content/salaries-and-benefits.md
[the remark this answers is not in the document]
gemmadallmandfe: @EllieNodder I'd suggest adding `and benefits` back into the h1/title if we are going to try `Teacher pay` - we do talk (briefly!) about holidays and pensions, and they are both relatively good; it also feels a bit friendlier?

app/views/content/salaries-and-benefits.md
gemmadallmandfe: @EllieNodder We may take a hit on search but the traffic we lose might tend towards existing teachers rather than people interested in teaching, so I think we just need to try and optimise for the latter as best we can

app/views/content/salaries-and-benefits.md
```diff
@@ -56,57 +56,52 @@ keywords:
 
 $new-salaries$
 
-For helping to shape the next generation, you’re entitled to a competitive salary, generous holidays, and a substantial pension.
+If you’re a primary or secondary teacher in England, your salary will depend on the type of school you work in, where the school is, and the pay range you’re on. 
 
-These salaries apply to maintained primary, secondary and special schools in England (schools funded by local authorities).
+These pay ranges apply to maintained schools (schools funded by local authorities).
```
gemmadallmandfe, line 61: @EllieNodder Is it clear enough that `These` refers to the pay ranges below rather than the mention in the preceding sentence?

Would maybe suggest putting '`The pay ranges below apply to` and then changing the following sentence to `often follow these pay ranges.`?

app/views/content/salaries-and-benefits.md
gemmadallmandfe: @EllieNodder I'm happy with this approach if the pay team are - as long as they are comfortable that we are not being misleading by only quoting the highest figure

app/views/content/salaries-and-benefits.md
```diff
@@ -56,57 +56,52 @@ keywords:
...
-If you’re an established and exceptional teacher, and regularly show the highest standards of classroom teaching, you can be put onto a higher pay scale.
+Some schools also have a leading practitioner pay range for teachers with exemplary teaching skills who lead the improvement of teaching in their school.
 
-Although they may not lead departments, leading practitioners coach and mentor other teachers and induct trainees and early career teachers (ECTs).
+What this will involve will depend on your school, but could include:
 
-The teacher pay scales for leading practitioners are:
+* coaching and mentoring other teachers, including trainees and early career teachers
+* undertaking a national professional qualification (NPQ) 
+* supporting your school to produce schemes of work to reduce workload
```
gemmadallmandfe, line 104: @EllieNodder `schemes of work` feels very jargony - I don't know what it means! Is there another way to describe this?

gemmadallmandfe:
@EllieNodder Couple more minor comments for you - but very happy for this to go to the pay team for final approval and to be merged next week (w/c 11/09) - it's a definite improvement!

app/views/content/salaries-and-benefits.md
philspray, line 104:
```suggestion
* supporting your school with plans to reduce workload
```

philspray (commented):
Other than the existing comments, all looks good to me

EllieNodder:
Approval given by policy team @gemmadallmandfe @philspray

gemmadallmandfe (approved):
Looks good to me

PR #3476, opened 2023-08-16
Removing full stop from link on Ukraine page

EllieNodder:
### Context

There's a full stop in the IRP link on the Ukraine page that we should take out.


gemmadallmandfe (approved):
Looks good to me

PR #3516, opened 2023-09-06
Add unsafe-inline and unsafe-eval only in script_src and style_src

plisovin:
### Context

Looks like ```frame_ancestors``` doesn't allow to have ```unsafe-inline``` and ```unsafe-eval``` in the CSP header.

```
The Content-Security-Policy directive 'frame-ancestors' does not support the source expression ''unsafe-inline''
The Content-Security-Policy directive 'frame-ancestors' does not support the source expression ''unsafe-eval''
```

Just putting those 2 in script_src and style_src, I think that was the original intention


asatwal (approved):
👍 LGTM

PR #3518, opened 2023-09-06
Add 'Events' to the extra-navigation component

garethwilliamsdesign:
### Trello card

[Trello 4791](https://trello.com/c/SPQkrvoI)

### Context

We are rearranging the header following user research which suggests that the removal of 'Events' from the main navigation would benefit the structure and usability of the website.

Remove 'Events from the main navigation and add this link to the extra-navigation bar, along with a link to 'Get an advisor'.

### Guidance to review

Review the removal of 'Events' from the main navigation and addition of additional links to the extra-navigation bar.

### DIFF
--- a/app/components/header/extra_navigation_component.html.erb
+++ b/app/components/header/extra_navigation_component.html.erb
@@ -1,12 +1,22 @@
 <%= tag.div(class: classes, data: { controller: "searchbox", "searchbox-search-input-id-value" => search_input_id }, role: "navigation") do %>
   <ul class="extra-navigation__list">
-    <li class="extra-navigation__link extra-navigation__mail">
-      <%= link_to("/mailinglist/signup/name", class: "link--black") do %>
-      Register your interest<!--
-   --><span class="icon icon-mail" aria-hidden="true"></span>
-      <span class="icon icon-mail-white" aria-hidden="true"></span>
-      <% end %>
-    </li>
+    <span class="extra-navigation__flex">
+      <li class="extra-navigation__link extra-navigation__border">
+        <%= link_to("/mailinglist/signup/name", class: "link--black") do %>
+        Sign up for emails
+        <% end %>
+      </li>
+      <li class="extra-navigation__link extra-navigation__border">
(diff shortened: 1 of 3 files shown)


philspray (commented):
Looks good to me, just one tiny thing but are the icons slightly different colours? The calendar one looks darker to me than the email one.

gemmadallmandfe:
> Looks good to me, just one tiny thing but are the icons slightly different colours? The calendar one looks darker to me than the email one.

Agree with this @garethwilliamsdesign @philspray  - the icons do look a little bit odd being different colours

They are also different heights which I think looks slightly awkward on mobile in particular? Howver, I know we have always struggled with icon ratios so I'm not sure whether this is fixable!

app/components/header/extra_navigation_component.html.erb
```diff
@@ -1,12 +1,17 @@
...
-      <% end %>
-    </li>
+    <span class="extra-navigation__flex">
+      <li class="extra-navigation__link extra-navigation__mail">
+        <%= link_to("/mailinglist/signup/name", class: "link--black") do %>
+        Sign up for emails<span class="icon icon-mail" aria-hidden="true"></span><span class="icon icon-mail-white" aria-hidden="true"></span>
+        <% end %>
+      </li>
+      <li class="extra-navigation__link extra-navigation__calendar">
+        <%= link_to("/events", class: "link--black") do %>
```
gemmadallmandfe, line 10: @garethwilliamsdesign @philspray @EllieNodder 

I think we should probably point this link to `/events/about-get-into-teaching-events` rather than the main listing page, as we are really wanting to promote GIT events first and foremost, particularly given this term's events have just been launched?

We do link back to other

gemmadallmandfe: However, if we do that, we will need to consider the link text on either this bonus bar or on the home page events block, as then we'd have two instances of `Find an event` links going to different locations - not good in terms of accessibility

plisovin (approved):
LGTM

gemmadallmandfe (approved):
Looks good to me

PR #3527, opened 2023-09-12
Updating events block on homepage

EllieNodder:
### Trello card

https://trello.com/c/5GoU0pKA/5023-remove-events-from-nav-bar-and-update-home-page-block

### Context

We need to update the events block on the homepage so that it is linking to the /events/about-get-into-teaching-events page rather than /events and tweak the link text so it's not the same as in the bonus bar.


gemmadallmandfe (approved):
Looks good to me

PR #3529, opened 2023-09-12
Add channel id to events page and carry this param to the register form

plisovin:
### Trello card

[Trello-5001](https://trello.com/c/hEzzR5IP/5001-investigate-bam-channel-capture-for-events)

### Context

BAM (on campus agency) and the on campus team need to be able to attribute event sign-ups in the same way that they do for mailing list and adviser

For mailing list and adviser, BAM send traffic to a start of funnel page with a query parameter in the format ?channel=22750050

The channel is captured in hidden fields on the start of the mailing list funnel and the start of the adviser funnel respectively, and passed to CRM with the registration details on completion

BAM want to be able to do the same for events


gemmadallmandfe:
Tested with Tejas - the parameters look like they are passing through fine but the wrong channel was recorded in CRM - we think this is because the service that the API is validating against needs changing

Tejas will provide details!

plisovin:
> Tested with Tejas - the parameters look like they are passing through fine but the wrong channel was recorded in CRM - we think this is because the service that the API is validating against needs changing
> 
> Tejas will provide details!

The correct API method to load the channel IDs updated

app/models/events/steps/personal_details.rb
```diff
@@ -3,12 +3,33 @@ module Steps
     class PersonalDetails < ::GITWizard::Steps::Identity
       attribute :is_walk_in, :boolean
       attribute :event_id
+      attribute :channel_id, :integer
 
       validates :event_id, presence: true
 
+      def channel_ids
```
asatwal, line 10: I believe specs will need updating for new methods added ie `channel_ids` and `save`.

asatwal (commented):
Are there any end to end specs that can check that the channel is now passed through as expected ?

spec/models/events/steps/personal_details_spec.rb
```diff
@@ -34,4 +35,50 @@
       it { is_expected.to be_is_walk_in }
     end
   end
+
```
asatwal, line 38: 💯

spec/helpers/events_helper_spec.rb
```diff
@@ -184,6 +185,14 @@
 
       it { is_expected.to all(have_attributes(title: be_a(String), description: be_a(String), path: be_a(String))) }
     end
+
+    context "when the event has a channel" do
+      let(:params) { { channel: "123" } }
```
ekumachidi, line 190: @asatwal I've added the `channel` param to the spec here for clarity.

spec/helpers/events_helper_spec.rb
```diff
@@ -174,7 +174,7 @@
   describe "#categorise_events" do
     let(:events) { build_list(:event_api, 2) }
 
-    subject(:categorised_events) { categorise_events(events) }
+    subject(:categorised_events) { categorise_events(events, {}) }
```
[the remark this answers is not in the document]
ekumachidi: @asatwal This has been added in a new commit

martyn-w:
Looks good to me; but do not merge until the corresponding changes to the API and the API client are merged and manually released to production

martyn-w (dismissed):
👍  LGTM but do not merge until corresponding changes in the API and API-client are merged and manually deployed

martyn-w (approved):

martyn-w (approved):
👍  LGTM

tejasjinaraldfe (approved):
The creation channels are working when attached to URL: /events/about-get-into-teaching-events and the data is appearing in the CRM as expected.

**?channel=222750033: GITIS - On-campus - Webinars (33)**

- It looks like the creation channel name was changed in CRM to GITIS - On-campus - Virtual Pop-up event. Awaiting confirmation from Sarah BH and Julue Cunnington if that's correct. 
- I did two submissions just to confirm if I was attaching the right channel ID and the result was same on both the occasions.
- This is not an issue and this has been flagged to the related teams.

**?channel=222750045: GITIS - On-campus - Sign up to Advisers (45)**

- This creation channel defaults to GITIS - Events Service which ideally it shouldn't. Narendra is adding this creation channel under the missing subscription channel and it should work as expected.
- This issue is not on our end.

Happy to approve this PR.

PR #3536, opened 2023-09-14
Remove online Q&A events

philspray:
### Trello card

https://trello.com/c/pLbpSuaC

### Context

Events team are not offering online Q&A events during Autumn and Spring so we need to remove them from the frontend of the website.

 - removed types of events accordion from events page
 - removed filter from type of events

### DIFF
--- a/app/views/teaching_events/index.html.erb
+++ b/app/views/teaching_events/index.html.erb
@@ -6,8 +6,6 @@
 
   </header>
 
-  
-
   <div class="teaching-events__container">
     <div class="teaching-events__filter">
       <%= render partial: "teaching_events/index/filter" %>
--- a/app/views/teaching_events/index.html.erb
+++ b/app/views/teaching_events/index.html.erb
@@ -2,18 +2,11 @@
   <header>
     <h1 class="heading-l heading--box-yellow">Events</h1>
 
-    <p>
-      Find out more about getting into teaching at a free event where you can get all your questions answered.
-    </p>
-
-    <h2 class="heading-s">Get Into Teaching events</h2>
-
-    <p>Get expert advice, meet training providers and talk to teachers about a career in teaching. Run by the Department for Education (DfE).</p>
-
-    <h2 class="heading-s">Training provider events</h2>
-<p>Find out about local courses and how to apply. Hosted by teacher training providers.</p>
+    <p>Get expert advice, meet training providers and talk to teachers about a career in teaching at Get into Teaching events run by the Department for Education (DfE). You can also find out about local courses and how to apply at events run by teacher training providers.</p>
 
   </header>
+
+  
 
   <div class="teaching-events__container">
     <div class="teaching-events__filter">
--- a/app/views/teaching_events/index.html.erb
+++ b/app/views/teaching_events/index.html.erb
@@ -3,7 +3,7 @@
     <h1 class="heading-l heading--box-yellow">Events</h1>
 
     <p>Choose from 2 types of free events:</p>
-    
+
     <p><strong>Get Into Teaching events</strong> where you can talk to expert advisers, a range of local training providers and current and former teachers.</p>
 
     <p><strong>Training provider events</strong> where specific teacher training providers will talk about the courses they offer.</p>
--- a/app/views/teaching_events/index.html.erb
+++ b/app/views/teaching_events/index.html.erb
@@ -5,6 +5,14 @@
     <p>
       Find out more about getting into teaching at a free event where you can get all your questions answered.
     </p>
+    
+    <h3 class="heading-s">Get Into Teaching events</h3>
+    
+    <p>Get expert advice, meet training providers and talk to teachers about a career in teaching. Run by the Department for Education (DfE).</p>
+    
+    <h3 class="heading-s">Training provider events</h3>
+<p>Find out about local courses and how to apply. Hosted by teacher training providers.</p>
+
   </header>
 
   <div class="teaching-events__container">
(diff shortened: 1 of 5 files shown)


app/views/teaching_events/index.html.erb
```diff
@@ -5,27 +5,6 @@
     <p>
       Find out more about getting into teaching at a free event where you can get all your questions answered.
     </p>
-
```
gemmadallmandfe, line 8: @philspray Sorry - I wasn't very clear on the Trello ticket about the accordion!

I think we probably still need some description about the two remaining event types - but because there are only two, we might be able to ditch the accordion and just have text on the page

That depends on how much space this takes up on mobile though, and whether that will stop people getting to the filters?

app/views/teaching_events/index.html.erb
```diff
@@ -5,27 +5,6 @@
...
-    <details class="teaching-events__descriptions">
-      <summary class="teaching-events__descriptions--summary">Explore the different types of events</summary>
-
-      <dl>
-        <div>
-          <dt>Get Into Teaching events</dt>
-          <dd>Get expert advice, talk to teachers and connect with people like you considering a career in teaching. Run by the Department for Education (DfE).</dd>
```
gemmadallmandfe, line 15: @philspray If we are adding back in the GIT event description, can we please add 'meet training providers' as well, as this is a really key part of it

app/views/teaching_events/index.html.erb
```diff
@@ -6,26 +6,13 @@
       Find out more about getting into teaching at a free event where you can get all your questions answered.
```
gemmadallmandfe, line 6: @philspray This section feels a bit wordy, particularly on mobile, given there are only two event types now?

Should we shorten the first sentence / get rid of it / reword the other sections - or is it OK as is?

image

image

Paging @EllieNodder for her view as she was around when we first added this intro section/accordion!

app/views/teaching_events/index.html.erb
```diff
@@ -6,26 +6,13 @@
...
     </p>
 
-    <details class="teaching-events__descriptions">
-      <summary class="teaching-events__descriptions--summary">Explore the different types of events</summary>
+    <h2 class="heading-s">Get Into Teaching events</h2>
```
gemmadallmandfe, line 9: The alignment is a bit squiffy in this section now the accordion has been removed @philspray 

image

app/views/teaching_events/index.html.erb
```diff
@@ -2,30 +2,10 @@
   <header>
     <h1 class="heading-l heading--box-yellow">Events</h1>
 
-    <p>
-      Find out more about getting into teaching at a free event where you can get all your questions answered.
-    </p>
+    <p>Get expert advice, meet training providers, and talk to teachers at Get into Teaching events.</p>
```
gemmadallmandfe, line 5:
```suggestion
    <p>Get expert advice, meet training providers, and talk to teachers at Get Into Teaching events.</p>
```

EllieNodder: If the whole purpose of this text is to clarify the different event types, I wonder if we need to be a bit clearer about it? e.g. something more like:

> Choose from 2 types of free events: 
> 
> **Get Into Teaching events** where you can talk to expert advisers, a range of local training providers and current and former teachers. 
> 
> **Teacher training provider events** where specific training providers will talk about the courses they offer.

But then this does bulk up the text again, meaning the filters aren't as high up the screen (but if one of the filters is to choose an event type, and users are skipping past the info that explains this, then that feels a bit pointless).

@gemmadallmandfe @philspray

app/views/teaching_events/index/_filter.html.erb
```diff
@@ -19,7 +19,6 @@
   <div class="teaching-events__filter--group">
     <%= f.govuk_check_boxes_fieldset(:type, legend: { text: "Event type", tag: nil }) do %>
       <%= f.govuk_check_box :type, "git", label: { text: "DfE Get Into Teaching" } %>
-      <%= f.govuk_check_box :type, "onlineqa", label: { text: "DfE Online Q&A" } %>
```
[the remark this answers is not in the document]
gemmadallmandfe: We do use the event type filter to link back to (eg from the About GIT events page - linking back to training provider events only)

I also think if we didn't have this, we'd potentially make it harder to find GIT events, as they are far outnumbered by training provider events - something I'd be really loathe to do

Finally - we may possibly bring back online Q&A events as a separate type in the Spring or Summer team so for now, I'd rather not make any code changes!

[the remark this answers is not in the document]
philspray: Yes it is a good question. Worth a ticket on the backlog to investigate?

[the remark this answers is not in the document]
gemmadallmandfe: I've already got a ticket about back end changes if online Q&A doesn't come back - I'll add a note to that rather than raise a new ticket

But not something I think we should consider in the meantime - not a priority!

app/views/teaching_events/index.html.erb
gemmadallmandfe: Think this makes sense - and good to distinguish between multiple providers at GIT events vs one at provider events?

NB filter calls them 'Training provider events' so worth leading with that for consistency?

Choose from 2 types of free events:

**Get Into Teaching events** where you can talk to expert advisers, a range of local training providers and current and former teachers.

**Training provider events** where specific teacher training providers will talk about the courses they offer.

plisovin (approved):
LGTM

gemmadallmandfe (approved):
Looks good to me - happy for this to be merged

(As discussed, we may want to tweak the event descriptions further but will do so on a separate ticket/PR)

PR #3548, opened 2023-09-20
Remove google optimize

plisovin:
### Trello card

[Trello-5007](https://trello.com/c/PhQ6LOjt/5007-remove-google-optimize-from-site)

### Context

We have replaced Google Optimize with VWO.

We now need to remove Google Optimize from the website.


gemmadallmandfe:
@plisovin Do we need to remove the cookies or is this already covered?

plisovin:
> @plisovin Do we need to remove the cookies or is this already covered?

It's covered by removing ```google_optimize.js``` script which in turn load the ```https://www.googleoptimize.com/optimize.js``` which sets the cookies

PR #3560, opened 2023-09-25
Remove duplicated links from the footer

garethwilliamsdesign:
### Trello card

[Trello 5087](https://trello.com/c/LGgaAufv)

### Context

The footer currently features a list of links that are duplicates of the top navigation. These links are rarely clicked and present issues through their duplication. Removing these links will clean up the footer and remove any issues associated with this duplication.

The only link from this list that should not be removed is ‘Search’. This is due to accessibility issues. As such, this link will be added to the list of links in the bottom half of the footer instead.

Remove the duplicated links from the top half of the footer and add the 'Search' link to the list of links in the bottom half of the footer.

### Guidance to review

Check that the links in the top half of the footer have been removed and that 'Search' is now located alongside the links in the bottom half of the footer.

### DIFF
--- a/app/components/footer_component.html.erb
+++ b/app/components/footer_component.html.erb
@@ -42,7 +42,7 @@
     </div>
     <div class="site-footer-bottom">
       <div class="site-footer-bottom__logo-container">
-        <%= image_pack_tag 'static/dfelogo-white.svg', alt: "Department for Education", size: "124x73" %>
+        <%= image_pack_tag 'static/dfelogo-white.svg', alt: "Department for Education logo", size: "124x73" %>
       </div>
       <div class="site-footer-bottom__links-container">
         <%= link_to("Search", search_path) %>
--- a/app/components/footer_component.html.erb
+++ b/app/components/footer_component.html.erb
@@ -37,13 +37,8 @@
           <span class="visually-hidden">Youtube - opens in a new tab</span>
         </a>
       </div>
-      <div class="site-footer-top__links-container">
-      </div>
     </div>
     <div class="site-footer-bottom">
-      <div class="site-footer-bottom__logo-container">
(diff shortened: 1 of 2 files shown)


app/components/footer_component.html.erb
```diff
@@ -38,19 +38,14 @@
...
-          <a href="<%= resource.path %>"><%= resource.title %></a>
-        <% end %>
-        <%= link_to "Find an event near you", events_path %>
-        <%= link_to "Sign up for updates", "/mailinglist/signup/name" %>
-        <%= link_to("Search", search_path) %>
       </div>
     </div>
     <div class="site-footer-bottom">
       <div class="site-footer-bottom__logo-container">
         <%= image_pack_tag 'static/dfelogo-white.svg', alt: "Department for Education", size: "124x73" %>
```
gemmadallmandfe, line 45:
```suggestion
        <%= image_pack_tag 'static/dfelogo-white.svg', alt: "Department for Education logo", size: "124x73" %>
```

I think we should change this alt text to be slightly more descriptive @garethwilliamsdesign (I know this is what it is in prod currently!)

gemmadallmandfe, line 45: When using a screen reader on VoiceOver, the focus order feels wrong - the DfE logo alt text is read out between the YouTube link and the Search link

(It is not tabbed to during focus order, just read out by screen readers)

I'm not sure whether you can do anything about this?

garethwilliamsdesign: I don't really understand why that would be the case as the code is clearly in a logical order. I'll take a look into this today.

gemmadallmandfe (approved):
Looks good to me - happy for this to go out on **Monday** (subject to a tech review)

PR #3566, opened 2023-09-28
Add AKS review app terraform

RMcVelia:
### Trello card

https://trello.com/c/lTSOYCTM/589-deploy-application

### Context

Add review apps for AKS

Add code for AKS review app deployment
- terraform code
- Makefile changes
- change rails database.yml to use DATABASE_URL for aks
- exclude healthcheck from ssl

### Guidance to review

- check the existing paas review app for this PR is working as expected
- make review_aks terraform-apply-aks PR_NUMBER=n IMAGE_TAG=sha-2e9c219

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -22,7 +22,11 @@ test:
 
 production: &production
   <<: *default
+  <% if ENV.key?("VCAP_APPLICATION") %>
   url: <%= Rails.application.config.x.vcap_services.dig("postgres", 0, "credentials", "uri") %>
+  <% else %>
+  url: <%= ENV["DATABASE_URL"] %>
+  <% end %>
 
 rolling:
   <<: *production
--- a/terraform/aks/application.tf
+++ b/terraform/aks/application.tf
@@ -1,0 +1,50 @@
+locals {
+  environment  = "${var.environment}${var.pr_number}"
+}
+
+module "application_configuration" {
+  source = "./vendor/modules/aks//aks/application_configuration"
+
+  namespace              = var.namespace
+  environment            = local.environment
+  azure_resource_prefix  = var.azure_resource_prefix
+  service_short          = var.service_short
+  config_short           = var.config_short
+  secret_key_vault_short = "app"
+
+  # Delete for non rails apps
+  is_rails_application = true
+
+  config_variables = {
+    PGSSLMODE        = local.postgres_ssl_mode
+  }
+  secret_variables = {
+    DATABASE_URL = module.postgres.url
+    REDIS_URL    = module.redis-cache.url
+# below added from paas config
+    HTTPAUTH_PASSWORD = module.infrastructure_secrets.map.HTTP-PASSWORD,
+    HTTPAUTH_USERNAME = module.infrastructure_secrets.map.HTTP-USERNAME,
+    BASIC_AUTH        = var.basic_auth,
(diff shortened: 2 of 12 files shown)


config/database.yml
```diff
@@ -21,9 +21,17 @@ test:
   port: <%= ENV['DB_PORT'] %>
 
 production: &production
+  <% if ENV.key?("VCAP_APPLICATION") %>
+
   <<: *default
   url: <%= Rails.application.config.x.vcap_services.dig("postgres", 0, "credentials", "uri") %>
 
+  <% else %>
+  <<: *default
```
saliceti, line 30: Can this line be moved outside of the if so it's not repeated?

terraform/aks/application.tf
```diff
@@ -0,0 +1,50 @@
...
+
+module "web_application" {
+  source = "./vendor/modules/aks//aks/application"
+
+  is_web = true
+
+  namespace    = var.namespace
+  environment  = local.environment
+  service_name = var.service_name
+  probe_path   = "/healthcheck.json"
```
saliceti, line 42: The healthcheck.json may be too heavy for the probe as it checks all the dependencies, as per: https://github.com/DFE-Digital/get-into-teaching-app/blob/4aa365ce19dd09d6da9c08ba14e52ec9d4abac06/app/models/healthcheck.rb#L4
Can we replace it with a simple one like: https://github.com/DFE-Digital/teacher-pay-calculator/blob/main/config/routes.rb#L11

terraform/aks/database.tf
```diff
@@ -0,0 +1,31 @@
...
+  config_short                = var.config_short
+  cluster_configuration_map   = module.cluster_data.configuration_map
+  use_azure                   = var.deploy_azure_backing_services
+  azure_enable_monitoring     = var.enable_monitoring
+  azure_enable_backup_storage = var.enable_postgres_backup_storage
+  server_version              = "14"
+}
+
+
+module "redis-cache" {
```
saliceti, line 18: Can we pin server_version to 6?

terraform/aks/output.tf
```diff
@@ -0,0 +1,3 @@
+output "url" {
+  value = "https://${module.web_application.hostname}/"
```
saliceti, line 2: The `url` output may be used here for simplicity (https://github.com/DFE-Digital/terraform-modules/blob/main/aks/application/outputs.tf#L5)

terraform/aks/database.tf
RMcVelia: default is 6 in the module?
https://github.com/DFE-Digital/terraform-modules/blob/main/aks/redis/variables.tf#L49

terraform/aks/output.tf
RMcVelia: The output.tf is created by the new service template. Not sure what to change here?

terraform/aks/database.tf
saliceti: Yes but we may want to default to 7 in the module once it's available. So it should be pinned in GIT so GIT it is not updated automatically.

terraform/aks/output.tf
RMcVelia: updated

terraform/aks/database.tf
RMcVelia: pinned to 6

saliceti (dismissed):
🙏

saliceti (approved):

PR #3580, opened 2023-10-03
Word length added to PS section on Train to teach

laurahermionetennant92:
Word length added to PS section on Train to teach


philspray (approved):
Looks good to me.

PR #3591, opened 2023-10-06
fix command instruction

StackedPancakez:
### Trello card
None, quick ad hoc change.

### Context
The development setup steps show an incorrect command, fixed it to match what is in the Makefile.


asatwal (approved):
👍

PR #3595, opened 2023-10-06
Increase spacing between page numbers

garethwilliamsdesign:
### Trello card

[Trello 5116](https://trello.com/c/DlPHB0ig)

### Context

Silktide has identified that our page numbers are too close together on pages like /events and /blog. We therefore need to increase the spacing.

Add margin to the page numbers in order to increase the spacing between them.

### Guidance to review

Check that the page numbers no have 4px margin either side. Or, comparing against the live site, visually check that you can see an increase in the spacing between each page number.

### DIFF
--- a/app/webpacker/styles/components/pagination.scss
+++ b/app/webpacker/styles/components/pagination.scss
@@ -11,7 +11,7 @@
 
   .prev,
   .first {
-    padding-right: 8px;
+    padding-right: 15px;
   }
 
   .next,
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
@garethwilliamsdesign Spacing looks good

My only comment would be that on narrow screens, this now wraps the `Next` onto a line below
image

(iPhone 4 / Galaxy Fold in the emulator)

I appreciate you don't have much width to play with here but I suspect that the `Next` link is now going to be too close to the page numbers!

I'm not sure what proportion of our users use screens this narrow (but we can find out)

gemmadallmandfe (approved):
I think we might still have an issue with the wrapping on very narrow screens in that the second line is probably too close to the first

However, I'm happy for us to merge this as is as it is definitely an improvement on what is in prod currently - and then we can revisit this again if it is still an issue

Just needs a tech eye from @StackedPancakez !

StackedPancakez (approved):
Looks good from my end. First time I have encountered Sass `mq` and use of until, so an interesting TIL for me as well!

PR #3597, opened 2023-10-10
Remove the span tag from ul list, update class of list to be the flex container, move margin from being on the right of the span, to the left of the search icon

StackedPancakez:
### Trello card
[5110](https://trello.com/c/zDW2QhTg/5110-fix-issues-with-list-html-in-bonus-bar)

### Context
Silktide has flagged an accessibility issue with the changes to the bonus bar as their is a `span` tag as a direct descendant of a `ul` tag in the top navigation with the search.

Removed the `span` tag within the `ul` list, and made the `ul` list act as the flex container. Adjust the margin of the search icon to be the equivalent that was previously on the `span` but in the opposite direction so that it displays with the same spacing as before.

### Guidance to review
Please check on the review apps that it looks and behaves as expected. And inspecting the elements to ensure that the only direct descendants of the `ul` tag are `li` tags and nothing else.

### DIFF
--- a/app/components/header/extra_navigation_component.html.erb
+++ b/app/components/header/extra_navigation_component.html.erb
@@ -1,6 +1,5 @@
 <%= tag.div(class: classes, data: { controller: "searchbox", "searchbox-search-input-id-value" => search_input_id }, role: "navigation") do %>
-  <ul class="extra-navigation__list">
-    <span class="extra-navigation__flex">
+  <ul class="extra-navigation__list extra-navigation__flex">
       <li class="extra-navigation__link extra-navigation__mail">
         <%= link_to("/mailinglist/signup/name", class: "link--black") do %>
         Sign up for emails<span class="icon icon-mail" aria-hidden="true"></span><span class="icon icon-mail-white" aria-hidden="true"></span>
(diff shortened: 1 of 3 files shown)


garethwilliamsdesign (changes requested):
The reason I'd included the `span` tag is to force that portion of the list to float to the left of the container on mobile. You can see the design in question in this [Trello ticket](https://trello.com/c/SPQkrvoI). So, we'll need to find a way to force the 'email' and 'event' links to the LHS of the container on mobile.

StackedPancakez:
@garethwilliamsdesign For now I have updated it to push the links to the left when on mobile. What I haven't done yet is what I think we may also need to do based on the picture you sent over slack, which is remove the icons, and include a separator in between the links when on mobile? If you could confirm if we need to make those changes also, or we can have a catchup / chat tomorrow, then we can get into next steps for the PR.

garethwilliamsdesign:
No, no! You’ve nailed it! That’s exactly what we wanted. We want to retain the icons, so, I’m happy for this to be merged!

asatwal (commented):
👍 LGTM

PR #3612, opened 2023-10-12
Update to latest internship providers list with new order of attributes

StackedPancakez:
### Trello card
[5135](https://trello.com/c/PxcxSHoo/5135-reorder-internship-listings-and-upload-revised-details)

### Context
New internship provider list provided so needs to be updated. The changes for new order are in an already merged separate PR. The new order seen in the list is the output of that change.

New internship provider list

### Guidance to review
Check review apps to see if info is correct.


gemmadallmandfe (approved):
Looks good to me

PR #3655, opened 2023-10-26
Add simple feature flag for feedback that can be enabled / disabled per env

StackedPancakez:
### Trello card
[4926](https://trello.com/c/nMIiZhyX/4926-support-paas-migration-of-git-website)

### Context
A migration of the production environment from PAAS to AKS will take place next week. For this, we want to temporarily disable anyone accessing or using the feedback pages.

Add a simple feature flag that can be set per environment to enable / disable access to the feedback pages.

### Guidance to review
Currently development (and therefore also review apps) are set with the feature flag turned on, so testing should be that you are able to use the feedback functionality with no issues, if you would like to test what will happen when it is turned off, please let me know and I can push a change that turns it off for the development environment as well (which I will revert before merging).

### DIFF
--- a/app/components/footer_component.html.erb
+++ b/app/components/footer_component.html.erb
@@ -41,7 +41,7 @@
     <div class="site-footer-bottom">
       <div class="site-footer-bottom__links-container">
         <%= link_to("Search", search_path) %>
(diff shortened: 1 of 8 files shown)


asatwal (approved):
👍 LGTM

PR #3675, opened 2023-11-01
Remove api.smooch.io from zendesk list of secure headers

StackedPancakez:
Remove api.smooch.io from secure_headers.


gemmadallmandfe (approved):
Looks good to me

PR #3686, opened 2023-11-07
Fixed breadcrumb pseudo-element issue

garethwilliamsdesign:
### Trello card

[Trello 5217](https://trello.com/c/dbccy9Yx)

### Context

On the ‘Teacher training funding’ page, in the breadcrumb trail below the header navigation, there is a separator arrow, added with a CSS ::before pseudo-element, before each link in the breadcrumb trail except the first link. With JAWS screen reader, the virtual cursor silently stops at these elements without announcement.

This affects JAWS screen reader users who encounter an additional unexplained silent stop whilst browsing in context, which could prove confusing to them.

Ensure that purely decorative content is hidden from assistive technologies.

### Guidance to review

Test the fix using a screenreader. The virtual cursor should no longer recognise the pseudo-element.

### DIFF
--- a/app/webpacker/styles/components/breadcrumb.scss
+++ b/app/webpacker/styles/components/breadcrumb.scss
@@ -12,17 +12,18 @@
 
     .govuk-breadcrumbs__list .govuk-breadcrumbs__list-item {
       &:not(:first-child) {
-        margin-left: 20px;
+        margin-left: 16px;
         padding-left: 20px;
 
         &:before {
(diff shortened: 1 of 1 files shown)


gemmadallmandfe (approved):
Looks good to me from a front end perspective - tested with VoiceOver. When tabbing, there is no longer a silent stop on the breadcrumb arrow; instead the arrow is skipped as per keyboard users

PR #3688, opened 2023-11-08
Update video titles on About GIT events page

gemmadallmandfe:
### Trello card
https://trello.com/c/G4BmLwd6

### Context
The iframe titles for the two videos on the About Get Into Teaching events page are the same, which is confusing for screen reader users.


StackedPancakez (approved):
LGTM

PR #3699, opened 2023-11-13
Fixed hierarchy on the 'About GIT Events' page

garethwilliamsdesign:
### Trello card

[Trello 5221](https://trello.com/c/Eyhz0AR0)
[Trello 5220](https://trello.com/c/PlOQWjcb)
[Trello 5219](https://trello.com/c/5tSGT5nS)

### Context

#5221: The text ‘What to expect’ acts like a heading, as it introduces content and is visually different than the rest of text, but only has span tags .

This can be problematic for some assistive technology users as they can navigate a page solely using headings, meaning all information not segregated with headings could be missed.

Furthermore, the level 2 headings ‘At our Get Into Teaching events’ and ‘Reserve your place’ are introduced by ‘What to expect’, therefore, they should be acting as level 3 headings instead.

#5220: Headings had been used in a way that did not relay the relationships correctly to screen reader users. The ‘Events in your region’ heading introduces the headings ‘East of England’, ‘North East’, ‘London’ and ‘North West’, but the same heading level has been applied i.e. h2.

This can be confusing and disorientating for screen reader users as the heading structure is not conveyed correctly

#5219: Headings had been used in a way that did not relay the relationships correctly to screen reader users. The level 3 heading ‘What they’re saying’ is being introduced by the level 2 heading ‘Can’t find what you’re looking for?’. This is incorrect as they are not related and makes the heading structure illogical, which might confuse and disorientate screen reader users. 

#5221: Review which headings are needed and what the structure should be (not necessarily following DAC's solution).

#5220: Change listings on 'About GIT Events' page to have h3 headings rather than h2.

#5219:

1. Consider if 'Can't find what you're looking for' should be an h3 to sit under 'Events in your region'.
2. Change 'What they're saying' heading on About GIT events page from h3 to h2.
3. Change 'What they're saying' heading on individual events pages from h3 to h2 (presumably this is a shared component and this is not needed as a separate step though?).

### Guidance to review

Check the above solutions have been implemented.

### DIFF
--- a/app/components/categories/card_component.rb
+++ b/app/components/categories/card_component.rb
@@ -1,7 +1,7 @@
 class Categories::CardComponent < ViewComponent::Base
   attr_reader :title, :description, :path, :heading_tag
 
-  def initialize(card:, heading_tag: "h3")
+  def initialize(card:, heading_tag: "h2")
     @title       = card.title
     @description = card.description
     @path        = card.path
--- a/app/components/categories/card_component.rb
+++ b/app/components/categories/card_component.rb
@@ -1,7 +1,7 @@
 class Categories::CardComponent < ViewComponent::Base
   attr_reader :title, :description, :path, :heading_tag
 
-  def initialize(card:, heading_tag: "h2")
+  def initialize(card:, heading_tag: "h3")
     @title       = card.title
     @description = card.description
     @path        = card.path
--- a/app/views/teaching_events/about_git_events.html.erb
+++ b/app/views/teaching_events/about_git_events.html.erb
@@ -6,7 +6,7 @@
 
 <div class="row inset">
   <section class="col col-720 statement">
-    <span class="pink xlarge bold">What to expect</span>
+    <h2 class="pink xlarge bold">What to expect</h2>
   </section>
 
   <section class="col col-720">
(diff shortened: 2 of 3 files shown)


app/views/teaching_events/about_git_events.html.erb
```diff
@@ -55,7 +55,7 @@
             </ul>
           <% end %>
           <div>
-            <h2 class="heading-s">Can't find what you're looking for?</h2>
+            <h4 class="heading-s">Can't find what you're looking for?</h2>
```
gemmadallmandfe, line 58: I don't think this should be an h4 @garethwilliamsdesign @MaxineCl @caitlin-reidy? As an h4, it is nesting under the last event listed at the moment (which in this environment is the National one)

Shouldn't be just be an h3 - so under `Events in your region` but on the same level as each event listing?

app/components/categories/card_component.rb
```diff
@@ -1,7 +1,7 @@
 class Categories::CardComponent < ViewComponent::Base
   attr_reader :title, :description, :path, :heading_tag
 
-  def initialize(card:, heading_tag: "h2")
+  def initialize(card:, heading_tag: "h3")
```
gemmadallmandfe, line 4: It looks this has affected the category pages @garethwilliamsdesign 

For example, on Train to be a teacher category page:

**Production - top category cards are h2s**
image

**Review app - top category cards are h3**
image

Let's discuss the best way to address this - it might be that we need to add an h2 subcategory heading on the top of the Train to be a teacher page, like there is on Is teaching right for me, as this changes the card level to be h3:

**Production**
image

gemmadallmandfe (approved):
Looks good to me

The 'Can't find what you're looking for heading is temporarily the wrong level within this structure but will be the right level once the component cards heading issue is resolved separately, so happy to leave this for now

PR #3706, opened 2023-11-21
Create separate Events card component

garethwilliamsdesign:
### Trello card

[Trello 5220](https://trello.com/c/PlOQWjcb)

### Context

Headings had been used in a way that did not relay the relationships correctly to screen reader users. The ‘Events in your region’ heading introduces the headings ‘East of England’, ‘North East’, ‘London’ and ‘North West’, but the same heading level has been applied i.e. h2.

Create a new card component for the 'About Get Into Teaching Events' page which allows us to change the headings from h2s to h3s.

### Guidance to review

Check the headers on the card component on the /events/about-get-into-teaching-events page are h3s.

### DIFF
--- a/app/components/events/card_component.html.erb
+++ b/app/components/events/card_component.html.erb
@@ -1,0 +1,8 @@
+<li class="category__nav-card">
+  <%= link_to(path, class: "link--black") do %>
+    <div class="category__nav-card--content">
+      <%= content_tag(heading_tag, title) %>
+      <%= tag.p(helpers.safe_html_format(description)) %>
+    </div>
+  <% end %>
+</li>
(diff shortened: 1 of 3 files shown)


gemmadallmandfe (commented):
Looks good to me and will make future maintenance easier @garethwilliamsdesign 

However, it has failed a sonarcloud test (although I'm not sure why as it seems to me that there is a `<ul>` container)

Happy for this to be merged once this has been sorted out (if needed)

gemmadallmandfe (approved):
John has confirmed that this shouldn't be a problem as the `ul` is outside the component but still on the page

Happy for this to be merged tomorrow morning

StackedPancakez (approved):
LGTM, sonarqube error is a non-error, in the sense that it just can't see across view files and see that the li is in a ul via component rendering. So all good I think.

PR #3712, opened 2023-11-24
Add html html attributes for accessibility audit for screen readers

StackedPancakez:
### Trello card
[5227](https://trello.com/c/pviWctFJ/5227-dac-audit-2023-cookie-modal-focus-escaping)

### Context
Screen reader does not focus correctly on cookie modal.

Add `role="dialog"` and `aria-modal="true"` html attributes to the element to ensure screen reader picks it up.

### Guidance to review
Use a screen reader like voice over on your phone, and swipe through page once the cookies modal appears, it should first step through the elements on the cookie modal, not the page behind.


gemmadallmandfe (approved):
Looks good to me from a front end perspective - I have tested using VoiceOver on iPhone and focus is initially on and remains on the cookie modal, whereas previously the focus on the content behind the modal

PR #3717, opened 2023-11-28
Missing expanded aria menu

StackedPancakez:
### Trello card
[5243](https://trello.com/c/4CjQRgKC/5243-dac-audit-2023-missing-expanded-aria-2-menu-button-on-mobile)

### Context
The accessibility audit showed that these expandable components did not have the correct aria attributes available. These needed to be added and values updated accordingly with user action.

Add the role, aria-expanded and aria-controls attributes to the relevant html elements, and include logic to adjust accordingly.

### Guidance to review
In the review app ensure html attributes are present and changing as expected in the DOM when each of the toggle elements are clicked by the user.


gemmadallmandfe (approved):
Tested successfully using VoiceOver on iPhone- happy for this to be merged

martyn-w (approved):
👍  LGTM

PR #3724, opened 2023-11-30
Update brand colour palette

garethwilliamsdesign:
### Trello card

[Trello 5302](https://trello.com/c/IYgHMJlP)

### Context

Havas are supplying an updated brand colour palette for the new creative campaign. The colours will meet digital accessibility standards for WCAG 2.1 AA. We therefore need to update the website colour palette.

Replace the old colour palette with the updated version and check how this impacts the website as a whole.

### Guidance to review

Sitewide.


gemmadallmandfe:
@garethwilliamsdesign Do the cookie modal hover state and the events link hover state meet the required contrast?

peteryates (approved):
It's a thing of beauty.

gemmadallmandfe:
@garethwilliamsdesign I've found another green that needs fixing - on `/landing/food-for-thought`

image


Can you either add the change to this PR or raise a separate ticket to be addressed separately? Thanks!

app/webpacker/styles/colours.scss
```diff
@@ -1,24 +1,25 @@
 $black: #1d1d1b;
-$blue: #2881be;
-$blue-dark: #0076bd;
+$blue: #00b1eb;
+$blue-dark: #1877cf;
```
gemmadallmandfe, line 3: @garethwilliamsdesign This link blue that we were given is actually slightly worse in terms of colour contrast ratio (4.59 compared with our current 4.85).

Although this meets AA for normal, is it worth using the darker blue that was provided (#1557ac) instead which would then have a ratio of 7.02? But you may think this is not distinct enough from black - have a look and let's discuss!

Alternately we could consider using the GDS blue #1d70b8 (given we are using the GDS CTA green) which would have a ratio of 5.17 - this meets AA and is an improvement on what we have currently in production plus what we have been given as an updated colour

app/webpacker/styles/colours.scss
```diff
@@ -1,24 +1,25 @@
...
+$pink: #f162a3;
 $pink-dark: #d34491;
 $pink-dark-90: rgba($pink-dark, .9);
-$purple: #8967aa;
+$purple: #ad52a4;
 $purple-dark: #7d58a0;
 $purple-dark-90: rgba($purple-dark, .9);
 $white: #ffffff;
 $white-90: rgba($white, .9);
 $yellow-light: #fef0cb;
```
gemmadallmandfe, line 21: @garethwilliamsdesign I will raise a **separate** ticket to review the additional colours such as this `$yellow-light` in case there are further changes you want to make in light of the brand colour changes?

But this would be a low priority and one for a future sprint - not now!

app/webpacker/styles/colours.scss
garethwilliamsdesign: I agree and am happy to use GDS blue. I'll make the change and we can review.

app/webpacker/styles/colours.scss
garethwilliamsdesign: Sounds good.

gemmadallmandfe (approved):
Looks good to me - to be merged on Tues 9 Jan alongside https://github.com/DFE-Digital/get-into-teaching-app/pull/3763

PR #3736, opened 2023-11-30
Add search aria attributes

StackedPancakez:
### Trello card
[5293](https://trello.com/c/ERhhH5wv/5293-dac-audit-2023-missing-expanded-aria-1-search-bar)

### Context
The accessibility audit showed that these expandable components did not have the correct aria attributes available. These needed to be added and values updated accordingly with user action.

Add the role, aria-expanded and aria-controls attributes to the relevant html elements, and include logic to adjust accordingly.

### Guidance to review
In the review app ensure html attributes are present and changing as expected in the DOM when each of the toggle elements are clicked by the user.


gemmadallmandfe (approved):
Tested successfully using VoiceOver on iPhone 11 - expanded and collapsed attributes were both read out as expected

I've also made sure the search is still as expected for keyboard only and mouse users on desktop and for non- screen reader users on mobile

Happy for this to be merged when it has had tech eyes on it!

ekumachidi (approved):
Looks good to me

PR #3744, opened 2023-12-06
Update healthchecks_controller.rb

martyn-w:
De-activate the CSP header on the healthcheck page, to allow the system to be monitored without over-sized CSP headers

### Trello card
[5209](https://trello.com/c/5QRmY2Jz/5209-investigate-header-size-which-is-preventing-statuscake-monitoring)

### Context

The Content Security Policy (CSP) headers of the site were causing the monitoring system (StatusCake) to error as they were oversized (> 17k). 

De-activate the CSP headers on the monitoring page https://getintoteaching.education.gov.uk/healthcheck.json

As a separate task, we will review which headers are still required and try and trim down the list

### Guidance to review

In the review app, navigate to the /healthcheck.json endpoint and verify there are no CSP headers being returned.

### DIFF
--- a/app/controllers/healthchecks_controller.rb
+++ b/app/controllers/healthchecks_controller.rb
@@ -1,5 +1,9 @@
 class HealthchecksController < ApplicationController
-  def show
+   def show
+     # De-activate the CSP header on the healthcheck page, to allow the system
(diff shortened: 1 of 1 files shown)


saliceti (dismissed):
Thanks!

PR #3756, opened 2023-12-13
Remove caption from why-teach component

garethwilliamsdesign:
### Trello card

[Trello 5411](https://trello.com/c/QHDOlYs6)

### Context

The ‘Why teach' subtitle colour contrast with the grey background of this component fails accessibility checks.

We propose to remove the subtitle and update the content of the h2 tag.

### Guidance to review

Check that the 'Why teach?' caption has been removed from the 4 examples of this component that exist. Also, check that the content has been updated to maintain clarity and context.


app/views/content/train-to-be-a-teacher/_why-teach.html.erb
```diff
@@ -1,11 +1,9 @@
-<section class="why-teach" aria-label="Why teach?">
+<section class="why-teach" aria-label="Inspire the next generation">
```
gemmadallmandfe, line 1: @garethwilliamsdesign This component doesn't appear on the front end by the looks of it - should we just delete this file altogether?

garethwilliamsdesign: Sure. I'll do that after UCD.

app/views/content/non-uk-teachers/_why-teach.html.erb
```diff
@@ -1,11 +1,9 @@
-<section class="why-teach" aria-label="Why teach?">
+<section class="why-teach" aria-label="Join a world class education system">
```
MaxineCl, line 1: @garethwilliamsdesign @gemmadallmandfe I think this would be better with the H2 as Why teach in England? and we get rid of the Join a world class education system. But if you disagree and want to leave it I think world-class should be hyphenated.

garethwilliamsdesign: I'm happy with your suggestion if @gemmadallmandfe is, too?

gemmadallmandfe: Yes - happy with @MaxineCl's suggestion!

app/views/content/train-to-be-a-teacher/_why-teach.html.erb
gemmadallmandfe: Isn't `app/views/content/how-to-apply-for-teacher-training/_why-teach.html.erb` the one on the How to apply page?

gemmadallmandfe (approved):
Looks good to me

PR #3762, opened 2023-12-15
Add message re holiday response time on adviser funnel completed page

gemmadallmandfe:
### Trello card
https://trello.com/c/nQlFHYJN

### Context
We need to tell people that it may take longer to be allocated an adviser during the holiday period.


martyn-w (approved):
👍  LGTM

PR #3763, opened 2023-12-15
Changed font colour from white to black in non-CTA components

garethwilliamsdesign:
### Trello card

[Trello 5414](https://trello.com/c/uwgNl4bg)

### Context

We are replacing our colours with refreshed brand colours supplied by Havas. Our h2s have changed from darker blue background with white font to be lighter blue with dark font - dark font is more in keeping with the brand

We want to reconsider where we are still using white font in other places as we want to keep white font just for CTAs as a principle.

All text on a coloured background that is non-CTA should change from being white to black, as per the brand guidelines.

### Guidance to review

Check all non-CTA headers that have a coloured background (such as h2s) and ensure the font colour is black, not white.

### DIFF
--- a/app/webpacker/styles/headings.scss
+++ b/app/webpacker/styles/headings.scss
@@ -157,7 +157,7 @@
   @include highlight-text($purple, $black);
 }
 
-.heading--highlight-white-on-purple {
+.heading--highlight-black-on-purple {
   @include highlight-text($purple, $black);
 }
 
--- a/app/webpacker/styles/headings.scss
+++ b/app/webpacker/styles/headings.scss
@@ -102,7 +102,7 @@
 }
 
 .heading--box-green {
-  @include box-heading($green, $white);
+  @include box-heading($green, $black);
 }
 
 .heading--box-grey {
@@ -114,7 +114,7 @@
 }
 
 .heading--box-purple {
-  @include box-heading($purple, $white);
+  @include box-heading($purple, $black);
 }
 
 .heading--box-pink {
@@ -158,7 +158,7 @@
 }
 
 .heading--highlight-white-on-purple {
-  @include highlight-text($purple, $white);
+  @include highlight-text($purple, $black);
 }
 
 .heading--highlight-blue {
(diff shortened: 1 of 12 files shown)


app/webpacker/styles/headings.scss
```diff
@@ -158,7 +158,7 @@
 }
 
 .heading--highlight-white-on-purple {
```
gemmadallmandfe, line 160: @garethwilliamsdesign this feels quite confusing if it is still called `highlight-white-on-purple`? does the name need changing? I presume this would mean making other changes elsewhere though

gemmadallmandfe:
@garethwilliamsdesign I think the embedded mailing list heading fails colour contrast ratio?

Example on `/landing/train-to-teach`

image

gemmadallmandfe:
@garethwilliamsdesign I think this fails as well:

`/funding-and-support/scholarships-and-bursaries`

image

gemmadallmandfe:
@garethwilliamsdesign I think these fail as well - purple does not work at normal font sizes!

`/events`

image

gemmadallmandfe:
@garethwilliamsdesign And finally the flags:

`/subjects/maths`
`/explore-teaching-advisers`

image


image

app/webpacker/styles/vertical-tags.scss
```diff
@@ -19,8 +19,8 @@
     display: inline-block;
     margin-left: -2em;
 
-    color: $white;
-    border: 1px solid $purple;
-    background: $purple;
+    color: $black;
+    border: 1px solid $blue;
+    background: $blue;
```
gemmadallmandfe, line 24: Should we save blue for the h2s @garethwilliamsdesign - is it a bit confusing to use blue for flag content?

gemmadallmandfe: I know that the headings on these pages are other colours though, so maybe it doesn't matter!

gemmadallmandfe (approved):
Looks good to me

PR #3784, opened 2024-01-08
Update University of Winchester link on Assessment Only page

gemmadallmandfe:
### Trello card
https://trello.com/c/SiSrnPor

### Context
University of Winchester have moved their assessment only page, so we need to update the link


martyn-w (approved):
👍  LGTM

PR #3764, opened 2023-12-18
Remove Zimbabwe and direct links to Apply for QTS service from Teach in England

laurahermionetennant92:
Remove Zimbabwe and direct links to service from Teach in England

PR #3793, opened 2024-01-10
Fix broken link on /train-to-be-a-teacher/assessment-only-route-to-qts

martyn-w:
Fixes broken link following section rename in https://github.com/DFE-Digital/get-into-teaching-app/pull/3764


laurahermionetennant92 (approved):
all looks good

PR #3798, opened 2024-01-11
Update career changer page with campaign messaging.

MaxineCl:
### Trello card
https://trello.com/c/2EHQde8t/1924-update-content-on-how-do-i-change-to-a-career-in-teaching-page-to-reflect-campaign-messaging

### Context
We need to make updates to the website to align messaging with the new creative campaign launching on 15 January 2024

Update the content in the first paragraph to make it more inspiring and align with the campaign messaging.

### DIFF
--- a/app/views/content/is-teaching-right-for-me/if-you-want-to-change-career.md
+++ b/app/views/content/is-teaching-right-for-me/if-you-want-to-change-career.md
@@ -17,7 +17,7 @@ navigation_title: How do I change to a career in teaching?
 navigation_description: Use your experience to inspire students and bring your skills and expertise to the classroom.
 ---
 
-Turn your passion into your profession with a career change into teaching. Not only can you use the skills and expertise you’ve already developed to inspire young people, teaching gives you the opportunity to grow and fulfill your own potential too.
+Turn your passion into your profession with a career change into teaching. Not only can you use the skills and expertise you’ve already developed to inspire young people, teaching gives you the opportunity to grow and fulfil your own potential too.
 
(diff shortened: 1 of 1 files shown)


app/views/content/is-teaching-right-for-me/if-you-want-to-change-career.md
```diff
@@ -17,7 +17,7 @@ navigation_title: How do I change to a career in teaching?
 navigation_description: Use your experience to inspire students and bring your skills and expertise to the classroom.
 ---
 
-A career change into teaching allows you to use the skills and expertise you’ve already developed to inspire young people.
+Turn your passion into your profession with a career change into teaching. Not only can you use the skills and expertise you’ve already developed to inspire young people, teaching gives you the opportunity to grow and fulfill your own potential too.
```
gemmadallmandfe, line 20:
```suggestion
Turn your passion into your profession with a career change into teaching. Not only can you use the skills and expertise you’ve already developed to inspire young people, teaching gives you the opportunity to grow and fulfil your own potential too.
```
Think `fulfil` is more usual in UK? But one to check!

Otherwise changes look good to me!

MaxineCl: I had fulfil originally and it showed up as being the wrong spelling! So I added an l. I've double checked and it's slightly contradictory but The Guardian style guide, which I defer to after GOV.UK, has one l so I think you're right we should go with fulfil.

gemmadallmandfe (approved):
Looks good to me

PR #3806, opened 2024-01-17
added content for chatGPT

caitlin-reidy:
### Trello card
https://trello.com/c/sNeWp0IZ/5335-update-personal-statement-page-with-guidance-on-using-ai

### Context
Add content on chatGPT in line with what find and apply have produced

Added content to the first paragraph about chatGPT

### DIFF
--- a/app/views/content/how-to-apply-for-teacher-training/teacher-training-personal-statement.md
+++ b/app/views/content/how-to-apply-for-teacher-training/teacher-training-personal-statement.md
@@ -59,6 +59,11 @@ Teacher training providers want to see your passion and that you understand the 
 
 Make sure you check your spelling and grammar in your application. You want to make the best possible impression.
 
+You can use ChatGPT or other artificial intelligence (AI) tools to help you write your personal statement. Remember that 
+
+*  AI tools are easy to notice because they use bland language and will not be able to give details about you as a person
+* your account to apply for teacher training may be blocked if you consistently submit personal statements that look like they have been written with AI tools
+
 ## Do I use the same personal statement for each application?
 
 You can use the same personal statement for every course you apply to.


app/views/content/how-to-apply-for-teacher-training/teacher-training-personal-statement.md
```diff
@@ -59,6 +59,11 @@ Teacher training providers want to see your passion and that you understand the
 
 Make sure you check your spelling and grammar in your application. You want to make the best possible impression.
 
+You can use ChatGPT or other artificial intelligence (AI) tools to help you write your personal statement. Remember that 
+
+*  AI tools are easy to notice because they use bland language and will not be able to give details about you as a person
+* your account to apply for teacher training may be blocked if you consistently submit personal statements that look like they have been written with AI tools
+
```
MaxineCl, line 66: @caitlin-reidy I think this might be better as it's own section? I'm not sure it quite fits under How long should it be? Although I can see why you included it in here as it doesn't really fit in the other sections either!

And if you do create a new section on using AI then the sentence above about checking spelling and grammar could potentially go at the end of it? Might make more sense to have it there?

MaxineCl: @caitlin-reidy is there an extra space before AI in the 1st bullet or do you need another space after your? They don't quite look aligned?

MaxineCl: @caitlin-reidy I think we may still need to include that if they use AI they're less likely to be successful. But I think it needs to be clear that they're likely to be unsuccessful if you just rely on AI to write it for you, as opposed to using it to prompt/ help you? 

And the same with the last bullet as I found it a bit confusing as we say it's ok to use it but then your account my get blocked if you keep using it (I appreciate this is what Apply has said). So again I think it's more your account may be blocked if you consistently submit personal statements using AI to write the whole personal statement for you. Rather than using it to help/ improve what you've already written or to inspire you type thing.

caitlin-reidy: @MaxineCl do we want to draw that much attention to chatgpt as having its own section though? Especially seeing as the overarching tone is dont really use/be very careful if you do.

MaxineCl: @caitlin-reidy good point! But I think Apply has it as a section by itself so I think it would be OK? And I think people will use it anyway so we want to make sure if they do they're not giving themselves less of a chance of being successful?

PR #3814, opened 2024-01-22
Change-Talk-to-adviser-H3s

MaxineCl:
### Trello card
https://trello.com/c/0fAiDmAi/5526-switch-h3s-and-content-round-on-adviser-page-cards-on-help-and-support-category-page

### Context
The H3s on the 2 grey cards in the Talk to an adviser section on the help and support category page are difficult to differentiate between and may cause confusion for users who don’t understand there are different types of adviser.

- Switch the H3s and the content round so the h3s make it clear which user group they're for
- Switch the order of the cards round so the if you have a degree one comes before the if you're a final year student as it makes more sense this way round.


gemmadallmandfe (approved):
Looks good to me

PR #3833, opened 2024-01-30
Subcategory added to non-UK category page

laurahermionetennant92:
Subcategory added to non-UK category page

### DIFF
--- a/app/views/content/non-uk-teachers/fees-and-funding-for-non-uk-trainees.md
+++ b/app/views/content/non-uk-teachers/fees-and-funding-for-non-uk-trainees.md
@@ -1,6 +1,6 @@
 ---
 title: "Fees and financial support for non-UK trainee teachers"
-subcategory: Train to teach in England
+subcategory: If you want to train to teach
 description: |-
   Learn more about funding and financial support for non-UK students, including details of the international relocation payment.
 related_content:
--- a/app/views/content/non-uk-teachers/fees-and-funding-for-non-uk-trainees.md
+++ b/app/views/content/non-uk-teachers/fees-and-funding-for-non-uk-trainees.md
@@ -1,5 +1,6 @@
 ---
 title: "Fees and financial support for non-UK trainee teachers"
+subcategory: Train to teach in England as a non-UK citizen
 description: |-
   Learn more about funding and financial support for non-UK students, including details of the international relocation payment.
 related_content:
@@ -7,7 +8,6 @@ related_content:
     Teacher pay and benefits in England : "/is-teaching-right-for-me/teacher-pay-and-benefits"
     A day in the life of a teacher in England : "/a-day-in-the-life-of-a-teacher"
 promo_content:
-
 navigation: 20.20
 navigation_title: Fees and financial support for non-UK trainee teachers
 navigation_description: Learn more about teacher training fees in England and financial help for physics and languages trainees.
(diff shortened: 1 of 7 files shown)


app/views/content/non-uk-teachers/fees-and-funding-for-non-uk-trainees.md
```diff
@@ -1,13 +1,13 @@
 ---
 title: "Fees and financial support for non-UK trainee teachers"
+subcategory: Train to teach in England as a non-UK citizen
```
MaxineCl, line 3: @laurahermionetennant92 I don't think the sub category title should be the same as the title of one of the cards. So I think it could be good to make it less similar and a little more broad - maybe something along the lines of 9but not this) if you want to train to teach.

We also have non-UK in the page title so I don't think we need to include it in the sub head for this or the teacher one.

MaxineCl, line 3: @laurahermionetennant92 I don't think the sub category title should be the same as the title of one of the cards. So I think it could be good to make it less similar and a little more broad - maybe something along the lines of 9but not this) if you want to train to teach.

We also have non-UK in the page title so I don't think we need to include it in the sub head for this or the teacher one.

app/views/content/non-uk-teachers/ukraine.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: "Ukrainian teachers and trainees coming to the UK"
+subcategory: Train to teach in England as a non-UK citizen
```
MaxineCl, line 3: @laurahermionetennant92 as Ukraine covers both trainees and teachers I think it's a little confusing having it in the train to teach section and it might make more sense to have its own separate blue subhead at the bottom of the page. Maybe something like If you're from Ukraine?

app/views/content/non-uk-teachers/ukraine.md
```diff
@@ -8,7 +9,7 @@ related_content:
     Who do you want to teach? : "/is-teaching-right-for-me/who-do-you-want-to-teach"
 promo_content:
     - content/non-uk-teachers/promos/adviser-promo-ukraine
-navigation: 20.50
+navigation: 20.40
 navigation_title: Ukrainian teachers and trainees coming to the UK
 navigation_description: Find information and advice for Ukrainian citizens seeking teaching jobs and training places in England.
 date: "2022-06-24"
```
MaxineCl, line 15: @laurahermionetennant92 As the site is focused on getting people into teacher training can we switch these round so the title and description have trainees first and then teachers.

app/views/content/non-uk-teachers/teach-in-england-if-you-trained-overseas.md
```diff
@@ -1,13 +1,14 @@
 ---
 title: "Teach in England if you trained outside the UK"
+subcategory: Teach in England as a non-UK citizen
```
MaxineCl, line 3: @laurahermionetennant92 Could this be something like If you're already a teacher? to make it different to the title on the card?

app/views/content/non-uk-teachers/fees-and-funding-for-non-uk-trainees.md
laurahermionetennant92: Makes sense

laurahermionetennant92:
Hi @MaxineCl I've made those fixes – please can you take another look? Thank you :)

app/views/content/non-uk-teachers/international-qualified-teacher-status.md
```diff
@@ -1,13 +1,14 @@
 ---
 title: Get international qualified teacher status (iQTS)
+subcategory: Get international qualified teacher status (iQTS)
```
MaxineCl, line 3: @laurahermionetennant92 the H2 is the same as the title on the card. Maybe it could be something along the lines of how to train to teach without coming to England - but worded a bit better!

MaxineCl (approved):

PR #3889, opened 2024-03-07
dual courses content

caitlin-reidy:
### Trello card
https://trello.com/c/nKJfbAKt/5482-update-wording-about-dual-training-courses-on-qualifications-page

### DIFF
--- a/app/views/content/is-teaching-right-for-me/qualifications-you-need-to-teach.md
+++ b/app/views/content/is-teaching-right-for-me/qualifications-you-need-to-teach.md
@@ -54,10 +54,9 @@ This could be through having a degree in the subject, or having:
 * a degree related to but not in the subject – for example, your degree is in engineering but you’d like to teach physics
 * an unrelated degree but relevant professional experience
 
-You can also train to teach two subjects at primary or secondary. You can:
+You can also train to teach primary with a specialist subject, for example primary with maths, so you can develop your knowledge of that particular subject.
 
-* choose primary with a subject, this will allow you to teach more of that other subject in your primary school
-* choose two subjects at secondary, this will allow you to teach both subjects in your secondary school
+Or you can train to teach two subjects at secondary, for example PE with computing, so you can develop the skills to teach in an additional subject.
 
 Talk to your training provider to find out what you can train to teach.
 
--- a/app/views/content/is-teaching-right-for-me/qualifications-you-need-to-teach.md
+++ b/app/views/content/is-teaching-right-for-me/qualifications-you-need-to-teach.md
@@ -56,7 +56,7 @@ This could be through having a degree in the subject, or having:
 
 You can also train to teach primary with a specialist subject, for example primary with maths, so you can develop your knowledge of that particular subject.
 
-Or you can train to teach two subjects at secondary, for example PE with computing, so you can develop the skills to teach in an additional subject.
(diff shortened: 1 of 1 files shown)


app/views/content/is-teaching-right-for-me/qualifications-you-need-to-teach.md
```diff
@@ -54,12 +54,15 @@ This could be through having a degree in the subject, or having:
 * a degree related to but not in the subject – for example, your degree is in engineering but you’d like to teach physics
 * an unrelated degree but relevant professional experience
 
+You can also train to teach two subjects at primary or secondary. You can:
+
+* choose primary with a subject, this will allow you to teach more of that other subject in your primary school
```
gemmadallmandfe, line 59:
```suggestion
* choose primary with a specialist subject, which will allow you to teach more of that subject in your primary school
```
@MaxineCl I'm not sure that `allow` is the right word here?

app/views/content/is-teaching-right-for-me/qualifications-you-need-to-teach.md
```diff
@@ -54,12 +54,15 @@ This could be through having a degree in the subject, or having:
...
+* choose two subjects at secondary, this will allow you to teach both subjects in your secondary school
```
gemmadallmandfe, line 60:
```suggestion
* choose two subjects at secondary, which will allow you to teach both subjects in your secondary school
```
@MaxineCl again, not sure about `allow`

MaxineCl: @gemmadallmandfe for some reason I can't add the comment by the line, the + won't open but I don't think the wording of this sentence is quite right? **You can also train to teach two subjects at primary** or secondary. 

Looking on Find you can apply to train in primary with a specialist subject but this says train to teach 2 subjects and primary isn't a subject?

app/views/content/is-teaching-right-for-me/qualifications-you-need-to-teach.md
```diff
@@ -54,12 +54,14 @@ This could be through having a degree in the subject, or having:
 * a degree related to but not in the subject – for example, your degree is in engineering but you’d like to teach physics
 * an unrelated degree but relevant professional experience
 
+You can also train to teach primary with a specialist subject, for example primary with maths, so you can develop your knowledge of that particular subject.
+
+Or you can train to teach two subjects at secondary, for example PE with computing, so you can develop the skills to teach in an additional subject.
```
gemmadallmandfe, line 59:
```suggestion
Or you can train to teach two subjects at secondary, for example PE with computing, so you can develop the skills to teach an additional subject.
```

gemmadallmandfe: Is PE with computing a common combination? If not, is it worth putting one that is more common?

MaxineCl: @gemmadallmandfe I'm not sure. When I looked on Find it lists courses as PE with an Ebacc subject in the what subject do you want to teach. So I assumed PE with one of these would be the most popular and chose computing so it wasn't the same as the primary example. I can check to see if we have data on which is the most common combination with PE. Or we can say PE with an Ebacc subject and then list out what these are?

gemmadallmandfe (approved):
Looks good to me

PR #3917, opened 2024-03-27
Expandable/dropdown menus

martyn-w:
### Trello card
[Expandable/dropdown menus](https://trello.com/c/nG5KWcY3/5668-build-dropdown-expandable-menus)
New dropdown menu system for mobile and desktop

### Guidance to review
Please test at multiple screen resolutions (mobile, tablet and desktop)


MylesJarvis:
Just playing around in Chrome with a mobile view, when I expand out a section, it seems that there isn't sufficient padding on the left hand side, so the first letter is almost lost: 
image

gemmadallmandfe:
**Keyboard only** 
The menus work using keyboard only but I don't think the order is entirely logical

I can open the `Is teaching right for me` sub-menu using keyboard only but have to tab through all the other top-level options before I get to the sub-menu

https://github.com/DFE-Digital/get-into-teaching-app/assets/87642394/543da8e3-3d3a-41c4-9420-4d3c65ff2b09

https://www.rnib.org.uk/ is a good example of better ordering

@martyn-w

gemmadallmandfe:
Top level menu is missing a hover state @martyn-w ?

Production hover state
image

New hover state (none)
image

gemmadallmandfe:
@jenhadfield-dfe @emmaachesongray I think we need to have a principle of having subcategory headings on each category page, otherwise the menu feels a bit inconsistent

image


eg in this screenshot, content pages are marked in red but are styled in two different ways in the menu
and the category page (marked in blue) looks the same as the content pages above it

This is not the fault of the design in any way (in fact, it copes really well with what it's got to work with!) but more about our inconsistent content

I don't think this should stop us deploying the menus when ready - we can add categories in retrospect (which we are already going to do re the TDA PR)

gemmadallmandfe:
@jenhadfield-dfe @mariosh-e 

Not one for this PR (and it shouldn't hold up the menus deployment!) but just flagging that the grey header on the help and support page merges in with the menu, so we should look to change it soon!

image

gemmadallmandfe:
**FONT COLOUR**

I'm not sure whether this is a real issue or not @jenhadfield-dfe @martyn-w - I get it in the Chrome emulator but can't replicate on my phone (iPhone11/Safari)

The subcategory has white text on a pink background which fails colour contrast (and is also not on brand!)

However, on my phone it is black text on pink, which is fine

Screenshot from emulator:
image

jenhadfield-dfe:
> **DESKTOP HOVER STATE - EXPANDED MENU**
> 
> @jenhadfield-dfe @martyn-w I'm not sure the hover state on the expanded menu items meets accessibility guidelines?
> 
> We need something other than colour to indicate hover state. It is OK for the View all option, as the underline disappears on hover. However for the other options, there is only a slightly darker grey background?
> 
> image

Ideally Id have an underline appear on hover, but our link style throughout the site disappears on hover. We could make all the menu items have a constant underline that disappears on hover to be consistent with link styles but when I tried this on the design it looked quite visually overwhelming (image attached - see what you think) 

I'd classed these items in the navigation as buttons, so when you click the buttons on the site they just have a slight colour change on hover so I'd aligned with this style. 

Screenshot 2024-05-09 at 14 02 00

jenhadfield-dfe:
Screenshot 2024-05-09 at 14 07 50
How about this for the hover state on the main nav items?

jenhadfield-dfe:
> **DESKTOP HOVER STATE**
> 
> Top level menu is missing a hover state @martyn-w ?
> 
> Production hover state image
> 
> New hover state (none) image

hiya, would it be possible to reinstate the hover state from the existing menu item and I will then look at making the menu styles more consistent with gov and other systems as discussed

martyn-w:
> > **DESKTOP HOVER STATE**
> > Top level menu is missing a hover state @martyn-w ?
> > Production hover state image
> > New hover state (none) image
> 
> hiya, would it be possible to reinstate the hover state from the existing menu item and I will then look at making the menu styles more consistent with gov and other systems as discussed

I have reinstated the hover state on the top level menu (desktop), please review and check it is now OK?

martyn-w:
> **FONT COLOUR**
> 
> I'm not sure whether this is a real issue or not @jenhadfield-dfe @martyn-w - I get it in the Chrome emulator but can't replicate on my phone (iPhone11/Safari)
> 
> The subcategory has white text on a pink background which fails colour contrast (and is also not on brand!)
> 
> However, on my phone it is black text on pink, which is fine
> 
> Screenshot from emulator: image

This issue should now be fixed in the latest version

gemmadallmandfe:
I've successfully retested the following issues:

* DESKTOP KEYBOARD ONLY - Tab order
* DESKTOP HOVER STATE
* FONT COLOUR

**Outstanding issues**

* DESKTOP HOVER STATE - EXPANDED MENU
Agreed with @martyn-w that we would go with the hover state as is and fix this under a new ticket once @jenhadfield-dfe has looked at the hover state across the whole site next week

app/components/header/navigation_component.rb
```diff
@@ -19,14 +20,130 @@ def all_resources
 
   private
 
-    def nav_link(link_text, link_path)
-      tag.li class: class_name(link_path) do
-        link_to_unless_current(link_text, link_path, class: "link--black link--no-underline") { tag.div(link_text) }
+    def sync_mode(mode)
+      mode == :mobile ? :desktop : :mobile
+    end
```
carlosmartinez, line 25: this seems slightly counterintuitive at first glance

carlosmartinez: maybe there's a way of expressing what `sync` means in this context

martyn-w: yep I think I could phrase that better 👍

app/views/pages/browse.html.erb
```diff
@@ -0,0 +1,15 @@
+<!-- The <ul> element is dynamically generated -->
+<!-- //NOSONAR -->
```
carlosmartinez, line 2: do we need this?

app/webpacker/controllers/navigation_controller.js
```diff
@@ -35,6 +22,179 @@ export default class extends Controller {
     }
   }
 
+  handleNavMenuClick(event) {
+    // fires when the user clicks on a navigation menu item
+    const item = event.target.closest('li');
+
+    if (item) {
```
carlosmartinez, line 29: suggest avoid nesting with `if (!item) return;`

app/webpacker/controllers/navigation_controller.js
```diff
@@ -35,6 +22,179 @@ export default class extends Controller {
...
+      const directLink = item.dataset.directLink === 'true';
+      const itemSync = this.getTargetItem(item.dataset.syncId);
+      const childMenu = this.getTargetItem(item.dataset.childMenuId);
+      const childMenuSync = this.getTargetItem(item.dataset.childMenuSyncId);
+      const toggleSecondaryNavigation = item.dataset.toggleSecondaryNavigation;
+
+      if (!directLink) {
```
carlosmartinez, line 36: again i think better to return rather than nest

app/webpacker/controllers/navigation_controller.js
```diff
@@ -35,6 +22,179 @@ export default class extends Controller {
...
+      if (nextSyncItem) {
+        nextSyncItem.focus();
+      }
+
+      event.preventDefault();
+    }
+  }
+
+  getTarget(id) {
+    if (id) {
```
carlosmartinez, line 81: ditto

app/webpacker/controllers/navigation_controller.js
```diff
@@ -35,6 +22,179 @@ export default class extends Controller {
...
+  }
+
+  getTargetItem(id) {
+    if (id) {
+      return this.getTarget(id).querySelector('#' + id);
+    }
+  }
+
+  toggleIconExpanded(item) {
+    if (item) {
```
carlosmartinez, line 97: `if (!item) return false;`

app/webpacker/controllers/navigation_controller.js
```diff
@@ -35,6 +22,179 @@ export default class extends Controller {
...
+      const icon = item.querySelector('span.nav-icon');
+      if (icon) {
```
carlosmartinez, line 99: `if (!icon) return false;`

app/webpacker/controllers/navigation_controller.js
```diff
@@ -35,6 +22,179 @@ export default class extends Controller {
...
+          icon.classList.add('nav-icon__contracted');
+          item.classList.remove('selected');
+          item.querySelector('a,button').ariaExpanded = false;
+          return true;
+        }
+        return false;
+      }
+    }
+  }
+
```
carlosmartinez, line 127: you could possibly factor out a shared method here, eg

```
toggleIcon(item, expand = true) {
  if (!item) return false;

  const icon = item.querySelector('span.nav-icon');
  const linkOrButton = item.querySelector('a, button');
  
  if (!icon || !linkOrButton) return false;

  const currentClass = expand ? 'nav-icon__contracted' : 'nav-icon__expanded';
  const newClass = expand ? 'nav-icon__expanded' : 'nav-icon__contracted';
  const ariaExpanded = expand ? true : false;

  if (icon.classList.contains(currentClass)) {
    icon.classList.replace(currentClass, newClass);
    item.classList.toggle('selected', expand);
    linkOrButton.ariaExpanded = ariaExpanded;
    return true;
  }

  return false;
}
```

which is used by

```
toggleIconExpanded(item) {
  return this.toggleIcon(item, true);
}

toggleIconContracted(item) {
  return this.toggleIcon(item, false);
}
```

app/webpacker/controllers/navigation_controller.js
```diff
@@ -35,6 +22,179 @@ export default class extends Controller {
...
+  hideMenu(menu) {
+    if (menu) {
+      if (!menu.classList.contains('hidden-menu')) {
+        menu.classList.add('hidden-menu');
+        return true;
+      } else {
+        return false;
+      }
+    }
+  }
```
carlosmartinez, line 148:
```suggestion
  toggleMenuVisibility(menu, show = true) {
    if (!menu) return false;

    const shouldHide = !show;
    const currentlyHidden = menu.classList.contains('hidden-menu');

    if (currentlyHidden === shouldHide) {
      menu.classList.toggle('hidden-menu', shouldHide);
      return true;
    }

    return false;
  }

  showMenu(menu) {
    return this.toggleMenuVisibility(menu, true);
  }

  hideMenu(menu) {
    return this.toggleMenuVisibility(menu, false);
  }
```

spec/components/header/navigation_component_spec.rb
```diff
@@ -28,4 +49,46 @@
   specify "extra resources are last" do
     expect(component.all_resources.map(&:title)).to end_with(extra_resources.values)
   end
+
+  context "when using a desktop browser" do
+    it "renders a dropdown menu for category links" do
+      expect(page).to have_css("#secondary-navigation > div.category-links > ol.category-links-list > li[id='page-five-category-1-desktop'] > button")
```
carlosmartinez, line 55: is there a less brittle and more readable way to test these things? eg

```
within('#secondary-navigation') do
  within('div.category-links') do
    expect(page).to have_button('whatever the button text is')
  end
end
```

carlosmartinez:
LGTM! Just a few suggestions

spec/components/header/navigation_component_spec.rb
martyn-w: can't seem to make this approach work? - the within blocks are skipped over

carlosmartinez (approved):

gemmadallmandfe (approved):
Looks good to me from a front-end perspective (apart from the known issues which are ticketed separately)

Happy for this to be merged first thing on Monday!

PR #3934, opened 2024-04-05
Added deep link to engineers teach physics courses on Find

MaxineCl:
### Trello card
https://trello.com/c/QxlvDOy6/2187-update-link-to-engineers-teach-physics-courses-on-subject-page

### Context
We currently have a link on the Engineers teach physics page https://getintoteaching.education.gov.uk/subjects/engineers-teach-physics that goes to the Find start page. We agreed that we should try deep linking directly to the Engineers teach physics courses as there are less than 20 in total.

Update the link on the Engineers teach physics page to go straight to the ETP courses search results

### DIFF
--- a/app/views/content/is-teaching-right-for-me/engineers-teach-physics/_article.html.erb
+++ b/app/views/content/is-teaching-right-for-me/engineers-teach-physics/_article.html.erb
@@ -63,7 +63,7 @@
      <p>You'll also be invited to an annual community day and have the opportunity to network with existing physics teachers.
      </p>
 
-    <p><a href="https://www.find-postgraduate-teacher-training.service.gov.uk/results?engineers_teach_physics=true&degree_required=two_two&applications_open=false&l=2&subjects%5B%5D=F3">Find engineers teach physics teacher training courses</a>.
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
Looks good to me but just need to check with Find team

app/views/content/is-teaching-right-for-me/engineers-teach-physics/_article.html.erb
```diff
@@ -63,7 +63,7 @@
      <p>You'll also be invited to an annual community day and have the opportunity to network with existing physics teachers.
      </p>
 
-    <p>You can <a href="https://www.find-postgraduate-teacher-training.service.gov.uk/">search for physics teacher training courses</a> and then filter by 'Engineers teach physics'.
+    <p><a href="https://www.find-postgraduate-teacher-training.service.gov.uk/results?engineers_teach_physics=true&degree_required=two_two&applications_open=false&l=2&subjects%5B%5D=F3">Find engineers teach physics teacher training courses</a>.
```
MylesJarvis, line 66: Only suggestion, and happy to be overruled on it is to add the filter for 'courses that are open', which would be this URL:
`https://www.find-postgraduate-teacher-training.service.gov.uk/results?engineers_teach_physics=true&study_type%5B%5D=full_time&study_type%5B%5D=part_time&qualification%5B%5D=qts&qualification%5B%5D=pgce_with_qts&qualification%5B%5D=pgce+pgde&degree_required=two_two&applications_open=false&applications_open=true&l=2&subjects%5B%5D=F3`

It uses this filter: 
image


This gives 16 results instead of the current 17. Is our aim to give users access to the full suite of courses, or to show them those that they can apply to at this specific moment? cc @gemmadallmandfe

gemmadallmandfe (approved):
Looks good to me - although the link could be simplified, this is the structure used by Find currently so makes sense to replicate it

This can be deployed first thing on Monday

PR #3938, opened 2024-04-09
Non-UK citizen update

laurahermionetennant92:
Non-UK citizen update


laurahermionetennant92:
Hi @MaxineCl this was relatively painless as Gemma decided we didn't need to change the url. I've also checked for other instances of non-UK teachers and they are all right for the context :)

MaxineCl (approved):

PR #3950, opened 2024-04-16
Wording changed on bursaries and scholarships

laurahermionetennant92:
Users prompted to check with their training provider about using bursaries and scholarships to offset fees

### DIFF
--- a/app/views/content/non-uk-teachers/fees-and-funding-for-non-uk-trainees.md
+++ b/app/views/content/non-uk-teachers/fees-and-funding-for-non-uk-trainees.md
@@ -120,7 +120,7 @@ You cannot receive both a bursary and a scholarship.
 
 Bursaries and scholarships are tax-free payments from the UK government that you do not have to pay back. They are paid to you by your teacher training provider. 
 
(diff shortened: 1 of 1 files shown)


app/views/content/non-uk-teachers/fees-and-funding-for-non-uk-trainees.md
```diff
@@ -120,7 +120,7 @@ You cannot receive both a bursary and a scholarship.
 
 Bursaries and scholarships are tax-free payments from the UK government that you do not have to pay back. They are paid to you by your teacher training provider. 
 
-If your training provider asks you to pay some or all of your teacher training fees in advance, you can sometimes use a bursary or scholarship to meet these costs. 
+If some or all of your teacher training fees are due in advance, your training provider may let you use a bursary or scholarship to meet these costs – contact them to check.
```
MaxineCl, line 123: Does this need commas? If some, or all, ?
Your **teacher** training provider do we need to add teacher in here?
Quite a long sentence. Is it worth splitting in two? You'll need to contact them to check.

PR #3959, opened 2024-04-30
Values and Variables

martyn-w:
### Trello card
[Implement the technical mechanism adding variables for all currency values on the GIT site through a single centralised config file](https://trello.com/c/D7IeBXu9/5927-implement-the-technical-mechanism-adding-variables-for-all-currency-values-on-the-git-site-through-a-single-centralised-config-f)

### Context
Provide a centralised store of important values/dates, which can be maintained in one place


ekumachidi (dismissed):

MylesJarvis:
My test did the following in a review app: 

image

This rendered as:
image

LGTM 👍

docs/content.md
```diff
@@ -287,6 +288,41 @@ hero_bg_color: white
...
+
+```yaml
+dates:
+  example:
+    opening: 1st September 2024
+
+dates_example_closing: 31/12/2024
+```
+
+These values can then be used in markdown files by referencing the value as `$``value_name``$`, e.g. `$``dates_example_opening``$` or `$``dates_example_closing``$`. Note that structured composite keys will be flattened to a single key.
```
MylesJarvis, line 303: @martyn-w curious about the use of single and then double backtick in this part of the docs, it renders like this:

image

I find that quite confusing as I wasn't sure whether to put the backticks in. This might be my poor understanding of markdown, but can we not write the variable without it?

MylesJarvis: I know it's then added in the markdown implementation below, just curious about the disparity.

martyn-w: Sorry @MylesJarvis my mistake, I'll fix this

martyn-w:
Needs a tech review before merging

mattonfoot (approved):
this seems to do the job, LGTM

PR #3983, opened 2024-05-17
Support values/variables in I18n YAML files

martyn-w:
### Trello card
[Values to populate the YML file of the funding widget](https://trello.com/c/lncSfSkm/5926-implement-functionality-for-our-single-centralised-config-file-of-currency-values-to-populate-the-yml-file-of-the-funding-widget)

### Context
The funding widget uses internationalisation (I18n) libraries to manage the subject content, stored in YAML files in `config/locales`.

This PR brings in the Values system into the I18n translations. A value can be specified using the syntax `%{value_name}` in the yaml file.


docs/content.md
```diff
@@ -321,8 +321,25 @@ An example ERB implementation might be:
...
+      maths:
+        name: "Maths"
+        group: "Secondary"
+        sub_head: "Maths - starting salary: %{salaries_example_min}"
+        funding: "The opening date for applications is %{dates_example_opening} so get ready to apply soon."
+```
+
 A list of the current values available on the site can be viewed at the `/values` endpoint.
 
+Values should be named using only the characters `A` to `Z`, the numbers `0` to `9`, and the underscore `_` character. Unsupported characters such as the hyphen `-` are converted into underscores.
```
MylesJarvis, line 341: @martyn-w, we spoke about formatting and rules earlier this week and thought that the best approach to take should be to use lowercase only. So I think we should make that clear here and not use uppercase letters in the example.

martyn-w: good spot Myles, now updated

spec/models/value_spec.rb
```diff
@@ -7,6 +7,52 @@
...
 
+  describe "##normalise_key (class method)" do
+    subject { described_class.normalise_key(key) }
+
+    context "when the key contains unsupported characters" do
+      let(:key) { "this-is!a{test}_key#123£567" }
+
+      it { is_expected.to be(:this_is_a_test__key_123_567) }
+
+      it { is_expected.to be_a Symbol }
```
carlosmartinez, line 18: redundant given the first assertion?

spec/models/value_spec.rb
```diff
@@ -7,6 +7,52 @@
...
+    end
+  end
+
+  describe "##get (class method)" do
+    subject { described_class.get(key) }
+
+    context "when the key contains hyphens" do
+      let(:key) { "dates-example-closing" }
+
+      it { is_expected.to eql("31/12/2024") }
```
carlosmartinez, line 28: where does this date come from, and will the test break next year?

carlosmartinez (approved):
left a couple of questions, but happy to pre-emptively approve on the assumption that you'll do whatever is the right thing!

spec/models/value_spec.rb
martyn-w: yes, technically so - I was trying to signify to a future dev that the keys must be symbols rather than strings. If its all the same, I think I'll leave this test in.

spec/models/value_spec.rb
martyn-w: the key is an example in an earlier PR and is used as part of the documentation, so shouldn't necessarily be updated next year.

PR #3989, opened 2024-05-21
Test variables with and without hyphens in YAML files

MylesJarvis:
### Context

Not to merge, this is a test of another branch with some new variables. This attempts to put them into the funding widget, which is populated via a YML file.


MylesJarvis:
This works in its current implementation:
image

However, what it leads us to is a usage of underscores in YML files but the ability to use hyphens in ERB files. Therefore, there's going to be a wider piece of work to convert variables across the site to remove hyphens and to replace them with underscores (it might be that we use double underscores to indicate differences between 'levels' defined in the YML file and where we are effectively using a hyphen in a part of the variable.

config/values/widget.yml
```diff
@@ -0,0 +1,3 @@
+widget:
+  subject:
+    art-and-design-bursary: £10,000
```
MylesJarvis, line 3: So in our prod implementation, this will be:
`art__and__design__bursary`

and that will also be reflected in the YML

config/locales/en.yml
```diff
@@ -85,7 +85,7 @@ en:
         name: "Art and design"
         group: "Secondary"
         sub_head: "Art and design - Secondary"
-        funding: "Bursaries of £10,000 are available for trainee art and design teachers if you’re eligible (non-UK citizens without indefinite leave to remain in the UK are unlikely to be eligible)."
+        funding: "Bursaries of %{widget_subject_art_and_design_bursary} are available for trainee art and design teachers if you’re eligible (non-UK citizens without indefinite leave to remain in the UK are unlikely to be eligible). Also inserting another variable as a test, our AO low value: %{fees_assessment_only_low} and one without any hyphens in, our low iQTS fee: %{fees_iQTS_low}."
```
MylesJarvis, line 88: Based on my other comment, this will be `%{widget_subject_art__and__design__bursary`

PR #4012, opened 2024-06-04
Update naming convention for variables already in prod

MylesJarvis:
### Trello card

N/A

### Context

We're updating the naming convention for how we define variables in the codebase, we're removing hyphens. This PR removes hyphens for those that are already implemented in producation.

No visual changes, but here are some screenshots of the values that have updated, and that they still render correctly:

These are from this page, the only page that uses this updated variable set https://get-into-teaching-app-review-4012.test.teacherservices.cloud/train-to-be-a-teacher/assessment-only-route-to-qts

image

image

### Guidance to review

Does this all look ok?


martyn-w (approved):
👍  LGTM

gemmadallmandfe (approved):
Looks good to me

PR #4033, opened 2024-06-14
Update link styles 

jenhadfield-dfe:
### Trello card
https://trello.com/c/0p7NvVfj/6017-review-link-hover-styles

### Context
The links on the Git website have an underline and on hover they disappear. This causes an issue for links that don’t need an underline eg. links in navigation items. It could also be confusing for someone using a screen magnifier who is hovering and might not realise a link is there.

[Gov.uk](http://gov.uk/) and services we point our users to, use a different style, where links are underlined and on hover the underline is bolder. For links without underline, the underline appears on hover.

updating links styles to include an thicker underline on hover

### DIFF
--- a/app/webpacker/styles/category.scss
+++ b/app/webpacker/styles/category.scss
@@ -64,7 +64,7 @@
       h3 {
         text-decoration: underline;
       }
-
+      
       &:focus h2 {
         color: $white;
       }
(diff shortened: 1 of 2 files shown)


gemmadallmandfe (approved):
Looks good to me

martyn-w (approved):
👍 LGTM

PR #4039, opened 2024-06-17
Test for monetary values

martyn-w:
### Trello card
[Identify monetary values](https://trello.com/c/2EHdqcwc/6114-implement-a-test-that-identifies-when-a-user-inserts-a-monetary-value-outside-of-our-centralised-configuration-file)

### Context
We need to identify where we use monetary values (e.g. £123,456) in our markdown and html files, so they can be moved to the values/variables system

New tests to scan markdown and html files for monetary values

### Guidance to review
NB: These tests _**will not pass**_ until the hard-coded values are removed. UPDATE: these tests are now passing.


martyn-w:
# DO NOT MERGE
these tests will not pass until the html/markdown files are corrected.

carlosmartinez:
The code looks fine, but is it a "feature test"? Do we want to continue running it on every build, or is it more of a one-off thing to find instances of monetary values in static files? If the latter, I think it would be better as a Rake task.

Or if we want to run the check on a specific cadence (eg every day or week) it could be a job, added to the crontab. Apply has some examples of this sort of thing. https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/app/workers/detect_invariants_daily_check.rb

martyn-w:
@carlosmartinez good questions; I think we definitely _do_ want to run this on every build (it's actually fairly quick to run) to immediately catch any new monetary values entering our content. I don't think a schedule will work well in this case: failing content could successfully be merged into the master branch if this test only ran on a daily/weekly schedule. And a rake task will be tricky for our content creators, who mainly rely on github actions to pickup issues rather than a local development environment.

We have similar content checks to verify URLs do not contain our hostname or "localhost" - these are defined as feature tests and run on build, and work effectively to alert the team when there are issues.

carlosmartinez:
> @carlosmartinez good questions; I think we definitely _do_ want to run this on every build (it's actually fairly quick to run) to immediately catch any new monetary values entering our content. I don't think a schedule will work well in this case: failing content could successfully be merged into the master branch if this test only ran on a daily/weekly schedule. And a rake task will be tricky for our content creators, who mainly rely on github actions to pickup issues rather than a local development environment.
> 
> We have similar content checks to verify URLs do not contain our hostname or "localhost" - these are defined as feature tests and run on build, and work effectively to alert the team when there are issues.

Thanks. That makes sense. I feel like they're not really feature tests, and should probably run as a separate build step, but if there's precedent then 👍 to proceed!

carlosmartinez (approved):

spencerldixon (approved):
LGTM 👍

PR #3546, opened 2023-09-20
Remove the feedback component from the website

garethwilliamsdesign:
### Trello card

[Trello 5049](https://trello.com/c/fjUSFOIS/5049-remove-the-feedback-component-from-the-website)

### Context

The current feedback component (located just above the footer), is simply a link to the feedback form and is duplicated in the footer. We have decided to remove the component but retain the footer link as this is a more typical location for such a feature.

### Changes proposed in this pull request

Removal of the feedback component across the website.

### Guidance to review

Check for the presence of the feedback component above the footer on various page layouts. If it's visible, further changes will be required.

PR #4046, opened 2024-06-25
Remove feedback component

martyn-w:
### Trello card
[Remove feedback bar](https://trello.com/c/Cv8uBCEe/6096-remove-feedback-bar-component-from-the-code)

### Context
Code clean-up of redundant code following PR https://github.com/DFE-Digital/get-into-teaching-app/pull/3546/

Remove old code

### Guidance to review
There should be no visible changes to the website


app/components/footer_component.html.erb
```diff
@@ -2,14 +2,6 @@
   <%= render Footer::TalkToUsComponent.new %>
 <% end %>
 
-<% if feedback? %>
-  <%= render Footer::FeedbackComponent.new %>
-<% end %>
-
-<% if mailing_list? %>
```
gemmadallmandfe, line 9: @martyn-w Just wanted to check why we are removing this mailing_list reference - is it not used?

app/components/footer_component.rb
```diff
@@ -1,20 +1,11 @@
...
 
   def talk_to_us?
     @talk_to_us
   end
-
-  def feedback?
-    @feedback
-  end
-
-  def mailing_list?
```
gemmadallmandfe, line 17: @martyn-w As per my other comment - why are we removing this as well? Is it not used?

app/components/footer_component.html.erb
martyn-w: Hi @gemmadallmandfe, the mailing list component in the footer was removed in March 2023 (see PR https://github.com/DFE-Digital/get-into-teaching-app/pull/3196) - however this little piece of code was accidentally missed in that PR. Fortunately the logic continued to work correctly as the default behaviour in ruby is to treat an uninitialised `@instance` variable as null, rather than throw an error.

Removing this code just cleans up an earlier oversight; it should have no effect on the website.

app/components/footer_component.rb
martyn-w: it is not used and can be safely removed.

asatwal (approved):
👍 LGTM

PR #4062, opened 2024-07-04
Content updates to event information page / prevent 2d scrolling

martyn-w:
reduce possibility for 2d scrolling

### Trello card
[Amend design of provider event details to prevent horizontal scrolling](https://trello.com/c/urOtiYGF/6115-amend-design-of-provider-event-details-to-prevent-horizontal-scrolling-two-dimensions)

### Context
In some cases, long URLs to event provider websites can cause the event details pages scroll in two dimensions, which is an accessibility issue.

Tweak design and content to reduce chance of 2D scrolling.

### Guidance to review
Test with a few different event types


gemmadallmandfe:
@martyn-w Do you have a test where a provider has chosen to have an email link for registration rather than a website? We should just test that scenario before this is deployed

cc @jenhadfield-dfe

gemmadallmandfe (approved):
Looks OK from a front-end perspective - further changes to follow via a separate ticket

asatwal (approved):
👍  LGTM

PR #4081, opened 2024-07-22
Delete visa minimum savings variables

gemmadallmandfe:
### Trello card
https://trello.com/c/dQdKBnmR

### Context
We are centralising our financial values into one place for easier maintenance. This PR removes previously created variables related to minimum savings for different visa types, as our visa guidance has recently been simplified to remove this information.

### Guidance to review
Check that these variables aren't in use anywhere.


Sarah-DfE (approved):
Looks good to me @gemmadallmandfe

PR #4085, opened 2024-07-23
Implement visa savings threshold variables

gemmadallmandfe:
### Trello card
https://trello.com/c/E0dM5tZa

### Context
We are centralising our financial values into one place for easier maintenance. This PR moves the visa savings threshold variables from the salaries value file to the nonuk values file, and adds the variables to the two visa pages


Sarah-DfE (approved):
Looks good to me!

PR #4102, opened 2024-07-25
Install and configure sidekiq

spencerldixon:
### Trello card

https://trello.com/c/pxanTErV/6329-install-queue-processor

### Context

We need to be able to run background jobs within the application

- Install sidekiq
- Adds a `worker` and `redis` to the Procfiles for easier development
- Updates the documentation to reflect this
- Configures a `dfe_analytics` queue to ringfence analytics to their own queue
- Creates a `config/sidekiq.yml` for additional configuration in one place
- Sets some sensible defaults for `ActiveJob` retries, discard on not finding records etc
- Adds `dump.rdb` to the gitignore to not pick up redis dumps in your commits

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -127,6 +127,7 @@ Rails.application.configure do
   config.rails_semantic_logger.processing = false
   config.rails_semantic_logger.format = :json
   config.rails_semantic_logger.add_file_appender = false
+  config.rails_semantic_logger.filter = proc { |log| log.name != "DfE::Analytics::SendEvents" }
   config.semantic_logger.add_appender \
     io: $stdout,
     level: Rails.application.config.log_level,


terraform/aks/config/production.tfvars.json
```diff
@@ -15,5 +15,7 @@
       },
     "external_url": "https://getintoteaching.education.gov.uk/healthcheck.json",
     "enable_logit": true,
-    "enable_prometheus_monitoring": true
+    "enable_prometheus_monitoring": true,
+    "sidekiq_replicas" : 1,
+    "sidekiq_memory_max" : "1Gi"
```
spencerldixon, line 20: I've set this as 1 for now, but we could bump it up if needed

terraform/aks/application.tf
martyn-w, line 1: The worker definition in `terraform/aks/application.tf` needs a little updating to ensure that worker pods include the PR number.

From @saliceti :
> I think the issue is with environment = local.app_name_suffix . app_name_suffix  is not defined, it should have errored, not sure why
> the webapp has environment = local.environment
> ah it was set with app_name_suffix = var.app_name == null ? var.environment : var.app_name but var.app_name defaults to null
> confusingly, local.environment is different than var.environment here

martyn-w (approved):
👍  LGTM

@saliceti please could you review the Terraform scripts too?

martyn-w (approved):
👍  LGTM

PR #4106, opened 2024-07-31
Southfields,  0017, 0024, stories-zainab jpg deleted from images.yml

mariosh-e:
### Trello card
https://trello.com/c/UD9VZ0FI

### Context
Carry out a final check of expired images
Running a final check to make sure that the only images now in GitHub are images for which we have a current licence


martyn-w (changes requested):
There are a. number of failing spec tests which need to be fixed before this can be merged. I'll try and fix them now...

martyn-w (approved):
👍  LGTM

PR #4108, opened 2024-07-31
Replace rack-page_caching with actionpack-page_caching

spencerldixon:
### Trello card

https://trello.com/c/wpcleikM/6326-update-replace-rack-pagecaching-library

### Context

The `rack-page_caching` gem that we use hasn't been touched in 10 years and is the source of multiple `no implicit conversion of nil into String` errors being reported on production

- Remove the `rack-page_caching` gem
- Replace with `actionpack-page_caching` gem
- Refactor the middleware that filters out dynamic pages from being cached
- Adjust tests to handle new gem

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -85,8 +85,8 @@ gem "view_component", "~> 3.10.0"
 
 gem "google-api-client", ">= 0.53.0", require: false
 
+gem "actionpack-page_caching"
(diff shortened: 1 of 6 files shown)


gemmadallmandfe:
@spencerldixon Can we close this PR for now and re-open it when we pick this up again?

PR #4127, opened 2024-08-08
Fix codespaces not working since introduction of redis dependency

spencerldixon:
### Trello card

https://trello.com/c/AJTpvoFL/6362-create-a-devcontainer-environment-for-git-in-vscode?filter=member:spencerldixon

### Context

Introducing sidekiq has broken codespaces for developers using them as there is now a redis dependency.

- Update devcontainer config to include redis 6.0.14 (same as production)
- Modernise docker compose file


martyn-w (approved):
👍 LGTM

PR #4134, opened 2024-08-14
Update guidance on non-UK components

emmaachesongray:
### Trello card
https://trello.com/c/kfPvMtUp/6310-add-content-guidance-re-non-uk-inset-and-expander-components

### Context
We've recently introduced new non-UK instet and expander components, so need to update our guidance on how and when to use them.


emmaachesongray:
@jenhadfield-dfe this should be all ready for you :)

jenhadfield-dfe (approved):

PR #4152, opened 2024-08-23
Life as a teacher footer navigation component

jenhadfield-dfe:
### Trello card
https://trello.com/c/Jd9HvLCS/6406-phase-5-repurpose-navigation-component-for-life-as-a-teacher-section

### Context
Repurpose the navigation component used on the homepage to be used throughout the pages within the Life as a teacher section

Ensure that the component is updatable - eg. It’s the same on all pages and we only have to update it in once place, rather than on each page

Duplicated the directory file used on the homepage and added to the life as a teacher section. updated pages with the component in the section


GayeCord:
@jenhadfield-dfe can this PR be closed now?

PR #4156, opened 2024-08-28
Improve regex for financial variables check

spencerldixon:
### Trello card

https://trello.com/c/7WVoxjci/6486-extend-test-checking-for-financial-values-to-cover-other-formats?filter=member:spencerldixon

### Context

We recently replaced all hardcoded financial values on the site with variables controlled in a set of config files, to make future updates easier. We have a test that checks for £ used outside these config files, however, the test does not cover other formats.

- Adds html entities and codes for $, £, and € in `MonetaryChecker::MONEY_REGEX`


gemmadallmandfe (commented):
Looks OK to me from a front-end perspective (correct variable in salaries banner) @spencerldixon @sarahcrack

martyn-w (approved):
👍  LGTM

PR #4158, opened 2024-08-28
Degree subject autocomplete

martyn-w:
### Trello card
[Degree subject autocomplete in adviser flow](https://trello.com/c/IEZ1yrxP/6412-implement-the-degree-subject-autocomplete-git-adviser-flow-happy-path-with-javascript)

Allow a wide selection of degree subject choices by using the DfE reference data gem for degree subjects, coupled with an auto-complete box which allows the user to type in any input

### Guidance to review
Requires javascript: if this is disabled it will default to a `<select>` list


gemmadallmandfe:
We need to change the hint text now that degree subject uses auto-complete rather than presenting the list of ITT subjects

image

@Sarah-DfE Can you help with this please?

For reference, this is what is used on Apply for this question (but we might want to put the select part before the type your own if we use this?)

image

Sarah-DfE:
@martyn-w Just to check, the degree subjects don't appear until someone starts typing?

@gemmadallmandfe wonder if adding not to include BA etc might make it more confusing. Was thinking of keeping it super simple, so something like 'Enter your degree subject or select from the suggestions. '

martyn-w:
hi @Sarah-DfE

> @martyn-w Just to check, the degree subjects don't appear until someone starts typing?

yes, that's correct 👍

Sarah-DfE:
Thanks @martyn-w 

@gemmadallmandfe In that case, we could have 'Type to enter your degree subject or select from the suggestions.'

This way people should know that they have to start typing to see the list.

spencerldixon (approved):
LGTM! 👍

gemmadallmandfe (approved):
Looks good to me - happy for this to be merged on Monday

PR #4159, opened 2024-08-29
Webchat sourcecode

martyn-w:
### Trello card
[Test webchat widget](https://trello.com/c/C9Mu6qKX/6449-test-webchat-widget)

### Context
Test that the webchat widget runs correctly

Changes include:
* latest version of chat client (8th October, `index-77d633bb.js`)
* CSP security headers
* caching on chat availability API to reduce client requests
* increase polling interval on chat availability api to reduce client requests
* Environmental variables to configure chat targets

### Guidance to review

Test in multiple browsers, mobiles and tablets


spencerldixon (dismissed):
LGTM! 👍

gemmadallmandfe:
@martyn-w I've tested using keyboard only and that works well

However, when I tested using VoiceOver on iPhone 11, the chat widget opened successfully but focus remained on the GIT website page, which isn't right?

Also, I think we need to indicate to screen reader users that the chat opens in a new window
We do this elsewhere (eg bank holiday link in Talk to us component) with
image

gemmadallmandfe:
Retested on iPhone11 / iOS 17.6.1 and screen reader focus moves to the chat window when opened now

spencerldixon (dismissed):
LGTM 👍

spencerldixon (dismissed):
LGTM 👍

spencerldixon (approved):
LGTM 👍

PR #4171, opened 2024-09-02
Ensure that 2024 is removed from the list of teacher training years in adviser funnel

sarahcrack:
### Trello card

https://trello.com/c/N9d5ExNj

### Context

The adviser funnel asks users when they want to start their teacher training. Once the recruitment cycle is closed in September, we need to ensure that the list does not contain the current year (2024).

Updated date to 17 Sept and fixed tests and typos in test names.

### DIFF
--- a/spec/models/teacher_training_adviser/steps/start_teacher_training_spec.rb
+++ b/spec/models/teacher_training_adviser/steps/start_teacher_training_spec.rb
@@ -82,9 +82,9 @@ RSpec.describe TeacherTrainingAdviser::Steps::StartTeacherTraining do
       end
     end
 
-    context "when its after 3th September of the current year (2022)" do
+    context "when its after 17th September of the current year (2022)" do
       around do |example|
-        travel_to(Date.new(2022, 9, 4)) { example.run }
+        travel_to(Date.new(2022, 9, 17)) { example.run }
       end
 
       it "returns 'Not sure', and the next 3 years" do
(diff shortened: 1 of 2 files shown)


spencerldixon (approved):
LGTM 👍

spec/models/teacher_training_adviser/steps/start_teacher_training_spec.rb
```diff
@@ -157,59 +157,59 @@
       wizardstore["degree_options"] = TeacherTrainingAdviser::Steps::HaveADegree::DEGREE_OPTIONS[:studying]
     end
 
-    it "returns current calendar year + 2 years if first year and before 4th September" do
+    it "returns current calendar year + 2 years if first year ends before 17th September" do
```
gemmadallmandfe, line 160: @sarahcrack @spencerldixon I don't think this was a typo?

ie `first year` refers to the degree status option chosen (similarly `second year`, `final year`, `other` and doesn't refer to the date?

spencerldixon: ah that makes more sense, it just didn't read right while we were working on them. Yes let's change them back

sarahcrack: @gemmadallmandfe @spencerldixon I have amended the test wording and raised a new PR: https://github.com/DFE-Digital/get-into-teaching-app/pull/4172.

PR #4173, opened 2024-09-02
Add test to check for orphan yaml variables in the content

spencerldixon:
### Trello card

https://trello.com/c/oxjKRhKw/6385-create-a-test-to-verify-any-variable-that-is-entered-on-the-site-also-exists-in-a-yml-file?filter=member:spencerldixon

### Context

As part of our approach to implementing a centralised set of files for surfacing numeric values across the site, we want to make the system more foolproof/less prone to potential errors
This means that we need to be able to check the codebase for mistakes that team members may have made when creating new pages.
Specifically, introducing a variable that is not present in any of our .yml files where we define these variables
This is to protect against typos and to act as a second line of defence against human error when reviewing.

- Adds a test that looks for `$thing$`, `%{thing}`, `<%= v :thing %>` and `#{v :thing}` style variables in the front end and checks them against the keys within any file in `config/values/*.yml`.
- Raises a failure for each instance of a yaml key that could not be found in the front end with the key and file path.


spencerldixon:
@martyn-w @gemmadallmandfe right now, this test scans for variables in the content, and values in `config/values/*.yml`, and does a difference of the two (yaml - content). This returns all yaml variables, excluding any used in content, which leaves only the once that aren't used.

Creating a check for variable in the content that aren't listed in the yaml is the same but backwards. We'd do a test for `content - yaml` which would return all content variables, excluding any listed yaml variables, leaving values in the content which don't appear in the yaml files.

The problem with this is we've also got some instances where we set up these values in the front matter, then use them in the same way we use yaml variables in the code. See [https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/views/content/blog/abigails-career-progression-story.md?plain=1#L5](https://github.com/DFE-Digital/get-into-teaching-app/blob/master/app/views/content/blog/abigails-career-progression-story.md?plain=1#L5) where we set this variable up on line 5 then use it on line 27.

This means that `content - yaml` returns a bunch of variable like the above, which are valid, but don't have a corresponding yaml and I'm not sure the best way of differentiating these. Open to suggestions.

martyn-w (commented):
This is an excellent approach to solving the problem "are there any defined variables which are not being used in the content?"

However, it doesn't solve the problem "is there any content which is attempting to use an undefined variable?"

That second question is indeed a tricky one! As you point out, sometimes `$something$` is a variable, and sometimes it is a call to a component defined elsewhere. 

I think to solve this, we will need to parse the frontmatter of the markdown files and identify calls to different component types. And then for each call to a component type, pick out the keys and store these in a list of `component_keys`.

See the file `lib/template_handlers/markdown.rb` for parsing the markdown and identifying components. There are currently six component types to identify which could have their own child component keys: `quote, quote_list, inset_text, youtube_video, steps, expander`. (NB: `quote_list` is not currently used.)

(We don't need to worry about parsing ERB files for component keys because they simply call the component directly, without a key.)

Then we can do:

```
undefined_variables = content_variables_values - yaml_variable_values - component_keys
```

...and throw an error if `undefined_variables` is not empty.

---

One other small point: in ERB files, `<%= v :bla %>` is shorthand for `<%= value :bla %>` so make sure to search for both `v` and `value`.

spec/views/identify_orphan_variables_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      content = File.open(file).read
+      front_matter = FrontMatterParser::Parser.new(:md).call(content).front_matter
+
+      COMPONENT_TYPES.map { |type| front_matter[type] }.compact.map(&:keys).flatten
+    end
+  end
+
+  let!(:orphan_frontend_variables) do
+    orphan_frontend_variables = content_files_and_values.transform_values { |values|
+      values.reject { |value| (yaml_files_and_values.values + component_files_and_keys.values).flatten.include?(value) }
```
martyn-w, line 56: might be more efficient to store:
`(yaml_files_and_values.values + component_files_and_keys.values).flatten` in a `let(:var)` rather than re-evaluate for every `|value|`

martyn-w (dismissed):
👍  very impressive 👍 

have made a small suggestion about adding another let(:var) , but I think this definitely solves the problem

martyn-w (approved):
👍  LGTM

PR #4182, opened 2024-09-09
Teaching as a career subcategory page

emmaachesongray:
### Trello card
https://trello.com/c/2XrUvUmY/6527


app/views/content/life-as-a-teacher/teaching-as-a-career/_categories.html.erb
```diff
@@ -0,0 +1,29 @@
+<div class="row">
+  <section class="category col-space-m">
+    <section class="container category__cards main-section">
+      <nav class="category__nav-cards">
+        <ul>
+          <%= render Categories::CardComponent.new(heading_tag: "h2", card:
+            OpenStruct.new(
+              title: "Change to a career in teaching",
```
gemmadallmandfe, line 8: I wonder about the order of this?

ECT info is relevant to everyone - career changers info (which is a bit weak currently) is not

So I wonder if it makes more sense to be ECT / Insights / Career changers for now?

(Appreciate that there will be new content to come which would then go at the top of the page!)

app/views/content/life-as-a-teacher/teaching-as-a-career/_categories.html.erb
```diff
@@ -0,0 +1,29 @@
...
+              path: "/life-as-a-teacher/teaching-as-a-career/change-careers",
+              image: "static/images/content/hero-images/0029.jpg"
+            )) %>
+          <%= render Categories::CardComponent.new(heading_tag: "h2", card:
+            OpenStruct.new(
+              title: "Early career teachers (ECTs)",
+              description: "Find out about the support early career teachers get when they start teaching, including mentor support and training based on the early career framework (ECF).",
+              path: "/life-as-a-teacher/teaching-as-a-career/early-career-teachers"
+            )) %>
+          <%= render Categories::CardComponent.new(heading_tag: "h2", card:
```
gemmadallmandfe, line 19: Presume we will look to move this to an inspiration component (will this be covered on a separate ticket?)?

app/views/content/life-as-a-teacher/_categories.html.erb
```diff
@@ -11,6 +11,15 @@
               image: "static/images/content/hero-images/teacher.jpg"
             )) %>
         </ul>
+        <ul>
+          <%= render Categories::CardComponent.new(heading_tag: "h2", card:
+            OpenStruct.new(
+              title: "Teaching as a career",
+              description: "Learn about the career progression in teaching, and the support available for early career teachers and career changers.",
+              path: "/life-as-a-teacher/teaching-as-a-career",
+              image: "static/images/content/hero-images/teacher3.jpg"
```
gemmadallmandfe, line 20: I love this image but we have the same teacher twice (out of three images) on the Life as a teacher page now - should we use a different image instead @jenhadfield-dfe ?

app/views/content/life-as-a-teacher/teaching-as-a-career.md
```diff
@@ -0,0 +1,15 @@
+---
+title: "Teaching as a career"
+heading: "Teaching as a career"
+category: Life as a teacher
```
gemmadallmandfe, line 4: The `Pay and benefits` page and the `Explore subjects` page both have an image in the header (as a result of this PR) and then don't have images on the cards

The new `Teaching as a career` page doesn't have an image in the header but does have images on the top card

Is this deliberate? Should they be consistent? @jenhadfield-dfe

gemmadallmandfe (commented):
Looks great @emmaachesongray @jenhadfield-dfe 

Just a couple of questions - and then it is ready to go out :-)

app/views/content/life-as-a-teacher/teaching-as-a-career/_categories.html.erb
emmaachesongray: Good point, I'm not sure how we use the inspiration component so would rather we tackle that under a different ticket.

app/views/content/life-as-a-teacher/teaching-as-a-career.md
emmaachesongray: I deliberately used the image in the header for Pay and Benefits as when you only have 2 cards, it looks strange with one with an image and one without. In this PR we're moving the career ladder page from P+B so it drops it from 3 nav cards to 2. 

On the subjects page, we decided when doing the PR that it looked strange to have one subject in focus with an image but not the others - we felt they should all be equally weighted. 

I wonder if it now looks strange to not be consistent across all subcategory pages? @jenhadfield-dfe what do you think?

gemmadallmandfe: ah that makes sense - but I think we should also consider that more users will be on mobile, so we shouldn't necessarily make decisions just looking at desktop?

app/views/content/life-as-a-teacher/_categories.html.erb
jenhadfield-dfe: updated the image to a different one

gemmadallmandfe (approved):
Looks good to me

PR #4193, opened 2024-09-12
Remove Edge Hill and Bristol from iQTS page

gemmadallmandfe:
### Trello card
https://trello.com/c/mKEqavjz

### Context
We need to update the iQTS page to remove Edge Hill and Bristol from the list of providers for the time being


Sarah-DfE (approved):
Looks good to me

PR #4245, opened 2024-10-14
Delete research page and associated PDFs

gemmadallmandfe:
### Trello card
https://trello.com/c/CgH42dXe

### Context
We are removing the PR page and all PDF research reports as they get little traffic and the PDFs are not accessible.


config/routes.rb
```diff
@@ -17,10 +17,11 @@
 
   YAML.load_file(Rails.root.join("config/redirects.yml")).fetch("redirects").tap do |redirect_rules|
     redirect_rules.each do |from, to|
+      to_mapped = to.gsub("/packs/", Rails.env.test? ? "packs-test" : "packs")
```
martyn-w, line 20: just to make the substitution a bit tighter, how about
```
to_mapped = to.gsub(/^\/packs\//, Rails.env.test? ? "/packs-test/" : "/packs/")
```

martyn-w (approved):
👍  LGTM

PR #4261, opened 2024-10-18
Fix heading overlap

spencerldixon:
### Trello card

https://trello.com/c/wo4ao4Nj/6765-fix-overlapping-headings-on-inspiration-navigation-cards?filter=member:spencerldixon

### Context

Headings on inspiration navigation cards that wrap onto two lines are overlapping.
We need to fix the styling so that the words aren’t overlapping.

I find the `highlight-text` mixin isn't working properly and causes multilines of text to overlap.

- Amend the `highlight-text` mixin to apply styles to a second `span` tag that is set to `relative` to create a new stacking context where the text sits on top of the previous span, used as the background.
- Allows existing titles to remain unchanged, but to drop in the new `span` wherever we encounter overlap

### DIFF
--- a/app/components/content/inspiration_card_component.html.erb
+++ b/app/components/content/inspiration_card_component.html.erb
@@ -9,7 +9,9 @@
         <div class="category__nav-card__wrapper__col1">
           <div class="category__nav-card--content">
             <%= content_tag heading_tag do %>
-              <%= content_tag :span, title %>
+              <%= content_tag :span do %>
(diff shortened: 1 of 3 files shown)


gemmadallmandfe (approved):
Looks good to me

Have tagged @jenhadfield-dfe for a review before this is merged

jenhadfield-dfe (approved):

martyn-w (approved):
👍 LGTM

PR #4262, opened 2024-10-18
Update link opening style

sarahcrack:
### Trello card

https://trello.com/c/aGCWN9KR/6618-update-link-opening-style-so-that-links-open-in-the-same-tab-by-default

### Context

We’ve developed some standards for how links work on our site: internal and external links will only open in a new tab if the user would lose information such as filling out a form eg. Privacy policy. We need to make this change site wide - all links open in the same tab by default

- Deleted or refactored relevant code in controllers and transformers that relate to opening styles of links. 
- Deleted references to `target="_blank"` and `<span class="visually-hidden">(opens in new window)</span>` where appropriate/ relevant to links on pages.
- Removed attributes from new footer so links don’t open in new tab.
- Changed the content of the chat button link so it currently displays _'Open chat in new tab'_ (see attached screenshot).
- Created the `NewTabLinks` class which will append _(opens in new tab)_ to anchor links with a class of `new-tab`, unless they also have a class of button. All anchor links with `new-tab` class will have `target="_blank" rel="noopener"` added.
- Created tests for `NewTabLinks` and modified chat component test to reflect new wording of link.

### Guidance to review
- Spot check of links to ensure they open in the same tab. 
- Check whether any further links need the `new-tab` class (e.g. links within forms). 

Screenshot 2024-10-31 at 16 33 14


config/locales/en.yml
```diff
@@ -857,9 +857,9 @@ en:
       teacher_training_adviser_steps_overseas_country:
         country_id_html: |-
           You can apply to train or teach from any country in the world, but you may need a visa. Our guidance for
-          <a target="_blank" href="/non-uk-teachers/teach-in-england-if-you-trained-overseas">international teachers</a>
-          and <a target="_blank" href="/non-uk-teachers/train-to-teach-in-england-as-an-international-student">international
-          trainees</a> explains how to apply for the correct visa or immigration status for your employment or study in the UK.
+          <a href="/non-uk-teachers/teach-in-england-if-you-trained-overseas">international teachers</a>
+          and href="/non-uk-teachers/train-to-teach-in-england-as-an-international-student">international
```
martyn-w, line 861: html seems corrupted here, should be `<a href="bla">bla</a>`

martyn-w:
In the Get an Adviser section, there are some links which do not open in a new tab, e.g.
https://get-into-teaching-app-review-4262.test.teacherservices.cloud/teacher-training-adviser/sign_up/returning_teacher

click on What is QTS? > click any link in the expanded section

martyn-w:
Hi @sarahcrack on the "Which country do you live in?" page of Get An Adviser `/teacher-training-adviser/sign_up/overseas_country` there are a couple of links which I think should open in a new window:

* international teachers
* international trainees

martyn-w (dismissed):
👍 LGTM from tech perspective - please get a review for content

app/components/footer_component.html.erb
```diff
@@ -66,42 +60,34 @@
...
-
       <hr class="site-footer__divider">
     </section>
-
     <div class="site-footer__social-icons">
       <h2 class="visually-hidden">Support links</h2>
-      <a href="https://www.facebook.com/getintoteaching" class="site-footer-top__social__link" target="_blank" rel="noopener">
+      <a href="https://www.facebook.com/getintoteaching" class="site-footer-top__social__link">
         <%= helpers.fab_icon("facebook-f", "icon") %>
-        <span class="visually-hidden">Facebook - opens in a new tab</span>
```
gemmadallmandfe, line 77: @sarahcrack I think we need to keep the visually hidden line for each of the social media icons with the social media site name, but without the `opens in new tab`

If we don't have this visually hidden information, there is no way for screen reader users to understand what these links are

ie

*  `<span class="visually-hidden">Facebook</span>`
*  `<span class="visually-hidden">Instagram</span>`
*  `<span class="visually-hidden">LinkedIn</span>`
*  `<span class="visually-hidden">X, formerly known as Twitter</span>`
*  `<span class="visually-hidden">YouTube</span>`

sarahcrack: thanks @gemmadallmandfe! I am going to test with aria labels initially as I think this should be sufficient for screen readers if the links are opening in the same tab. I will add/test with visually-hidden text if aria labels don't work as expected. 

Facebook example with aria label:
` <a href="https://www.facebook.com/getintoteaching" class="site-footer-top__social__link" aria-label="Facebook"><%= helpers.fab_icon("facebook-f", "icon") %></a>`

gemmadallmandfe: Tested successfully using iOS/VoiceOver :-)

gemmadallmandfe (approved):
Looks good to me from a front-end perspective - happy for this to be merged after a final tech review

PR #4275, opened 2024-10-24
Remove overseas qualification checks from Get the most from an event page

gemmadallmandfe:
### Trello card
https://trello.com/c/u1PedLiF

### Context
Overseas qualification checks are no longer offered at GIT events so we need to remove the reference to this on the Get the most from an event page.


config/images.yml
```diff
@@ -466,6 +466,9 @@
 "static/images/content/events/teaching-event.jpeg":
   # A man smiling wearing a blue Get Into Teaching sweatshirt talking to a lady with banner behind them.
   alt: ""
+"static/images/content/events/event-image.jpg":
+  # a woman talking to a woman holding a coffee cup with an event banner in the background.
```
emmaachesongray, line 470:
```suggestion
  # A woman talking to another woman while holding a coffee cup with an event banner in the background.
```

emmaachesongray (approved):
Looks good to me

PR #4282, opened 2024-10-25
Delete grasp every opportunity blog

emmaachesongray:
### Trello card
https://trello.com/c/yBHP8oup/6782-delete-and-redirect-blog-post-grasp-every-opportunity


gemmadallmandfe:
@emmaachesongray This PR has failed as the blog you are deleting is referenced in a spec test:
https://github.com/DFE-Digital/get-into-teaching-app/blob/master/spec/features/blog_spec.rb

The reference will need replacing with another blog name (one you are not deleting for a while!) and then worth just getting a tech review on this

gemmadallmandfe (approved):
Looks good to me from a front end perspective - would suggest getting a tech review before merging though!

spencerldixon (approved):
LGTM 👍

PR #4293, opened 2024-10-29
Improve QuoteWithImage component for use in markdown files

spencerldixon:
### Trello card

https://trello.com/c/kO92SK1i/6849-tweak-quote-component-so-it-can-be-used-in-the-middle-of-markdown-pages?filter=member:spencerldixon

### Context

We need to tweak the quote component (if possible!) so that it is more flexible and we can use it in the middle of markdown pages when needed

- Improves the mobile responsiveness of the `QuoteWithImageComponent` by removing fixed widths and allowing it to expand and contract more gracefully.
- Adds `QuoteWithImageComponent` to the markdown parser to allow it to be utilised within markdown files
- Aligns `QuoteWithImageComponent` content with the title at various breakpoints


jenhadfield-dfe:
Screenshot 2024-11-04 at 13 53 27
The component doesn't work well in the markdown file when it includes an image - as the image is set to 50% width, its not picked up that at this size it should shrink down/go above

jenhadfield-dfe (approved):

PR #4295, opened 2024-10-29
Ignore migration exceptions

martyn-w:
### Trello card
[Improve pod stability during deployment](https://trello.com/c/l3CnCqqD/6852-improve-pod-stability-during-deployment-process)

### Context
pods occasionaly restart during boot because the migrations run on multiple pods - but in fact we don't have any migrations to run

This ignores migration exceptions when booting pods, reducing restarts and improving stability


spencerldixon (approved):
LGTM 👍

martyn-w:
This PR seems to break the review deployment process

martyn-w:
Closing for a better solution to this problem

PR #4315, opened 2024-11-01
Run migrations prior to booting app and worker

martyn-w:
### Trello card
[Improve pod stability](https://trello.com/c/l3CnCqqD/6852-improve-pod-stability-during-deployment-process)

### Context
On deployment to production, pods can crash if migrations are run on multiple pods at the same time

This PR is based on https://github.com/DFE-Digital/itt-mentor-services/pull/893/ . It ensures migrations run as a kubernetes job prior to booting the rails app/worker

### DIFF
--- a/terraform/aks/application.tf
+++ b/terraform/aks/application.tf
@@ -33,7 +33,6 @@ module "application_configuration" {
 
 # Run database migrations
 # https://guides.rubyonrails.org/active_record_migrations.html#preparing-the-database
-# https://github.com/ilyakatz/data-migrate
 # Terraform waits for this to complete before starting web_application and worker_application
 resource "kubernetes_job" "migrations" {
   metadata {
--- a/terraform/aks/application.tf
+++ b/terraform/aks/application.tf
@@ -31,8 +31,88 @@ module "application_configuration" {
   }
 }
 
+# Run database migrations
+# https://guides.rubyonrails.org/active_record_migrations.html#preparing-the-database
+# https://github.com/ilyakatz/data-migrate
+# Terraform waits for this to complete before starting web_application and worker_application
+resource "kubernetes_job" "migrations" {
+  metadata {
+    name      = "${var.service_name}-${var.environment}-migrations"
+    namespace = var.namespace
+  }
+
+  spec {
+    template {
+      metadata {
+        labels = { app = "${var.service_name}-${var.environment}-migrations" }
+        annotations = {
+          "logit.io/send"        = "true"
+          "fluentbit.io/exclude" = "true"
+        }
+      }
+
+      spec {
+        container {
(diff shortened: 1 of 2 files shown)


spencerldixon (dismissed):
LGTM 👍

terraform/aks/application.tf
```diff
@@ -31,8 +31,88 @@ module "application_configuration" {
   }
 }
 
+# Run database migrations
+# https://guides.rubyonrails.org/active_record_migrations.html#preparing-the-database
+# https://github.com/ilyakatz/data-migrate
+# Terraform waits for this to complete before starting web_application and worker_application
+resource "kubernetes_job" "migrations" {
+  metadata {
+    name      = "${var.service_name}-${var.environment}-migrations"
```
RMcVelia, line 40: the name needs to include the pr-number for review apps, otherwise they will have the same name i.e. "get-into-teaching-app-review-migrations"

terraform/aks/application.tf
```diff
@@ -31,8 +31,88 @@ module "application_configuration" {
...
+    namespace = var.namespace
+  }
+
+  spec {
+    template {
+      metadata {
+        labels = { app = "${var.service_name}-${var.environment}-migrations" }
```
RMcVelia, line 47: same as above regarding review apps

terraform/aks/application.tf
```diff
@@ -31,8 +31,88 @@ module "application_configuration" {
   }
 }
 
+# Run database migrations
+# https://guides.rubyonrails.org/active_record_migrations.html#preparing-the-database
+# https://github.com/ilyakatz/data-migrate
+# Terraform waits for this to complete before starting web_application and worker_application
```
ollietreend, line 37: ### TL;DR
No need to include a link to the [data_migrate](https://github.com/ilyakatz/data-migrate) gem here – this app doesn't use it, so it's not relevant.

```suggestion
# Run database migrations
# https://guides.rubyonrails.org/active_record_migrations.html#preparing-the-database
# Terraform waits for this to complete before starting web_application and worker_application
```

### The long story

At the time of https://github.com/DFE-Digital/itt-mentor-services/pull/893, we were using data_migrate to differentiate between 'schema' migrations and 'data' migrations. Therefore it seemed relevant to include a comment in this Terraform config to explain why we were running two Rake commands (`db:prepare` for Rails, and `data:migrate` for data_migrate).

However we later decided it was causing more pain that it was worth, so we dropped it in https://github.com/DFE-Digital/itt-mentor-services/pull/1022.

(Although now I look at it, we've still got that same comment hanging around in `main`. We forgot to tidy it up. So I'll remove it now!)

RMcVelia (approved):

PR #4317, opened 2024-11-04
Improve link on Get school experience page for accessibility

gemmadallmandfe:
### Trello card
https://trello.com/c/p8Xsb92t

### Context
We are using the same link text (`Get school experience`) in the menu to point to the Get school experience page on GIT and on the Get school experience page itself to point to the GSE service. Using the same link text to point to different destinations is not good practice in terms of accessibility.


Sarah-DfE (approved):
Looks good to me!

PR #4340, opened 2024-11-07
Allow link text in funding widget

sarahcrack:
### Trello card

https://trello.com/c/U5rbitfq/6926-allow-link-text-in-funding-widget

### Context

We want to be able to add link text to the subject-specific content in the funding widget on our [bursaries and scholarships](https://getintoteaching.education.gov.uk/funding-and-support/scholarships-and-bursaries) page.

Adapted widget code to allow anchor tags by removing `html_safe` from `.erb` file and incorporating into `.rb` file for `funding_results`.

### Guidance to review
- check it's a suitable approach from a code perspective.
- check it has the intended behaviour needed.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -60,7 +60,7 @@ en:
         name: "Biology"
         group: "Secondary"
         sub_head: "Biology - Secondary"
-        funding: "Bursaries of %{bursaries_postgraduate_biology} are available for trainee biology teachers if you're eligible."
+        funding: "Bursaries of %{bursaries_postgraduate_biology} are available for <a href='/life-as-a-teacher/explore-subjects/biology'>trainee biology teachers</a> if you're eligible."
       business_studies:
         name: "Business studies"
         group: "Secondary"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -60,7 +60,7 @@ en:
         name: "Biology"
         group: "Secondary"
         sub_head: "Biology - Secondary"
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -60,7 +60,7 @@ en:
         name: "Biology"
         group: "Secondary"
         sub_head: "Biology - Secondary"
-        funding: "Bursaries of %{bursaries_postgraduate_biology} are available for trainee biology teachers if you're eligible."
+        funding: "Bursaries of %{bursaries_postgraduate_biology} are available for <a href='/life-as-a-teacher/explore-subjects/biology'>trainee biology teachers</a> if you're eligible."
```
martyn-w, line 63: better to use double quotes for the attributes  - I think you can escape them like this:

```
funding: "Bursaries of %{bursaries_postgraduate_biology} are available for <a href=\"/life-as-a-teacher/explore-subjects/biology\">trainee biology teachers</a> if you're eligible."
```

martyn-w (approved):
👍  LGTM

Sarah-DfE:
Looks good to me from a content point of view too. Presumably the link will be the same for each subject, just the name of the subject will change @sarahcrack ?

sarahcrack:
> Looks good to me from a content point of view too. Presumably the link will be the same for each subject, just the name of the subject will change @sarahcrack ?

Hi @Sarah-DfE Yes you just need to go into the `config/locales/en.yml` file and add the anchor links under each subject where you would like to link to the relevant subject page :) You can use biology (funding section), line 63, as a template for this. Links are very similar and looks like you can just edit the subject name at the end of the url.

PR #4352, opened 2024-11-13
Calculate lastmod date by hashing the content of the page

spencerldixon:
### Trello card

https://trello.com/c/z93KaJn6/6682-change-setting-of-lastmod-in-the-sitemap-to-reflect-date-page-actually-updated?filter=member:spencerldixon

### Context

We want to change how `<lastmod>` is set in the sitemap so that it reflects when a page has changed

- Creates a `PageModificationTracker` class which takes a list of content pages, requests them, and hashes the body. This is then stored against the path of the page with a timestamp in the database. When a pages content changes (and this task is run), the calculated hash will be different and subsequently updated in the database with the new hash and timestamp. 
- Replaces how `lastmod` is calculated in the `SitemapsController` to take date from the `PageModification` entries in the database. Order of presedence is as follows:

1. `date:` in markdown file.
2. `updated_at` from the `PageModification` record for that page (if found)
3. Default lastmod date

This allows you to override `PageModification` records by including a date in the markdown file should you wish.

### Guidance to review

Task is run with

```
require 'page_modification_tracker'

PageModificationTracker.new(host: "localhost:3000").track_page_modifications`
```

or as a job...

```
TrackPageModificationsJob.perform_later(host: "localhost:3000")
```

### DIFF
--- a/lib/page_modification_tracker.rb
+++ b/lib/page_modification_tracker.rb
@@ -32,9 +32,9 @@ class PageModificationTracker
       end
     rescue StandardError => e
       Sentry.configure_scope do |scope|
-        scope.set_context('page_modification', { path: path })
+        scope.set_context("page_modification", { path: path })
       end
-      Sentry.capture_exception(exception)
+      Sentry.capture_exception(e)
       next
     end
   end


config/lastmod.yml
```diff
@@ -0,0 +1,298 @@
+---
```
martyn-w, line 1: not sure if this file should be in source control... a change to any file may lead to a change in this file. Could it be generated during the build pipeline or stored in redis / database?

gemmadallmandfe:
@spencerldixon We would probably need to clear out the existing `date:` values on our content pages (can be done as a separate PR) as lots of them would have been set a long time ago and would be irrelevant now

app/controllers/sitemap_controller.rb
```diff
@@ -30,21 +30,28 @@ def build
...
-            xml.lastmod(DEFAULT_LASTMOD)
+            xml.lastmod(lastmod_date(events_path))
           end
         end
       end
     end
   end
 
+  def lastmod_date(path, metadata = nil)
+    page_mod = PageModification.find_by(path: path)
```
martyn-w, line 49: I think it will be more efficient to return the `metadata&.dig(:date)` if its present, before querying the database if its not

spec/lib/page_modification_tracker_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+        PageModification.create!(path: path, content_hash: content_hash)
+      end
+
+      it "does not update the page modification" do
+        expect { subject }.not_to(
+          change { PageModification.find_by(path: path).updated_at },
+        )
+      end
+    end
+
```
martyn-w, line 51: could you add a test to prove that a change to the `authenticity_token` does not invalidate the hash

.github/workflows/actions/deploy/action.yml
```diff
@@ -93,3 +93,8 @@ runs:
         env:
           HTTPAUTH_PASSWORD: ${{ steps.fetch-secrets.outputs.HTTP_PASSWORD }}
           HTTPAUTH_USERNAME: ${{ steps.fetch-secrets.outputs.HTTP_USERNAME }}
+
+      - name: Track Page Modifications
+        shell: bash
+        run: |
+          bundle exec rails runner "TrackPageModificationsJob.perform_now(host: '${{steps.variables.outputs.healthcheck}}')"
```
martyn-w, line 100: this command is currently failing in the PR:

```
 bundle exec rails runner "TrackPageModificationsJob.perform_now(host: 'get-into-teaching-app-review-4352.test')"
 ...
/home/runner/work/_temp/4a1225d2-c151-4b05-a310-0a43bb8af16a.sh: line 1: bundle: command not found
Error: Process completed with exit code 127.
```

martyn-w (dismissed):
👍 LGTM

well done!

martyn-w (approved):
👍 LGTM

gemmadallmandfe:
I've compared prod and review app sitemaps and am happy for this to be merged

I will raise a separate ticket to investigate three anomalies (events, about GIT events, mailing list start page) which still appear to be picking up the old default date. However, happy for this to be done separately!

PR #4368, opened 2024-11-21
Links styled as buttons

sarahcrack:
### Trello card 

https://trello.com/c/oklPGFGv/6930-dac-audit-2023-links-styled-as-buttons

### Context

On the ‘Get into Teaching’ homepage and subsequent pages throughout the service, there are a number of links that are visually styled as buttons. This can affect voice activation and screen reader users when interacting with a link that looks like a button. The DAC report suggests adding the ARIA `role="button"` attribute to the links to inform assistive technologies, such as screen readers and voice activation software, to treat the link as if it is a button.

- Created `LinkAsButton` component which generates a hyperlink styled as a button, with the `class="button"` and `role="button"`. This can be dynamically rendered across the site with additional attributes added if needed (please see example code below).  
- The component is now being rendered within components used on the homepage and the cookie acceptance component.

### Guidance to review

- check attributes are correct for links styled as buttons (homepage and cookie acceptance policy pop-up).
- to test the cookie pop up locally and simulate a first time user on each page reload, you can temporally change line 52 in `app/webpacker/javascript/cookie_preferences.js` to false: `this.cookieSet = false;`

- Example of code that can be used to render the `LinkAsButton` component:
```
        <%= render(Content::LinkAsButtonComponent.new(
            link_text: "Accept all cookies",
            href: "#",
            tabindex: -1,
            id: "biscuits-agree",
            css_class: "call-to-action-button",
            data: { action: "click->cookie-acceptance#accept", "cookie-acceptance-target": "agree" },
          )) %> 
```
`link_text` and `href` are compulsory but the other attributes listed are optional. You can only add attributes listed in the component; if you include an attribute that is not initialised, it will raise an error. Please note, `class="button"` and `role="button"` will be added/handled by the component. 

The above code will produce the following output:
```
<a href="#" class="button call-to-action-button" tabindex="-1" data-action="click->cookie-acceptance#accept" 
data-cookie-acceptance-target="agree" id="biscuits-agree" role="button">
   Accept all cookies
</a>
```


martyn-w (commented):
Hi Sarah,  what is the thinking for using a partial rather than a component? 

Given that the partial requires at least 2 parameters `url` and `link_text`, I wonder if a component would make more sense here? Otherwise there is no validation to explicitly declare the mandatory parameters.

app/components/content/link_as_button_component.rb
```diff
@@ -0,0 +1,30 @@
...
+    def initialize(link_text:, url:, options: {})
+      super
+
+      @link_text = substitute_values(link_text)
+      @url = url
+      @options = options
+    end
+
+    def combined_options
+      allowed_keys = %i[class tabindex data id]
```
martyn-w, line 16: this is good 👍 
...but as `allowed_keys` will never change when the application runs, it is probably better defined as a constant at the top of the class, e.g.
```
ALLOWED_KEYS = %i[class tabindex data id]
```

app/components/footer/cookie_acceptance_component.html.erb
```diff
@@ -9,19 +9,29 @@
...
+            options: {
+              tabindex: -1,
+              id: "biscuits-agree",
+              class: "call-to-action-button",
+              data: { action: "click->cookie-acceptance#accept", "cookie-acceptance-target": "agree" },
+            } )) %>
+        <%= render(Content::LinkAsButtonComponent.new(
+            link_text: "Set cookie preferences",
+            url: cookie_preference_path,
+            options: {
```
martyn-w, line 29: I'm not sure of the value of wrapping the optional parameters in the `options {}` hash - maybe cleaner to just make them conventional optional parameters ?

app/components/content/link_as_button_component.rb
martyn-w: actually ignore this comment - I think better to move the parameters as named (optional) params, e.g.

```
def initialize(text:, href:, tabindex: nil, data: {}, id: nil, class: [])
```

spec/components/content/link_as_button_component_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    let(:options) { {} }
+
+    it "renders the link as a button with default class/role of button" do
+      render_inline(component)
+      expect(page).to have_link(link_text, href: url)
+      expect(page).to have_css("a.button[role='button']", text: link_text)
+    end
+  end
+
+  context "when given an unallowed attribute" do
```
martyn-w, line 30: good 👍

jenhadfield-dfe:
Looks good to me too!

PR #4375, opened 2024-11-28
Adviser promo component

martyn-w:
### Trello card
[Redesign and build adviser promo](https://trello.com/c/uMnEGbMH/6908-redesign-and-build-adviser-promo)

### Context
New adviser promo component

### Guidance to review
[See the example page here](https://get-into-teaching-app-review-4375.test.teacherservices.cloud/train-to-be-a-teacher/if-you-have-a-degree)


jenhadfield-dfe:
Looks great thanks!

app/components/calls_to_action/adviser_component.rb
```diff
@@ -0,0 +1,19 @@
+class CallsToAction::AdviserComponent < ViewComponent::Base
+  attr_reader :title, :text, :image, :link_text, :link_target
+
+  def initialize(
+    title: "Get free one-to-one support",
+    text: "An adviser with years of teaching experience can help you to become a maths teacher. Chat by phone, text, or email as little or often as you need.",
```
gemmadallmandfe, line 6:
```suggestion
    text: "An adviser with years of teaching experience can help you to become a teacher. Chat by phone, text, or email as little or often as you need.",
```
@martyn-w I presume this is the default text that can be overwritten in the front matter of a particular page as needed?

If so, I suggest we make the default text more generic in case it is used by accident!

app/views/content/train-to-be-a-teacher/assessment-only-route-to-qts.md
```diff
@@ -9,6 +9,10 @@ image: false
 backlink: /
 promo_content:
     - content/train-to-be-a-teacher/promos/adviser-promo-assessment-only
+cta_adviser:
```
gemmadallmandfe, line 12: @martyn-w We need to remove from this assessment only page before it is merged to prod

Would it be worth replacing an adviser promo on a page where we actually want the new promo?

Eg `/train-to-be-a-teacher/if-you-have-a-degree` (would need to use the bespoke wording)

gemmadallmandfe:
Looks good @martyn-w - a couple of small comments for you

We should also add a section (or update existing guidance?) in `content.md`, either as part of this PR or as a separate ticket

gemmadallmandfe:
Thanks for adding the promo to the If you have a degree page instead @martyn-w !

That's highlighted a small issue - the if you have a degree page is not full width - but the promo is

@jenhadfield-dfe I presume the adviser promo should resize in this case to be the same size as the Find promo above it?

image

spencerldixon (approved):
LGTM 👍

PR #4377, opened 2024-11-28
Remove manual dates from content pages

spencerldixon:
### Trello card

https://trello.com/c/JBHFjGrw/6987-remove-manual-dates-from-content-pages?filter=member:spencerldixon

### Context

We are changing the way that our sitemap lastmod value is generated so that we give a better signal to Google when we have fresh content

We want to remove all instances of date in the frontmatter of content pages as these are all very out of date

- Removes hardcoded `date` in frontmatter to allow Page Modification Tracker to determine last updated at date.


gemmadallmandfe (approved):
Looks good to me

PR #4379, opened 2024-11-28
Simplify text partials

sarahcrack:
### Trello card

https://trello.com/c/lG87kT9Y/7010-simplify-partials

### Context

We want to simplify the partials to remove the div element as this is no longer needed.

Removed `<div>`s from quote partials.

### DIFF
--- a/app/views/content/shared/quotes/_quote_career_change_jack.html.erb
+++ b/app/views/content/shared/quotes/_quote_career_change_jack.html.erb
@@ -1,5 +1,4 @@
 <%= render Content::QuoteComponent.new(
   text: "I was stuck in the corporate rat race in London and just wanted to do something worthwhile.",
-  name: "Jack, French and German teacher",
-  large: false
+  name: "Jack, French and German teacher"
 ) %>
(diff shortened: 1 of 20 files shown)


app/views/content/shared/quotes/_quote_career_change_jack.html.erb
```diff
@@ -1,7 +1,5 @@
...
- <%= render Content::QuoteComponent.new(
-      text: "I was stuck in the corporate rat race in London and just wanted to do something worthwhile.",
-      name: "Jack, French and German teacher",
-      large: false
-    ) %>
-</div>
+<%= render Content::QuoteComponent.new(
+  text: "I was stuck in the corporate rat race in London and just wanted to do something worthwhile.",
+  name: "Jack, French and German teacher",
+  large: false
```
spencerldixon, line 4: Looks like the default value for `large` is `false` on QuoteComponent, we could safely remove any `large: false` to tidy up further

spencerldixon (approved):
LGTM! 👍

PR #4383, opened 2024-11-29
Added manual pages to last mod tracking

spencerldixon:
### Trello card

https://trello.com/c/e7N621jD/7017-investigate-lastmod-date-for-non-standard-content-pages

### Context

We recently changed how we assign lastmod date to the pages in our sitemap (see attached ticket)

There are a few non-standard content pages that are still using the old default date of 2021-03-01

`/events/about-get-into-teaching-events`
`/events`
`/mailinglist/signup/name`

We want to investigate these and tweak the lastmod process if possible (or change which default date they use)

- Merge in a set of manual paths with empty frontmatter to the list of pages that lastmod iterates through

### DIFF
--- a/lib/page_modification_tracker.rb
+++ b/lib/page_modification_tracker.rb
@@ -2,6 +2,12 @@ require "action_dispatch/testing/integration"
 require "digest/sha1"
 
 class PageModificationTracker
+  MANUAL_PAGES = {
+    "/events/about-get-into-teaching-events" => {},
+    "/events" => {},
+    "/mailinglist/signup/name" => {},
+  }.freeze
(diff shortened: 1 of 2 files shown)


gemmadallmandfe (approved):

lib/page_modification_tracker.rb
```diff
@@ -41,6 +41,14 @@ def track_page_modifications
 
 private
 
+  def manual_pages
```
martyn-w, line 44: could this be a constant at the top of the class?

martyn-w (approved):
👍 LGTM

PR #4385, opened 2024-11-29
Add navigation directory component to life as a teacher category pages

jenhadfield-dfe:
### Trello card
https://trello.com/c/Jd9HvLCS/6811-phase-5-repurpose-navigation-component-for-life-as-a-teacher-section


config/images.yml
```diff
@@ -308,8 +308,14 @@
...
   alt: ""
+"static/images/content/homepage/directory1.jpg":
+  # 3 teachers walking down a coridoor.
+  alt: ""
+"static/images/content/homepage/directory2.jpg":
+  # a teacher with a group of students.
+  alt: ""
 "static/images/content/homepage/directory3.jpg":
-  # An adviser talks to someone attending a event.
+  # An adviser attending a event.
```
gemmadallmandfe, line 318:
```suggestion
  # A member of the events team at an event
```
(I know this isn't alt text but we should probably be accurate in case this gets misused in future!)

app/views/content/life-as-a-teacher/age-groups-and-specialisms.md
```diff
@@ -12,4 +12,6 @@ content:
   - content/life-as-a-teacher/header
   - content/life-as-a-teacher/age-groups-and-specialisms/categories
   - content/life-as-a-teacher/how-to-become-a-teacher-cta
```
gemmadallmandfe, line 14: Do we need the how to become a teacher CTA as well as the directory? How to become a teacher is already a link in the directory and having this promo as well as the directory feels like a lot?

(this applies to other pages too)

app/views/content/life-as-a-teacher/_directory-nav.html.erb
```diff
@@ -0,0 +1,42 @@
...
+          <li><a href="/life-as-a-teacher/teaching-as-a-career/abigails-career-progression-story">Abigail's career progression story</a></li>
+        </ul>
+      <% end %>
+      <%= render Content::ImageBlockComponent.new(
+       title: "Take the next step",
+       title_color: "green",
+       image_path: "static/images/content/homepage/directory3.jpg"
+     ) do %>
+       <ul>
+         <li><a href="/teacher-training-adviser/sign_up/identity">Get an adviser</a></li>
```
gemmadallmandfe, line 34: The order of this feels slightly odd - should we group the support services together?
eg Adviser / Event / ML / How to

Should we add a link to the GSE page, as another support service? Or is that too many links

app/views/content/life-as-a-teacher/_directory-nav.html.erb
```diff
@@ -0,0 +1,42 @@
...
+         <li><a href="/steps-to-become-a-teacher">How to become a teacher</a></li>
+         <li><a href="/events">Find an event</a></li>
```
gemmadallmandfe, line 36: I think we should point to the About GIT events page here - we want to promote these

We would need to change the link text too eg `Attend a Get Into Teaching event` (or similar!)

app/views/content/life-as-a-teacher/_directory-nav.html.erb
```diff
@@ -0,0 +1,42 @@
...
+         <li><a href="/mailinglist/signup/name">Sign up for personalised updates</a></li>
```
gemmadallmandfe, line 37: I'm not sure this really describes what you are getting - is it worth having something like `Get tailored email advice`?

app/views/content/life-as-a-teacher/_directory-nav.html.erb
```diff
@@ -0,0 +1,42 @@
+<div class="row">
+  <section class="container">
+    <section class="image-blocks">
+      <%= render Content::ImageBlockComponent.new(
+        title: "Life as a teacher",
+        title_color: "yellow",
+        image_path: "static/images/content/homepage/directory1.jpg"
+      ) do %>
+        <ul>
+          <li><a href="/life-as-a-teacher/pay-and-benefits">Pay and benefits</a></li>
```
gemmadallmandfe, line 10: Having a static directory with constant links is useful in terms of maintenance but does mean you are linking to pages you are already on!

For example, this link is in the directory that appears on `/life-as-a-teacher/pay-and-benefits` - clicking on it just takes you to the top of the page you are already on

However, this is also the way that the footer works, so might not be an issue!

app/views/content/life-as-a-teacher/_directory-nav.html.erb
```diff
@@ -0,0 +1,42 @@
...
+          <li><a href="/life-as-a-teacher/change-careers">Change careers to teaching</a></li>
+        </ul>
+      <% end %>
+      <%= render Content::ImageBlockComponent.new(
+        title: "Teacher stories",
+        title_color: "pink",
+        image_path: "static/images/content/homepage/directory2.jpg"
+      ) do %>
+        <ul>
+          <li><a href="/life-as-a-teacher/pay-and-benefits/bens-favourite-things-about-teaching">Ben's favourite thing about teaching</a></li>
```
gemmadallmandfe, line 23: Should more stories be listed, now we have more?

app/views/content/life-as-a-teacher/_directory-nav.html.erb
gemmadallmandfe: NB How to become a teacher is already a link in the footer - ideally we shouldn't have duplicate links to the same destination but if we do, we need to make sure they have the same link text (which they do)

config/images.yml
jenhadfield-dfe: have updated this ta

app/views/content/life-as-a-teacher/age-groups-and-specialisms.md
jenhadfield-dfe: agreed - its too much, have removed

app/views/content/life-as-a-teacher/_directory-nav.html.erb
jenhadfield-dfe: updated

app/views/content/life-as-a-teacher/_directory-nav.html.erb
jenhadfield-dfe: we've kept it to just a few for now as the titles are longer and will update them when we can see if it gets used

app/views/content/life-as-a-teacher/_directory-nav.html.erb
jenhadfield-dfe: updated

app/views/content/life-as-a-teacher/_directory-nav.html.erb
jenhadfield-dfe: Yeah I get this, but I think people will use it to navigate the life as a teacher section so i don't think its an issue

app/views/content/life-as-a-teacher/_directory-nav.html.erb
```diff
@@ -0,0 +1,43 @@
...
+        </ul>
+      <% end %>
+      <%= render Content::ImageBlockComponent.new(
+       title: "Take the next step",
+       title_color: "green",
+       image_path: "static/images/content/homepage/directory3.jpg"
+     ) do %>
+       <ul>
+         <li><a href="/teacher-training-adviser/sign_up/identity">Get an adviser</a></li>
+         <li><a href="events/about-get-into-teaching-events">Attend a Get Into Teaching event</a></li>
```
gemmadallmandfe, line 35:
```suggestion
         <li><a href="/events/about-get-into-teaching-events">Attend a Get Into Teaching event</a></li>
```
going to a 404 currently (it's trying to add in `/life-as-a-teacher/` because the first slash is missing, I think?

jenhadfield-dfe: ah thanks, will update - couldn't run locally to check and its not working right now for some reason

emmaachesongray (approved):
Looks good to me from team content 👍

spencerldixon (approved):
LGTM 👍

gemmadallmandfe (approved):
Looks good to me

PR #4394, opened 2024-12-02
Time format check answers

sarahcrack:
### Trello card

https://trello.com/c/vSQni4QV/6962-change-time-formats-on-check-answers-page-in-adviser-funnel-on-git

### Context

Whilst reviewing the adviser funnel as part of the move to GIT, a few minor improvements were noticed. One of these concerns the time format used on the check answers page ([/teacher-training-adviser/sign_up/review_answers](https://getintoteaching.education.gov.uk/teacher-training-adviser/sign_up/review_answers)) for callback bookings. The time format should be e.g. 5:30pm.

- Minor tweak to `AnswerHelper` to format times with `:govuk_time_with_period`.
- When time is 12:00pm or 00:00am, `AnswerHelper` converts to _midday_ and _midnight_ as per [GDS guidance](https://www.gov.uk/guidance/style-guide/a-to-z#time).
- Updated `AnswerHelper` test. 

### Guidance to review

- Please double guidance, especially in relation to _midday_ and _midnight_ formatting (see above). This can be removed if not needed. 
- Cannot be merged until check is done with CRM to ensure time values passed are not affected.

### DIFF
--- a/app/helpers/answer_helper.rb
+++ b/app/helpers/answer_helper.rb
@@ -8,14 +8,16 @@ module AnswerHelper
     when Time
       formatted_time = answer.in_time_zone.to_formatted_s(:govuk_time_with_period)
 
-      # Convert "12:00pm" to "midday" as per GDS guidance
-      answer = if formatted_time == "12:00pm"
+      # Convert "12:00pm" to "midday" and "12:00am" to "midnight" as per GDS guidance
+      answer = case formatted_time
+               when "12:00pm"
                  "midday"
+               when "12:00am"
+                 "midnight"
                else
                  formatted_time
                end
     end
-
     safe_format(answer.to_s, wrapper_tag: "span")
   end
 
--- a/app/helpers/answer_helper.rb
+++ b/app/helpers/answer_helper.rb
@@ -6,7 +6,14 @@ module AnswerHelper
     when Date
       answer = answer.to_formatted_s(:govuk_date)
     when Time
-      answer = answer.in_time_zone.to_formatted_s(:govuk_time)
+      formatted_time = answer.in_time_zone.to_formatted_s(:govuk_time_with_period)
+
+      # Convert "12:00pm" to "midday" as per GDS guidance
+      answer = if formatted_time == "12:00pm"
+                 "midday"
+               else
+                 formatted_time
+               end
     end
 
     safe_format(answer.to_s, wrapper_tag: "span")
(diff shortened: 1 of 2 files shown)


Sarah-DfE:
Looks good from a content POV @sarahcrack

gemmadallmandfe:
@sarahcrack I've just discussed with @Sarah-DfE and if possible, we'd suggest the following changes:

* users select a range of time from the dropdown (eg `9:00am to 9:30am`) but we only display the **from** time on the answers page (eg `9:00am`) - are we able to show the range instead eg `9:00am to 9:30am`
* if so, we think it is better not to have `Midday` - ie it would be better to show `12:00pm to 12:30pm` as per the dropdown options

Happy to discuss if this is not possible!

app/helpers/answer_helper.rb
```diff
@@ -6,9 +6,18 @@ def format_answer(answer)
     when Date
       answer = answer.to_formatted_s(:govuk_date)
     when Time
-      answer = answer.in_time_zone.to_formatted_s(:govuk_time)
-    end
+      formatted_time = answer.in_time_zone.to_formatted_s(:govuk_time_with_period)
```
martyn-w, line 9: do the case test before casting the time to a format

martyn-w:
Confirmed that the time is correctly passed to the CRM:

Screenshot 2024-12-10 at 17 34 37

Screenshot 2024-12-10 at 17 41 07

martyn-w (approved):
👍 LGTM

gemmadallmandfe:
@sarahcrack I've retested this and I don't think it is quite working re `Midday` or `Midnight`, unfortunately

Midday is being displayed based on London time, rather than using the time zone where the user is

eg
* Change the time zone to Samoa (+13) or American Samoa (-11)
* Choose a time of 1:00am to 1:30am
* Callback time displayed in check answers is `Midday` whereas it should be 1:00am

Same happens for other time zones:
* Baku (+4) with 4pm => `Midday`
* Brasilia (-3) with 9am => `Midday`

Which means that `Midnight` will never be shown for other time zones - it always shows `12:00am`

Let's discuss whether it is worth looking at this further or not before you do anything!

gemmadallmandfe (approved):
Rested successfully - midday and midnight are now displayed for local time, not just for UTC

I have found a separate bug re the date displayed when midnight is chosen in a local time zone, but as this is already in production, I'll put a ticket in the backlog about this

Happy for this one to be merged after a tech review

martyn-w (approved):
👍  LGTM

PR #4430, opened 2024-12-11
Adds width to image blocks to prevent collapse with flexbox

spencerldixon:
### Trello card

https://trello.com/c/Jd9HvLCS/6811-phase-5-repurpose-navigation-component-for-life-as-a-teacher-section?filter=member:spencerldixon

### Context

When collapsing the page between tablet and mobile breakpoints, the content overlaps the footer.

- Adds width to `image-blocks` class to ensure that when they collapse they do not overflow

### DIFF
--- a/app/webpacker/styles/blocks.scss
+++ b/app/webpacker/styles/blocks.scss
@@ -72,6 +72,8 @@ $space-between-sections: 3.7em;
 }
 
 .image-blocks {
+  width: 100%;
+
   .image-block {
     img {
       width: 100%;


gemmadallmandfe:
@spencerldixon This looks better from an initial look at the review app but I think we should probably hold fire and deploy post code freeze so we can do a proper test

In the meantime, can you confirm which other components are affected by this styling change?

I've looked at the nav directory, the home page (directory and CTA blocks), the callback completion page (can't test events completion page as no event to test with!) and the 404 page - plus looked at various other template pages on the site

@jenhadfield-dfe @martyn-w Can you also review (but for deploy on 2 Jan!)

spencerldixon:
@gemmadallmandfe this fix is scoped to `image-blocks` only so shouldn't affect anything except the blocks on the homepage and the life as a teacher page.

jenhadfield-dfe:
looks good to me

gemmadallmandfe (approved):
Looks good to me

martyn-w:
On wide screens the image blocks overflow the righthand margin. Is this OK?

Screenshot 2024-12-17 at 09 12 00

jenhadfield-dfe:
> On wide screens the image blocks overflow the righthand margin. Is this OK?
> 
> Screenshot 2024-12-17 at 09 12 00

doesn't do this for me on chrome, is it displaying differently in different browsers?

gemmadallmandfe:
I can replicate this in Chrome
image

gemmadallmandfe:
@jenhadfield-dfe @spencerldixon Can you two decide what is needed with this when Jen is back from leave?

The PR as it stands is an improvement on prod, as it sorts out the overflow

However, there is a margin overflow issue which Martyn spotted and I replicated

Options:
1) Merge as is and put a ticket in the backlog (PRE-LIVE Development/Tech Debt column) to fix the margin overflow
2) Fix now as part of this PR and merge

My suggestion is that we only try and fix this now if it takes less than 2 hours - the fix is an improvement, the rest can wait unless very quick to sort out?

martyn-w (approved):
👍 ok, noting that the overlapping righthand margin will be fixed in a separate PR

GayeCord:
A new ticket to fix in the future has been raised, so @spencerldixon is this now ready to merge?

PR #4454, opened 2024-12-27
Removing reference to non-UK citizens on AO page

Sarah-DfE:
### Trello card

https://trello.com/c/vSiPgOaE/7141-update-content-on-assessment-only-page-to-replace-non-uk-citizens-with-outside-uk


gemmadallmandfe (approved):
Looks good to me

Before this is merged, please check with @martyn-w that no further changes are needed - the assessment only page is generated from two different files so I just want to make sure we don't need to make the changes anywhere else (doesn't look like it from a GitHub search!)

martyn-w (approved):
👍 LGTM - no changes required to the template in `lib/tasks/support/assessment-only-route-to-qts.md.erb`

PR #4465, opened 2025-01-02
create a partial guidance

joegibb:

### DIFF
--- a/docs/content.md
+++ b/docs/content.md
@@ -344,6 +344,54 @@ To create a partial go to the /app/views/content/shared folder then either selec
 ```yaml
 <div class="check-qualifcations">
   <h3>Check your qualifications</h3>
+  <p>To train to teach in England, you'll need:</p>
+  <ul>
+    <li>GCSEs at grade 4 (C) or above in English and maths (and science if you want to teach primary)</li>
+    <li>a bachelor's degree in any subject</li>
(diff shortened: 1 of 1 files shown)


emmaachesongray (commented):
This looks great! Only thing to add might be about using frontmatter in partials?

docs/content.md
```diff
@@ -344,20 +344,64 @@ To create a partial go to the /app/views/content/shared folder then either selec
...
+content:
+  - "content/shared/subject-pages/header"
+  - "content/life-as-a-teacher/explore-subjects/chemistry/article"
+```
+
+To create a variable that references the frontmatter simply copy the format below. This will look at the frontmatter, then whatever you reference in the square brackets. In this case, it will look for subject and will render chemistry. This can be changed for anything as long as it is referenced in the frontmatter. 
+```yaml
+<%= @front_matter["subject"] %> 
+```
+This means that you can use the following code over many  pages so that each page has its own unique subject rendered.
```
emmaachesongray, line 391:
```suggestion
This means that you can use the following code over many pages so that each page has its own unique subject rendered.  
```

emmaachesongray (commented):
teeny comment, but otherwise looks good to go :)

emmaachesongray (approved):
Looks good to me!

PR #4467, opened 2025-01-06
updating SEO content guidance

Sarah-DfE:
### Trello card
https://trello.com/c/TIEUPjp8/7146-update-seo-guidance-and-process

### Context
Updating content guidance so the SEO section is more detailed.


emmaachesongray (approved):
LGTM 👍

PR #4469, opened 2025-01-07
Add non-uk inset text to career changers pages

emmaachesongray:
### Trello card
https://trello.com/c/FbdUcUta/7198

### Context
Now the career changers section is split into 3 pages, we need to add the non-UK expander back into 2 of these pages.

The non-UK inset needs adding to the following pages:

Support for career changers | Get Into Teaching GOV.UK

How to change careers to become a teacher | Get Into Teaching GOV.UK


Sarah-DfE (approved):
Looks good to me!

PR #4485, opened 2025-01-14
Replace broken link on veterans page

gemmadallmandfe:
### Trello card
https://trello.com/c/dkGQTRW3

### Context
The Career Transition Partnership website has moved so we need to change the link on the support for veterans' page.


Sarah-DfE (approved):
Looks good to me!

PR #4506, opened 2025-01-22
Routes wizard early years and further education section

spencerldixon:
### Trello card

https://trello.com/c/qlszIoje/7280-add-early-years-and-further-education-to-routes-wizard-results?filter=member:spencerldixon

### Context

We want to add in a section to the results page referencing early years and further education for some of the answer combinations (but not all)

To begin with, we want to show this content only if the user has selected No for the degree question

- Adds helper for `no_degree` option selected
- Adds section for early years and further education


gemmadallmandfe (approved):
Looks good to me!

@Sarah-DfE are you happy with the content before this is merged into the main branch?

PR #4509, opened 2025-01-23
Send signal to Clarity when cookie consent is given

sarahcrack:
### Trello card

https://trello.com/c/sR8YBd2Z/7246-send-signal-to-clarity-when-cookie-consent-is-given

### Context

We were advised that we may need make changes to the way we drop Clarity cookies:

Clarity relies on cookies to connect page views into sessions, which are critical for many of Clarity’s features. Clarity is transitioning to requiring explicit consent signals for cookies in the EEA, UK and Switzerland.

We are currently not dropping any clarity cookies until a user has consented to accept cookies. However, we need to check if we then set a flag to advise Clarity of the consent status for each individual visitor to our site. Without this consent Clarity's recordings won't be linked together into multi-page sessions. Heat maps and data other than session recordings will continue to be available.

I could not see a flag set to advise Clarity of the users consent status, so followed the steps provided in Clarity's cookie consent guidance[: ](https://learn.microsoft.com/en-us/clarity/setup-and-installation/cookie-consent#step-2)[Clarity Cookie Consent](https://learn.microsoft.com/en-us/clarity/setup-and-installation/cookie-consent).

### DIFF
--- a/app/webpacker/javascript/clarity.js
+++ b/app/webpacker/javascript/clarity.js
@@ -45,11 +45,21 @@ export default class Clarity {
 
   listenForConsentChanges() {
     document.addEventListener('cookies:accepted', () => {
+      console.log('cookies:accepted event fired');
+
       if (this.consentValue('non-functional') === 'granted') {
         console.log('User consent granted');
         this.initContainer();
         window.clarity('consent');
         console.log('Clarity consent granted');
+      } else if (this.consentValue('non-functional') === 'denied') {
+        console.log('User consent denied');
+        window.clarity('consent', false); // Inform Clarity that consent is revoked
+        console.log('Clarity consent revoked');
+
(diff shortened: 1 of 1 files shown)


app/webpacker/javascript/clarity.js
```diff
@@ -46,7 +46,10 @@ export default class Clarity {
   listenForConsentChanges() {
     document.addEventListener('cookies:accepted', () => {
       if (this.consentValue('non-functional') === 'granted') {
```
martyn-w, line 48: Should we add another condition, so that if consent is later revoked, we inform Clarity not to track with:

```
window.clarity('consent', false)
```

martyn-w (dismissed):
👍  LGTM - just remove the console logs

martyn-w (approved):
👍  LGTM

PR #4511, opened 2025-01-23
Routes wizard tracking

spencerldixon:
### Trello card

https://trello.com/c/dGj3fYgS/7295-set-up-tracking-for-answer-combinations-given-within-the-route-wizard?filter=member:spencerldixon

### Context

We want to capture the combination of answers given to each of the 3 questions in the routes wizard (ie when someone clicks on the Complete button)

- Adds hidden divs with data tags for each answer to the completed page


gemmadallmandfe (approved):
Ticking on before of Jess who is happy from a tracking perspective

PR #4525, opened 2025-01-28
Routes wizard back button

spencerldixon:
### Trello card

https://trello.com/c/e2pnl2Bs

### Context

We need to test the routes wizard before it is deployed to production

- Adds a back button to the first question of the wizard
- Removes the Non-UK expander component


gemmadallmandfe (approved):
Looks good to me

Tested the following successfully:

* Back button on degree step
* Removal of non-UK expander on visas
* Aria label on separators so they are not read out by screen readers

PR #4530, opened 2025-01-29
Update funding on TDA for routes

emmaachesongray:
### Context
Remove funding content from TDA on routes, so it's aligned with salaried and TF options.


gemmadallmandfe (approved):
Looks good to me

spencerldixon (approved):
LGTM 👍

PR #4534, opened 2025-01-30
Replace apply promo

emmaachesongray:
### Trello card
https://trello.com/c/zWH7T0K6/7346

### Context
We've redesigned the apply promo to make it easier to use. 

Apply the new promo to the pages and delete the old promo file.


gemmadallmandfe (approved):
Looks good from a front-end perspective but definitely needs a tech review too!

PR #4535, opened 2025-01-30
Adds conditional extra question to degree status step of mailing list

spencerldixon:
### Trello card

https://trello.com/c/w6zuCY7I/7320-update-options-on-degree-step-in-mailing-list-and-add-conditional-graduation-year

https://trello.com/c/WcGx8GBu/7525-update-mailing-list-completion-page-logic-to-work-with-the-new-degree-step-options?filter=member:spencerldixon

### Context

We want to capture graduation year of current students so we can better segment different year groups

- Adds a conditional question block that is rendered when the corresponding radio button is selected
- Removes `degree_status_text` in favour of generic `format_option_value` helper
- Adds validations to graduation year field
- Refactors welcome guide and helpers on completion page to new logic


app/views/mailing_list/steps/_degree_status.html.erb
```diff
@@ -10,6 +10,11 @@
   f.object.degree_status_options,
   :id,
   ->(option) { degree_status_text(option) },
```
spencerldixon, line 12: Making a quick note, since I've introduced the `format_option_value` helper, I think we can get rid of this `degree_status_text` now as the new helper is much more generic and cleaner

gemmadallmandfe:
@spencerldixon Please could you update this error message to be `Choose a degree option from the list`

image

app/views/mailing_list/steps/_degree_status.html.erb
```diff
@@ -9,7 +9,19 @@
...
-  ->(option) { degree_status_text(option) },
+  ->(option) { format_option_value(option, "mailing_list_steps.teacher_training.degree_status") },
+  lambda { |option|
+    if option.id == MailingList::Steps::DegreeStatus::GRADUATION_YEAR_DEPENDENT_OPTION_ID
+      content_tag(:div, class: "field__hidden_subfield") do
+        f.govuk_text_field(
+          :graduation_year,
+          maxlength: 4,
+          label: { text: "Which year will you graduate?" },
+          hint: { text: "For example; #{Date.current.year + 2}"}
```
gemmadallmandfe, line 20:
```suggestion
          hint: { text: "For example, #{Date.current.year + 2}"}
```

gemmadallmandfe:
@spencerldixon Can you update the error message if no graduation year is entered to be `Tell us which year you will graduate`

image

gemmadallmandfe:
@spencerldixon Can you please replace this error message for year in the past to be `Enter the current year or a year in the future`

image

gemmadallmandfe:
@spencerldixon Can you please replace this error message with `Enter a valid graduation year` (NB we might want to tweak again before deploying as I'm not sure this is helpful enough!)

image

app/views/mailing_list/steps/completed.html.erb
```diff
@@ -13,24 +13,36 @@
...
+
+    <% if low_commitment? && (studying? || no_degree?) %>
+      <%= render "mailing_list/steps/completed/shared/is_teaching_right_for_me" %>
+    <% end %>
+
+    <% if no_degree? %>
+      <%= render(partial: "mailing_list/steps/completed/shared/train_to_teach_no_degree") %>
+
+      <p>
+      Consider <a href="/life-as-a-teacher/age-groups-and-specialisms/further-education-teachers">teaching in further education</a> without a degree.
```
gemmadallmandfe, line 40: Not sure why we hardcode this bit here - shouldn't it sit within a partial for consistency @spencerldixon ?

Would suggest this just goes within the `train_to_teach_no_degree` partial and appears for all no degree people to make it easier for now

app/models/mailing_list/steps/degree_status.rb
```diff
@@ -1,12 +1,29 @@
...
+      validates :graduation_year,
+                format: {
+                  with: /\A\d{4}\z/,
+                  message: "Tell us which year you will graduate",
+                },
+                numericality: {
+                  only_integer: true,
+                  greater_than_or_equal_to: Time.current.year,
+                  less_than_or_equal_to: Time.current.year + 10,
+                  message: "Enter the current year or a year in the future",
```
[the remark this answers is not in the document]
spencerldixon: Fixed! So this was a bit of a bug, it turns out you can't have multiple `numericality` validations together. I've broken this out into separate validations, should work as expected now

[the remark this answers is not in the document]
gemmadallmandfe: Retested successfully - I think we still need to update the error message itself but that is one for @Sarah-DfE to advise on!

gemmadallmandfe:
@spencerldixon I think we need to add an `h3` to the following partials for accessibility (wrong in prod - not anything you've done!)

Could you please add the following changes to this PR:

`_is_teaching_right_for_me.html.erb` 
* Add the following: <h3 class="heading-s">Explore what teaching is really like</h3>

`_steps_to_become_a_teacher.html.erb`
* Add the following: <h3 class="heading-s">Find out how to become a teacher</h3>

`_train_to_teach_no_degree.html.erb`
* Add the following: <h3 class="heading-s">Learn about routes into teaching</h3>

@Sarah-DfE Once these changes are in the PR, I'll walk you through them as they may need tweaking from a content perspective!

app/views/event_steps/_personalised_updates.html.erb
```diff
@@ -11,7 +11,7 @@
 <%= f.govuk_collection_select :degree_status_id,
       f.object.degree_status_options,
       :id,
-      ->(option) { degree_status_text(option) },
+      ->(option) { format_option_value(option, "mailing_list_steps.teacher_training.degree_status") },
```
martyn-w, line 14: should this be `event_steps` not `mailing_list_steps`?

spencerldixon: Ah I believe I'm just reusing the front end text from the mailing_list_steps translation here, it should read the same in this section as it does in the mailing list

martyn-w (dismissed):
👍 LGTM

gemmadallmandfe:
@spencerldixon Thanks for sorting the events funnel!

Is it possible to change the values in the dropdown to match what they are in prod/staging:

**Staging**
image


**Review app**
image

app/models/mailing_list/steps/degree_status.rb
```diff
@@ -1,12 +1,27 @@
...
 
       validates :degree_status_id,
-                presence: true,
+                presence: { message: "Choose a degree option from the list" },
                 inclusion: { in: :degree_status_option_ids }
 
+      validates :graduation_year,
+                format: {
+                  with: /\A\d{4}\z/,
+                  message: "Tell us which year you will graduate",
```
gemmadallmandfe, line 16:
```suggestion
                  message: "Enter your expected graduation year",
```

app/models/mailing_list/steps/degree_status.rb
```diff
@@ -1,12 +1,27 @@
...
+                },
+                numericality: { only_integer: true },
+                presence: { message: "Tell us which year you will graduate" },
```
gemmadallmandfe, line 19:
```suggestion
                presence: { message: "Enter your expected graduation year" },
```

app/models/mailing_list/steps/degree_status.rb
```diff
@@ -22,6 +37,22 @@ def degree_status_option_ids
...
         PickListItemsApiPresenter.new.get_qualification_degree_status
       end
+
+      def requires_graduation_year?
+        degree_status_id == GRADUATION_YEAR_DEPENDENT_OPTION_ID
+      end
+
+      def graduation_year_not_in_the_past
+        if graduation_year.present? && graduation_year < Time.current.year
+          errors.add(:graduation_year, "Enter the current year or a year in the future")
```
gemmadallmandfe, line 47:
```suggestion
          errors.add(:graduation_year, "Your expected graduation year cannot be in the past")
```

app/models/mailing_list/steps/degree_status.rb
```diff
@@ -22,6 +37,22 @@ def degree_status_option_ids
...
+        end
+      end
+
+      def graduation_year_not_too_far_in_the_future
+        if graduation_year.present? && graduation_year >= Time.current.year + 10
+          errors.add(:graduation_year, "Enter a valid graduation year")
```
gemmadallmandfe, line 53:
```suggestion
          errors.add(:graduation_year, "Your expected graduation year cannot be more than 10 years from now")
```

gemmadallmandfe (approved):
Looks good to me from a front-end perspective and testing through to CRM

However, needs a tech review first and also we need to sequence API and ruby client changes so this should not be merged yet

Gemfile
```diff
@@ -85,7 +85,9 @@ gem "text"
 gem "indefinite_article"
 
 gem "connection_pool"
-gem "get_into_teaching_api_client_faraday", ">= 3.1.3", github: "DFE-Digital/get-into-teaching-api-ruby-client", require: "api/client"
+# gem "get_into_teaching_api_client_faraday", ">= 3.1.3", github: "DFE-Digital/get-into-teaching-api-ruby-client", require: "api/client"
+# gem "get_into_teaching_api_client_faraday", ">= 3.5.1", github: "DFE-Digital/get-into-teaching-api-ruby-client", require: "api/client", branch: "add-graduation-year-mailing-list"
```
carlosmartinez, line 89:
```suggestion
```

app/helpers/welcome_helper.rb
```diff
@@ -1,16 +1,31 @@
...
-
-    degree_status.in?(final_year_student) || (
-      degree_status.in?(gradudate_or_postgraduate) && consideration_journey_stage.in?(allowed_graduate_consideration_stages)
-    )
+    low_commitment?(consideration_journey_stage: consideration_journey_stage) && graduate?(degree_status: degree_status)
+  end
+
+  def high_commitment?(consideration_journey_stage: consideration_journey_stage_id)
+    # I think I'll apply
+    consideration_journey_stage == 222_750_003
```
carlosmartinez, line 8: what do you think of extracting these magic numbers to constants?

app/models/crm/option_set.rb
```diff
@@ -41,7 +54,12 @@ def lookup_by_values(category, *values)
     end
 
     def lookup_const(category)
-      const = const_get(category.to_s.pluralize.upcase)
+      const = begin
+                const_get(category.to_s.upcase)
+              rescue NameError
+                const_get(category.to_s.pluralize.upcase)
+              end
```
carlosmartinez, line 61: suggest making the intention clearer here. i can understand what it's doing, but i have to read the code pretty closely to do so!

carlosmartinez:
Looks good. Just a couple of comments. Also, how to get the build passing?

Gemfile
```diff
@@ -85,7 +85,9 @@ gem "text"
...
+gem "get_into_teaching_api_client_faraday", ">= 3.5.1", github: "DFE-Digital/get-into-teaching-api-ruby-client", require: "api/client", branch: "feature/add_graduation_year_and_inference_response"
```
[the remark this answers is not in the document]
spencerldixon: Yep! This will be changed when the API gets merged

app/models/events/steps/personalised_updates.rb
```diff
@@ -30,7 +30,9 @@ def export
       end
 
       def degree_status_options
-        @degree_status_options ||= PickListItemsApiPresenter.new.get_qualification_degree_status
+        @degree_status_options ||= GetIntoTeachingApiClient::PickListItemsApi.new
+          .get_qualification_degree_status
+          .reject { |item| item.id == 222_750_006 }
```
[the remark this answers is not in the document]
spencerldixon: Not sure this is the right approach here as we're excluding this for a different reason and it is unrelated to the `GRADUATION_YEAR_DEPENDENT_OPTION_ID`.

So we've updated the mailing list degree step to be only 3 options and take a graduation year. To do this we added an extra option to the CRM `Not yet, I'm studying for one` (users will select this and enter their graduation year). Since we don't want to roll out this change to the event funnel just yet, we need to exclude the newly added `Not yet, I'm studying for one` option and retain the original CRM options.

Ideally, I'd set the presenter up so we could have another option, but since we're aiming to bring the two funnels in line after this PR (using the same API presenter call), the best way is to exclude this for now.

I've updated this to pull from the `Crm::OptionSet` which acts as a store for these crm values so at least it's not dealing with the raw ID

app/models/crm/option_set.rb
spencerldixon: Happy to do so! I tried refactoring it a few ways but always came back to this because it's nice and concise, do you have any ideas?

Originally this method only dealt with pluralised constants, which wasn't great when we added a singular constant ending in `advertising`, so you'd have to refer to it as `advertisings` which just felt horrible, so I added the check for singular first, then fall back to plural.

martyn-w:
👍 LGTM - but needs approval from @carlosmartinez

carlosmartinez (dismissed):

gemmadallmandfe (dismissed):
Looks good to me from the front end

PR #4549, opened 2025-02-04
Update Rails to version 7.0.8.7

sarahcrack:
### Trello card

https://trello.com/c/y1CdeQwo/7402-update-rails-to-version-7087-for-git

### Context

The version of Rails we are using for GIT is very out of date. We are currently on 7.0.2.4 (released April 2022) whereas the latest version is 8.0.1. We want to upgrade it in stages.

As a first step we want to upgrade to 7.0.8.7 to ensure that we have all the latest security patches.

- Bumped rails to 7.0.8.7. 
- fixed `"concurrent-ruby", "1.3.4"` in Gemfile - this was required to fix a dependency issue with Rails upgrade to version 7.0.8.7; concurrent-ruby v1.3.5 removed the dependency on logger which was causing errors (`require "logger"` did not seem to fix the issue on GiT). When updating to Rails 7.1, this gem fix can likely be removed.
- went through rails update task to incorporate and/or merge suggested files as part of the upgrade.
- checked app visually

### Guidance to review

- [useful article](https://stackoverflow.com/questions/79360526/uninitialized-constant-activesupportloggerthreadsafelevellogger-nameerror) on errors we were getting as part of the upgrade.


gemmadallmandfe (commented):
Looks good to me from a front end perspective (but haven't approved as the tech review is more important!)

Gemfile
```diff
@@ -4,7 +4,8 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby File.read(".ruby-version").chomp
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem "rails", "~> 7.0.2.3"
+gem "concurrent-ruby", "1.3.4" # Required to fix a dependency issue with Rails upgrade to version 7.0.8.7
```
martyn-w, line 7: unless a specific version of the gem is required, its better to use the looser version number `"~> 1.3.4"` to allow for small automatic upgrades

Gemfile
```diff
@@ -4,7 +4,8 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
...
+gem "rails", "7.0.8.7"
```
martyn-w, line 8: use version `"~> 7.0.8.7"`

martyn-w (dismissed):
👍 LGTM
I've made a small comment about using looser version numbers unless an absolutely specific version number is required

Gemfile
sarahcrack: Thanks Martyn! For this particular upgrade, we don't want to go any higher than version 1.3.4 for concurrent-ruby (1.3.5 seemed to be the issue with this Rails upgrade). So will leave as is for now (not allowing any patch updates) and remove/adapt with the next upgrade :)

martyn-w: Hi Sarah, sorry I didn't realise that - leave it exactly as you have specified 👍

martyn-w (approved):

PR #4553, opened 2025-02-05
Upgrade Ruby from 3.1 to 3.4

martyn-w:
### Trello card
[Upgrade the Ruby version on Get Into Teaching app from v3.1 to v3.4](https://trello.com/c/VYgJ0v2d/7387-upgrade-the-ruby-version-on-get-into-teaching-app-from-v31-to-v34)

### Context
In order to keep our platform up-to-date, we need to move newer version of Ruby (3.4)

Upgrading Ruby has required updating some spec tests, as there is now a different treatment of hash parameters vs named arguments when using spies and mocks.
It is also necessary to disable chrome-based capybara tests as newer versions of Chrome cause random failures when running within capybara.


spencerldixon (approved):
LGTM 🤞

gemmadallmandfe (approved):

PR #4554, opened 2025-02-05
Update action.yml

martyn-w:
### Trello card
[Fix OWASP check error](https://trello.com/c/tODlgers/7406-fix-owasp-check-error)

### Context
updating the zaproxy OWASP check version


spencerldixon (approved):
LGTM 👍

PR #4562, opened 2025-02-07
Update content on consideration step in events with mailing list funnel

emmaachesongray:
### Trello card
https://trello.com/c/pLaLtYEy/7413

### Context
We want to update the consideration step to make it easier for users to complete and more useful in terms of segmenting them.

We need to updated the content on the consideration step in the events with mailing list funnel.


gemmadallmandfe (approved):
Looks good to me (but needs a tech review before merging!)

spencerldixon (approved):
LGTM 👍

PR #4563, opened 2025-02-07
Bring mobile fonts larger than medium in line with tablet fonts

spencerldixon:
### Trello card

https://trello.com/c/8Fvr9Pzl/7407-adjust-h3-styling-so-that-h3s-are-bigger-on-mobile

### Context

Our h3s on desktop appear significantly bigger than those on mobile

This is particularly noticeable in our new routes wizard

- Reset all font sizes to 100% for consistency (i notice some fonts shrink at tablet size due to mobile and tablet being set to 106% and desktop being set to 100%)
- For mobile sizes, bring every font-size medium and above in line with tablet sizes (xsmall and small remain original sizes)

### DIFF
--- a/app/webpacker/styles/font-sizes.scss
+++ b/app/webpacker/styles/font-sizes.scss
@@ -1,6 +1,6 @@
 html {
-  @include mq($until: mobile) { font-size: 106%; }
-  @include mq($from: mobile, $until: tablet) { font-size: 106%; }
+  @include mq($until: mobile) { font-size: 100%; }
+  @include mq($from: mobile, $until: tablet) { font-size: 100%; }
   @include mq($from: tablet) { font-size: 100%; }
 }
 
--- a/app/webpacker/styles/font-sizes.scss
+++ b/app/webpacker/styles/font-sizes.scss
@@ -18,11 +18,11 @@ $_font-sizes: (
   mobile: (
     "xsmall":   nth($_relative-font-sizes, 1),
     "small":    nth($_relative-font-sizes, 2),
-    "medium":   nth($_relative-font-sizes, 3),
-    "large":    nth($_relative-font-sizes, 4),
-    "xlarge":   nth($_relative-font-sizes, 5),
-    "xxlarge":  nth($_relative-font-sizes, 6),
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
I'll leave @jenhadfield-dfe to review this properly as the expert

I think the h3s generally look much better on mobile now across the site

A couple of things I've noticed from a quick look around:

**1) Help and support page**

h3s are pretty big here compared with the h2s
However, I think this is more to do with the pink `Talk to us` h2 being smaller than other h2s on the page!

(Also the Non-UK citizens box should probably be an h3, not an h2 I think?)

image

**2) Other font sizes reduced**
Eg on assessment only listings, the h3 section heading (`National`) has increased in size which is good but the remainder of the listing (h4 provider name and other listing info) looks slightly smaller now - is this an issue?

Current
image

Review app
image


Similarly on the , the h4 looks smaller than the body text now?

Current
image

Review app
image

I'm not saying these need changing necessarily - just flagging for Jen's review

spencerldixon:
A bit of odd behaviour I noticed prior to this work is that when resizing the page from mobile to desktop, there would be a point where the font would shrink a little around tablet size to be smaller than the mobile size font.

I would have expected that the font should only get progressively larger from mobile to tablet to desktop, with mobile being the smallest possible font size.

It was a very subtle change, but felt off to me. I believe this was related to lines 2-3 being set to 106% font size, desktop and above being set to 100% font size. This may have created a small sizing zone where the font size remains the same but the jump between 106% and 100% makes the font appear to have shrunk below mobile size (106%).

To correct this I reset everything to 100%. This means that everything is 6% smaller, but it keeps everything much more consistent.

I'm a little unsure of this as the decision was made at some point to scale everything to 106% and I'm not sure why.

We can revert the change if needed and just go with the heading changes for now

jenhadfield-dfe:
hey, that makes sense, I agree with making everything consistent at 100% and think the changes look good. Looked at Gemmas review comments above and think they are all ok - but may investigate the H2 on the non-uk expander - but we could do that on a separate ticket

jenhadfield-dfe (approved):

PR #4568, opened 2025-02-07
updating the guidance on insets and expanders

Sarah-DfE:
### Trello card
https://trello.com/c/xnIsFyGt/7341-improve-content-guidance-on-insets-and-expanders

### DIFF
--- a/docs/content.md
+++ b/docs/content.md
@@ -283,9 +283,11 @@ When adding an iFrame elemet as part of Markdown content or a HTML page we shoul
 ></iframe>
 ```
 
-### Inset text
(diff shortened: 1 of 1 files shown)


emmaachesongray (commented):
Love the details you've added! I'm finding it a bit confusing because we have this section but then also one on 'details expander for non-uk content' which overlaps with this one. Could we combine them?

emmaachesongray (commented):
The contents table links need to be updated, but everything else looks super clear!

emmaachesongray (approved):
LGTM 👍

PR #4576, opened 2025-02-12
Removed unneeded consideration steps and renamed option value text

spencerldixon:
### Trello card

https://trello.com/c/GKPSwaTU/7449-remove-unneeded-consideration-step-options-and-reword-the-remaining-ones-in-mailing-list?filter=member:spencerldixon

### Context

We want to remove some of the consideration options and reword the remaining ones to make them easier to understand for users and easier to use by support services

We want to remove the two highlighted options from the front-end

We also want to change the label of the fourth option

FROM I'm very sure and think I'll apply

TO I think I'll apply

- Allow lists the options we want to filter from the PickList API for controlling presentation options in GIT
- Adds a `format_option_value` helper for replacing text in the translations
- Removes the two options from the front end by filtering out other options from the picklist
- Changes the label of the last option via translations

### DIFF
--- a/app/helpers/option_set_helper.rb
+++ b/app/helpers/option_set_helper.rb
@@ -3,4 +3,8 @@ module OptionSetHelper
     key = option.value.parameterize(separator: "_")
     I18n.t("helpers.answer.mailing_list_steps_degree_status.degree_status.#{key}", **{ default: option.value }.merge(Value.data))
   end
+
+  def format_option_value(option, translation_key = nil)
+    I18n.t("helpers.answer.#{translation_key}.#{option.id}", **{ default: option.value }.merge(Value.data))
+  end
 end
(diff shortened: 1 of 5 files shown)


app/helpers/option_set_helper.rb
```diff
@@ -3,4 +3,8 @@ def degree_status_text(option)
     key = option.value.parameterize(separator: "_")
     I18n.t("helpers.answer.mailing_list_steps_degree_status.degree_status.#{key}", **{ default: option.value }.merge(Value.data))
   end
+
+  def format_option_value(option, translation_key = nil)
```
spencerldixon, line 7: I've added this helper to be general purpose for translating picklist options by the id rather than their current value. With a bit of refactoring we could remove the above degree status specific helper in favour of this and keep everything generic rather than step specific.

martyn-w (approved):
👍  LGTM

gemmadallmandfe (approved):

However, this should not be merged until after we have demoed on Monday

gemmadallmandfe:
Successful demo with Direct Marketing team this morning

Happy for this to be merged @spencerldixon !

PR #4581, opened 2025-02-14
Format consideration journey pick list for event steps

spencerldixon:
### Trello card

https://trello.com/c/NqCOFYzi/7448-remove-unneeded-consideration-step-options-and-reword-the-remaining-ones-in-events-with-mailing-list-funnel?filter=member:spencerldixon

### Context

We want to remove the two highlighted options from the front-end

We also want to change the label of the fourth option

FROM I'm very sure and think I'll apply

TO I think I'll apply

Note that we should not change the value we are passing to CRM

Since we've now got two places we want to filter the picklist (mailing list and events), this PR introduceds a picklist presenter pattern that wraps the API, and returns just the options we want to deal with in the frontend.

- Filters pick list to just the items we want to display
- Formats the text on the front end to match new description
- Adds tests


martyn-w (approved):
👍 very good

gemmadallmandfe (approved):
Looks good to me from the front-end - have checked that consideration stage is captured within CRM

PR #4589, opened 2025-02-17
Bump rails to 7.1 on GiT

sarahcrack:
### Trello card

https://trello.com/c/U9SR1O4I/7445-update-rails-to-version-71-for-git

### Context

The version of Rails we are using for GIT is very out of date. The latest version is 8.0.1. We want to upgrade it in stages and
as an interim step we want to upgrade to 7.1.

- updated `gem "rails", "~> 7.1.3"`
- updated `gem "validates_timeliness", "~> 7.0.0"` as this was blocking the update.
- following the update, the app would not start up. It looks like the issue stems from Sidekiq 6.5.5 not supporting Rails 7.1 because it relies on `ActiveSupport::Logger.broadcast`, which was removed in Rails 7.1. It seems that updating logger configuration for sidekiq is one solution:  created a sidekiq initializer which ensures that sidekiq uses `Rails.logger` for logging instead of its default logger; this aligns sidekiq’s logs with the rest of the Rails application without overriding the global logger configuration in `application.rb`.
- updated `gem "rspec-rails", "~> 6.1.0"` which supports Rails 7.1 and allows rspec tests to run.
- updated redirects spec test which was failing after the updates.
- ran update task copying over code into new files.


spencerldixon (approved):
LGTM 👍

PR #4596, opened 2025-02-19
Replace footer tag in quotes with a div tag

spencerldixon:
### Trello card

https://trello.com/c/7bDrM3jr/7501-fix-erroneous-footer-landmark-in-quotes?filter=member:spencerldixon

### Context

ARC Toolkit has flagged that our quotes have an erroneous footer landmark, which means we have multiple areas on a page flagged as footers

See screenshot as an example (/life-as-a-teacher/why-teach/why-become-a-teacher)

This is confusing for screen reader users

- Replaces the `footer` tag with a `div` tag


martyn-w (approved):
👍 LGTM

gemmadallmandfe (approved):
Looks good to me - retested with ARC toolkit

PR #4600, opened 2025-02-19
Added return false for when no answers present for highlight check

spencerldixon:
### Trello card

https://trello.com/c/S05TkSxI/7504-fix-error-accessing-the-route-into-teaching-completed-page-with-no-or-partial-answers

### Context

The new Routes into teaching wizard is causing multiple sentry errors

This is triggered by having no or partial answers being passed to the highlighted logic

- Adds check to return false if there are no answers for the routes finder


martyn-w (approved):
👍 LGTM

PR #4601, opened 2025-02-19
Replaces header tag in landing hero component with div

spencerldixon:
### Trello card

http://trello.com/c/rYjt2grx/7505-remove-duplicate-header-landmark-on-category-sub-category-and-inspiration-pages?filter=member:spencerldixon

### Context

Pages should only have one headerelement which is used by keyboard users and screen readers when navigating the page

ARC Toolkit has flagged that pages using a particular header have two headerelements (see screenshot)

This can be seen in the code (see screenshot)

This header component is used at different levels eg

/life-as-a-teacher

/life-as-a-teacher/explore-subjects

/life-as-a-teacher/explore-subjects/art-and-design

- Replaces the header tag in the `LandingHeroComponent` with a div given that the `header` tag is being used for the main navigation bar


jenhadfield-dfe:
looks good to me too!

PR #4605, opened 2025-02-21
green arrow links on non-UK trainee pages

Sarah-DfE:
### Trello card
https://trello.com/c/1Bnkdvqf/7304-identify-where-new-cta-arrows-links-should-be-added-and-update-pages


app/views/content/non-uk-teachers/fees-and-funding-for-non-uk-trainees.md
```diff
@@ -20,8 +20,17 @@ expander:
...
+cta_arrow_link:
+  loans:
+    link_target: "https://www.gov.uk/student-finance"
+    link_text: "Check your eligibility and apply for student loans"
+  bursaries:
+    link_target: "/funding-and-support/scholarships-and-bursaries"
+    link_text: "Learn more about bursaries and scholarships for physics and languages"
+  find:
+    link_target: "https://find-teacher-training-courses.service.gov.uk"
+    link_text: "Find postgraduate teacher training and filter by salaried courses"
```
gemmadallmandfe, line 32: Is this slightly risky that we are highlighting salaried courses so much through the green arrow? And sending people straight to Find rather than explain more about salaried courses first (they may well miss the first link in the section?)

Would it be better to remove the `work and earn a salary while you train` link and replace the green arrow link with a 'Find out more about salaried teacher training`? Particularly as we are updating the salaried page with more guidance re placements, etc.?

PR #4615, opened 2025-02-27
Phone number styled as interactive on help and support page

sarahcrack:
### Trello card

https://trello.com/c/8JIryjs3/7550-gds-accessibility-audit-8-phone-number-not-styled-as-interactive-on-help-and-support-page-level-a

### Context

On the Help and support getting into teaching page, there is a telephone number link beneath “Call us” in the “Talk to us” section. The telephone number is visually styled similar to a heading or important text and does not look interactive. The visual styling does not match the programmatic information.

- visually styled the phone number as interactive (similar to the phone number in the footer)
- adapted `class` in HTML and added an `underline` styling

###  Guidance to review
Screenshot 2025-02-27 at 11 34 04

### DIFF
--- a/app/webpacker/styles/sections/talk-to-us.scss
+++ b/app/webpacker/styles/sections/talk-to-us.scss
@@ -101,7 +101,7 @@
 
     .telephone {
       font-weight: bold;
-      text-decoration: none;
+      text-decoration: underline;
       font-size: 1.3em;
     }
   }
--- a/app/webpacker/styles/sections/talk-to-us.scss
+++ b/app/webpacker/styles/sections/talk-to-us.scss
@@ -101,8 +101,7 @@
 
     .telephone {
       font-weight: bold;
-      text-decoration: underline;
-      font-size: 1.3em;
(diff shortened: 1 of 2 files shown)


app/webpacker/styles/sections/talk-to-us.scss
```diff
@@ -101,7 +101,7 @@
 
     .telephone {
       font-weight: bold;
-      text-decoration: none;
+      text-decoration: underline;
```
gemmadallmandfe, line 104: Do we need this `underline`? The phone link in the footer in prod (which uses the same `telephone` class) already has a visual underline as do all our links

Paging @jenhadfield-dfe who can advise much better than me!

jenhadfield-dfe: actually i think you could remove the text decoration entirely and it should still underline it. Also for the font size, i think it might be better at 1.75em rather than 1.3, so its larger and same size as the one in the footer, just a bit more obvious, seems a bit small now im looking at it

PR #4637, opened 2025-03-05
Page titles in routes funnel

sarahcrack:
### Trello card

https://trello.com/c/3FK5ZE6Z/7574-gds-accessibility-audit-26-insufficiently-descriptive-page-titles-in-routes-funnel

### Context

WCAG 2.4.2 Page titled: https://www.w3.org/WAI/WCAG22/Understanding/page-titled

Pages must have titles that describe the topic or purpose of the page. This helps users avoid having to read or search through content to see if it is relevant. Good titles are descriptive, meaningful and unique. In most browsers the title will usually be displayed in the top title bar or as the tab name.

Throughout the Routes journey, the page title does not communicate to users which journey they are going through, it only conveys the current step such as “Do you have an undergraduate degree?”. This includes the last page with a title of “Find your route into teaching” which is the same as the first page.

- removed page titles from routes wizard steps (views) and added text for page.
- adapted `StepsController` to set title for completed page.
- added `title` method to degree step model to display requested page title: `Find your route into teaching, degree step`. `location` and `unqualified_teacher` step can remain as is; default wizard behaviour displays required page title already

### DIFF
--- a/app/controllers/routes_into_teaching/steps_controller.rb
+++ b/app/controllers/routes_into_teaching/steps_controller.rb
@@ -6,6 +6,7 @@ module RoutesIntoTeaching
 
     before_action :set_page_title
     before_action :set_step_page_title, only: %i[show update]
+    before_action :set_completed_page_title, only: [:completed]
     before_action :noindex, except: %i[index]
     before_action :set_breadcrumb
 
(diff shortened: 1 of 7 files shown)


gemmadallmandfe (approved):

PR #4654, opened 2025-03-11
Welcome guide - Missing Q&A text alternative

sarahcrack:
### Trello card

https://trello.com/c/oFy4C5hE/7616-gds-accessibility-audit-2-welcome-guide-missing-qa-text-alternative-level-a

### Context

On the Welcome to teaching page, in the “04 You’re not the only one with questions” section, there is a Q&A. When using a screen reader, the ‘Q’ is announced before the question text, but the ‘A’ is announced as “figure”. Screen reader users are not provided the same information as visual users.

- added `role="group"` to `<figure>` element which wraps the Answer portion of the Q&A, to help screen readers process that content as a group, without announcing unnecessary HTML tags. It should prevent the `"figure"` from being announced (this was the initial issue identified in the audit but I could not replicate so have left this 'fix' in in case it works!).
- removed the 'Q' and 'A' lettering by deleting the pseudo elements in the CSS file
- made each question an h3 - this will differentiate the question from the answer

### Guidance to review

- check with alternative screen reader tools / devices as I could not replicate the 'figure' annoucement on VoiceOver for Mac
- decide if we need to keep role="group" on the `<figure>` tag.

### DIFF
--- a/app/views/content/welcome/landing/_youre-not-the-only-one-with-questions.html.erb
+++ b/app/views/content/welcome/landing/_youre-not-the-only-one-with-questions.html.erb
@@ -66,7 +66,7 @@
           What's the salary like?
         </h3>
 
-        <figure class="a" role="group">
+        <figure class="a">
           <blockquote>I know some people worry about the pay, especially when they start out. But my career progression has been really fast. Becoming a teacher is worth it.</blockquote>
 
           <figcaption>Jacqui</figcaption>
@@ -78,7 +78,7 @@
           What’s the work balance like?
         </h3>
 
-        <figure class="a" role="group">
+        <figure class="a">
           <blockquote>I sometimes do a couple of hours at the weekend so I'm ready for the week ahead. But I try to keep my weekends free to make sure I have some real downtime.</blockquote>
 
           <figcaption>Sarah</figcaption>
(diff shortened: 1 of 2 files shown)


app/views/content/welcome/landing/_youre-not-the-only-one-with-questions.html.erb
```diff
@@ -62,35 +62,35 @@
     </div>
     <div class="questions-and-answers">
       <div class="question-and-answer">
-        <p class="q">
+        <h3 class="q">
```
gemmadallmandfe, line 65: I think the `h3` works well :-) 

If we were going to keep the Welcome guide, I'd suggest we clear up all the classes/figcaptions etc. but I don't think it is worth it now!

app/views/content/welcome/landing/_youre-not-the-only-one-with-questions.html.erb
```diff
@@ -62,35 +62,35 @@
...
           What's the salary like?
-        </p>
+        </h3>
 
-        <figure class="a">
+        <figure class="a" role="group">
```
gemmadallmandfe, line 69: Do we still need the `role="group"` @sarahcrack ? (Same comment for other instances)

gemmadallmandfe:
I've tested this on VoiceOver and it seems fine to me - I think the h3s work better than the Q and A letters

Have left a minor comment but otherwise I'm happy with this if @jenhadfield-dfe and @martyn-w are!

app/views/content/welcome/landing/_youre-not-the-only-one-with-questions.html.erb
sarahcrack: hi @gemmadallmandfe yes I'm not sure tbh - that was the original idea I had to try and prevent 'figure' from being announced but as we can't replicate the issue/test it, shall I just remove from all `figure` tags for now?

martyn-w (approved):
👍  LGTM

PR #4656, opened 2025-03-11
Remove frontdoor caching for webchat

martyn-w:
### Trello card
[Investigate broken webchat](https://trello.com/c/tRj9pTp7/7639-investigate-why-webchat-is-unavailable)

### Context
Frontdoor caching on the webchat was breaking things because it could not differentiate between json and html

Remove caching on the webchat


spencerldixon (approved):
LGTM 👍

PR #4663, opened 2025-03-13
Back link not accessed when tabbing through a funnel step

sarahcrack:
### Trello card

https://trello.com/c/UsO5oNLZ/7609-gds-accessibility-audit-28-back-link-not-accessed-when-tabbing-through-a-funnel-step-level-a

### Context

When going through the Personalised email guidance journey and the Adviser journey, the question pages have a “Back” link above each question. When tabbing through the page, the back link is skipped entirely. It is possible to reach this link by tabbing backwards but this may be counterintuitive to users and the option may be missed entirely.

- the backlink needs to be tabbable after the last navigation link but before the first form field (forward tabbing).
- but must stay outside of `#main` so that `skip to main content` still works well/as expected for screen reader users.

- Previously, the `handleMenuTab` function in `navigation_controller.js` would move focus directly to the main content area when no next item was found in the navigation, in effect bypassing the `back-link` which is placed in between the last nav item and main content. 
- Added `.tab-after-nav-menu` class to the backlink via `BackComponent` so should apply to all funnels where backlink is used.

### Guidance to review

- Check on Mac VoiceOver and TalkBack to ensure `Skip to main content` still works as expected. 
- Double check all funnels to ensure back-link is tabbable.

### DIFF
--- a/app/components/back_component.html.erb
+++ b/app/components/back_component.html.erb
@@ -1,4 +1,4 @@
-<div class="back-link">
+<div class="back-link tab-after-nav-menu">
   <%= link_to text, path, **options %>
 </div>
 <!-- This 1x1 blank image ensures that iOS VoiceOver reads the title in the main
--- a/app/views/layouts/registration.html.erb
+++ b/app/views/layouts/registration.html.erb
@@ -11,10 +11,10 @@
 
     <section class="main-section">
       <section class="feature registration container">
-      <%= yield(:back_link) %>
-      <%= main_tag(class: "within-section") do %>
-        <%= yield %>
-      <% end %>
+        <%= yield(:back_link) %>
+        <%= main_tag(class: "within-section") do %>
+          <%= yield %>
+        <% end %>
       </section>
     </section>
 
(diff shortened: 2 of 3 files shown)


gemmadallmandfe:
@sarahcrack I can see that this works technically - the back link can now be tabbed to in either direction

However, I don't think this is right in terms of accessibility - sorry!

The #main section should start with the `h1` on the page so screen reader users can skip straight to this when selecting the `Skip to main content` link. However, in the review app, the `Skip to main content` link now lands them on `Back` which would be really confusing?

Looking at the example pattern in the design system, the back link should definitely come before the main:
https://design-system.service.gov.uk/patterns/question-pages/

image

gemmadallmandfe:
See https://github.com/DFE-Digital/get-into-teaching-app/pull/3779 - we moved the back link outside main previously

gemmadallmandfe (approved):
Looks good to me from a front-end perspective - works well in the funnels as the back link can be tabbed to after the menu, but is skipped if the user selects `Skip to main content`

I've also had a look at a few non-funnel pages as well

Needs a tech review before merging though!

martyn-w:
Rather than adding additional logic to decide whether to tab to the backlink or to the main section, why not just add the `.tab-after-nav-menu` class to the backlink?

martyn-w (approved):
👍 LGTM

gemmadallmandfe (approved):

PR #4669, opened 2025-03-17
Remove adviser promo from iQTS page

gemmadallmandfe:
### Trello card
https://trello.com/c/zfz2t8qA

### Context
Candidate support have agreed that iQTS is not a route supported by the adviser service


Sarah-DfE (approved):
Looks good to me

PR #4429, opened 2024-12-11
Revert "Fixes issue with wrapping on section containers that cause th…

martyn-w:
…e content to overlap the footer at certain breakpoints (#4427)"

This reverts commit 63debf4453b5b58be7a924c6e608b2df086e15f5.

PR #4427, opened 2024-12-11
Fixes issue with wrapping on section containers

spencerldixon:
### Trello card

https://trello.com/c/Jd9HvLCS/6811-phase-5-repurpose-navigation-component-for-life-as-a-teacher-section?filter=member:spencerldixon

### Context

When collapsing the page between tablet and mobile breakpoints, the content overlaps the footer.

### Changes proposed in this pull request

- Removes `flex-wrap` from section container class

### Guidance to review

Also found on the following pages, please check that these still work as expected:

- `life-as-a-teacher/_how-to-become-a-teacher-cta.html.erb`
- `life-as-a-teacher/change-careers/_what-teachers-have-to-say.html.erb`
- `life-as-a-teacher/pay-and-benefits/_what-teachers-have-to-say.html.erb`
- `life-as-a-teacher/teaching-as-a-career/_what-teachers-have-to-say.html.erb`

PR #4686, opened 2025-03-21
Fix overflow issue with text on events page

spencerldixon:
### Trello card

https://trello.com/b/lRXllOHK/git-website-sprint-board

### Context

On the Find an event near you or online page, some content is cut off the screen at 400% zoom. This includes some of the bullet points under “Events” and the text under “Do you have a teaching event?”. Moving the page view to see these items results in header items being cut off instead.

- Removes flex wrap from section container

### Guidance to review

Seems to fix the issue but I'm not sure if this affects anything else. Please test thoroughly!

### DIFF
--- a/app/webpacker/styles/layout.scss
+++ b/app/webpacker/styles/layout.scss
@@ -68,7 +68,6 @@ body {
 section.container {
   display: flex;
   flex-grow: 0;
-  flex-wrap: wrap;
   flex-direction: column;
   margin: 1em 0;
 
--- a/app/webpacker/styles/layout.scss
+++ b/app/webpacker/styles/layout.scss
@@ -68,8 +68,10 @@ body {
 section.container {
   display: flex;
   flex-grow: 0;
+  flex-wrap: wrap;
   flex-direction: column;
   margin: 1em 0;
+  word-break: break-word;
 
   > main.within-section,
   > .back-link {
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
@spencerldixon I don't think this is the answer, sadly!

You've tried this before: https://github.com/DFE-Digital/get-into-teaching-app/pull/4427

And we had to revert it immediately as the site went wonky: https://github.com/DFE-Digital/get-into-teaching-app/pull/4429

cc @martyn-w @jenhadfield-dfe

app/webpacker/styles/layout.scss
```diff
@@ -71,6 +71,7 @@ section.container {
   flex-wrap: wrap;
   flex-direction: column;
   margin: 1em 0;
+  word-break: break-word;
```
martyn-w, line 74: this looked like a good solution to the problem.... unfortunately the setting `word-break: break-word;` is deprecated so we shouldn't use it in a new implementation. 

Have a look at `overflow-wrap: break-word;` instead?

martyn-w (changes requested):
This setting is deprecated; have a look at `overflow-wrap: break-word;` instead

martyn-w (approved):
👍 LGTM

(but please get a review from Gemma/Jen too)

gemmadallmandfe (approved):
Looks good to me - I haven't found any issues generally with the site

I managed to replicate the original issue on the events page found by the accessibility auditor by changing my display settings to 1280x1024 and setting zoom to 400%. I could see the issue in staging but it is fixed in this review app.

However, I'm not convinced this has fixed the equivalent issue flagged on the bursaries and scholarship page!

I'd suggest we merge this on Monday and come back to the bursaries page @spencerldixon

PR #4689, opened 2025-03-27
Bump Rails to 7.2

sarahcrack:
### Trello card

https://trello.com/c/4aVlLTKy/7533-update-rails-to-version-72-for-git

### Context

The version of Rails we are using for GIT is very out of date. The latest version is 8.0.1. We want to upgrade it in stages.

TBC


sarahcrack:
FYI in case useful:

Error in mailing list and adviser components:
The main error that was coming through after bumping Rails to 7.2 was related to the mailing list and adviser funnels and a `NoMethodError` e.g.
```
Content::AdviserComponent with a different heading
     Failure/Error: <%= f.hidden_field :channel_id, value: f.object&.channel_id.presence || params[:channel] %>
     
     NoMethodError:
       undefined method 'channel_id' for false
```

Not an issue with Rails 7.1.

Rails 7.2 (with no model specified) / error in views
Terminal:
```
f.object: false
params[:channel]: nil
```

Rails 7.2 (including `model: nil` with `form_with`) / view works fine
Printed in browser/on page
```
DEBUG f.object: nil
DEBUG params[:channel]: nil
```

PR #4711, opened 2025-04-07
Adds degree status filter to the picklist presenter

spencerldixon:
### Trello card

https://trello.com/c/uqR0KCFq

### Context

Since we added a new option to the CRM, this has appeared in production as currently we don't filter down the picklist. We need to filter the picklist to just the options we want, fortunately, I made a presenter for just this occasion!

- Adds degree status to picklist api presenter to filter down picklist options
- Adds this to the degree status steps of both the mailing list and events wizards

### DIFF
--- a/app/models/events/steps/personalised_updates.rb
+++ b/app/models/events/steps/personalised_updates.rb
@@ -30,8 +30,7 @@ module Events
       end
 
       def degree_status_options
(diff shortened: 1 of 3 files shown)


martyn-w (approved):
👍  LGTM

PR #4728, opened 2025-04-10
add academic year value to student finance page

emmaachesongray:
### Trello card
https://trello.com/c/rFXcZkZ7/7732

### DIFF
--- a/app/views/content/funding-and-support/tuition-fee-and-maintenance-loans.md
+++ b/app/views/content/funding-and-support/tuition-fee-and-maintenance-loans.md
@@ -40,10 +40,10 @@ expander:
 inset_text:
   loan-increase:
     text: |-
-      From autumn 2025, the maximum fees that providers can charge for postgraduate teacher training courses have increased. This means the amount you can apply to borrow to cover tuition fees has also increased. The increased amounts are shown on this page.
+      For the <%= v :finance_academicyear %>, the maximum fees that providers can charge for postgraduate teacher training courses have increased. This means the amount you can apply to borrow to cover tuition fees has also increased. The increased amounts are shown on this page.
   maintenance-increase:
     text: |-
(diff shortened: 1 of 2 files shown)


app/views/content/funding-and-support/tuition-fee-and-maintenance-loans.md
```diff
@@ -40,10 +40,10 @@ expander:
 inset_text:
   loan-increase:
     text: |-
-      From autumn 2025, the maximum fees that providers can charge for postgraduate teacher training courses have increased. This means the amount you can apply to borrow to cover tuition fees has also increased. The increased amounts are shown on this page.
+      For the <%= v :finance_academicyear %>, the maximum fees that providers can charge for postgraduate teacher training courses have increased. This means the amount you can apply to borrow to cover tuition fees has also increased. The increased amounts are shown on this page.
```
gemmadallmandfe, line 43: This (and the other banner) hasn't worked:
image

gemmadallmandfe: However, I'm not sure this is the right approach

The banners will be deleted before next cycle - they are temporary. Once this happens, there will be no reference to academic year on the page at all.

I would maybe suggest adding a new line somewhere between line 49 and 64 that will be permanently on the page - and then leaving the banners as is for now?

config/values/finance.yml
```diff
@@ -1,4 +1,5 @@
 finance:
+  academicyear: 2025 to 2026 academic year
```
gemmadallmandfe, line 2: I don't know whether all of these financial amounts are updated at the same time (although they might well be) - if we have 25/26 figures for all of them now, this is OK!

gemmadallmandfe (approved):
Looks good to me

PR #4735, opened 2025-04-14
Implement rate limiting for the prod website

RMcVelia:
### Trello card

https://trello.com/c/lMG5Q82M/2313-rate-limiting-for-git

### Context

We've seen an increase in website scanning which can cause failures for normal users.
This PR implements rate limiting at the front door ingress, using the rate_limiting option that was previously tested and added to the domains terraform module. 
It sets max requests per 5 minute period for a single IP address.
Looking at previous stats I don't see any real usage of more than 200-300 requests, but we will start by limiting to 500 requests per 5 minute intervals and can tune further.

Update terraform configuration

### Guidance to review

make production domains-plan (adds new resources)
make test domains-plan (no change)


martyn-w (approved):
👍 LGTM

PR #4741, opened 2025-04-15
Remove search from front end

martyn-w:
### Trello card
[Remove search from front end](https://trello.com/c/eiCKsGAD/7727-remove-search-from-the-front-end-of-the-site)

### Context
Search has been replaced by an A-to-Z site index and is no longer required

Removes search throughout the site


app/components/footer_component.html.erb
```diff
@@ -82,7 +82,6 @@
     </div>
     <div class="site-footer-bottom">
       <div class="site-footer-bottom__links-container">
-        <%= link_to("Search", search_path) %>
```
gemmadallmandfe, line 85: Could you possibly shift the A-Z index so it appears first in the list (replacing Search) rather than last @martyn-w ? Thanks!

app/webpacker/styles/icons.scss
```diff
@@ -92,13 +92,6 @@
   height: 30px;
 }
 
-.icon-search {
```
spencerldixon, line 95: We should either remove the svg or restore this class if we want to keep it around for future use. I can't see it being used anywhere but `icon-search-black` is used.

jenhadfield-dfe: the black version may be used again in the future, but not this version

spencerldixon (approved):
LGTM 👍

gemmadallmandfe (approved):
Looks good to me

PR #4759, opened 2025-04-25
Fix opengraph tags

spencerldixon:
### Context

Opengraph images were not working on the `/trainees/got-into-teaching` page

- Changes our metatags helper to use `property` instead of `name` for opengraph tags as required by the spec
- Adds a `og:url` tag for the original url of the page as required on all pages by the spec


carlosmartinez (approved):

mariosh-e (approved):
Testing with debugger and worked all ok. Will check a couple of other pages once merged.

PR #4760, opened 2025-04-25
Use venue address for event map

spencerldixon:
### Trello card

https://trello.com/c/ylR7Hrwf

### Context

We want to update the Open venue map in Google mapsto use venue name as well as event address

- Use venue address instead of building address in events helper


carlosmartinez (approved):

gemmadallmandfe (approved):

PR #4765, opened 2025-04-30
Add english case study to why teach section

emmaachesongray:
### Trello card
https://trello.com/c/0tFrZs1q/7862


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_01_18_000000) do
+ActiveRecord::Schema[7.2].define(version: 2024_01_18_000000) do
```
gemmadallmandfe, line 13: @emmaachesongray Please check this change with Martyn before it goes out

app/views/content/life-as-a-teacher/why-teach/_what-teachers-have-to-say.html.erb
```diff
@@ -18,11 +18,20 @@
...
                 OpenStruct.new(
                   title: "Abigail's favourite things about teaching",
-                  description: "Hear what Abigail enjoys most about teaching, and the impact she has through her job.   ",
+                  description: "Hear what Abigail enjoys most about teaching, and the impact she has through her job.",
                   path: "/life-as-a-teacher/why-teach/abigails-favourite-things-about-teaching",
                   image: "static/images/content/case-studies/abigail.jpg"
                 )) %>
             </ul>
+            <ul>
+              <%= render Content::InspirationCardComponent.new(heading_tag: "h3", card:
```
gemmadallmandfe, line 27: Should we put this above Ben/Abigail? More interesting and new for most people?

gemmadallmandfe:
@emmaachesongray @mariosh-e

* The page design is consistent with others on the site (eg `/life-as-a-teacher/change-careers/becoming-a-parent-sparked-my-interest-in-teaching`) so I would suggest not looking at that as part of this PR (and also remember that most people are on mobile)

* The nav cards do look odd but I would suggest the quickest change is to move this case study back down to third position now that it does not have an image

db/schema.rb
martyn-w: this change is fine to keep in 👍

PR #4783, opened 2025-05-09
Better table formatting at high zoom

martyn-w:
### Trello card
[Reflow issue on bursaries page](https://trello.com/c/VPfsc5aM/7610-gds-accessibility-audit-21-reflow-issue-on-bursaries-page-level-aa)

### Context
The table of scholarships and bursaries on the bursaries page was pushing the page width out at high zoom (400% @ 1280px)

Changes the word-wrap settings for table headings

### Guidance to review

Test at normal zoom and at high zoom. Also test the cookies page which is subtly changed by this PR.

### DIFF
--- a/app/webpacker/styles/table.scss
+++ b/app/webpacker/styles/table.scss
@@ -15,6 +15,7 @@ table {
       margin-bottom: $table-spacing;
       border-bottom: 1px solid $grey-light;
       word-break: normal;
+      white-space: nowrap;
 
       &:first-child {
         padding-left: 0;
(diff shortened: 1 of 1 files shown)


gemmadallmandfe:
The table looks better @martyn-w but the rest of the page still has the same issue with body text needing horizontal scroll - so I don't think this has resolved the reflow issue?

image

carlosmartinez (approved):

martyn-w:
> The table looks better @martyn-w but the rest of the page still has the same issue with body text needing horizontal scroll - so I don't think this has resolved the reflow issue?
> 
> image

@gemmadallmandfe and I had a screen-share and found her system was set to scale at 150%, so the zoom level was effectively 600% not 400%.

gemmadallmandfe (approved):

We found that the issues I was still experiencing on my laptop at 400%/1280x1024 were due to an additional Windows setting which scaled the content a further 150% (ie 600% in total). When this was changed back to 100% (so 400% in total), the content wrapped properly on the screen

PR #4785, opened 2025-05-09
Update robots.txt for facebook 

martyn-w:
### Trello card
[Ensure image is pulled through when sharing on Facebook](https://trello.com/c/nbzQNVXp/7834-investigate-why-social-media-celebration-image-is-not-pulling-through-to-facebook-social-share)

### Context
Facebook requires an explicit share in the robots.txt file

Updates the robots.txt file

### Guidance to review
Test via Facebook share debugger: https://developers.facebook.com/tools/debug/


mariosh-e:
Got into Teaching image is now pulling through successfully. Other pages such as How to become a teacher, subject pages are pulling through too. Will do a full review on Monday.

martyn-w:
(comment sent via slack, here for posterity):

If I send a POST to `/api/teacher_training_adviser/candidates` which includes a `graduationYear` (e.g. 2028) and a `degreeStatusId` (e.g. 222750002), it returns an OK status with the inferred `degreeStatusId`, which doesn’t necessarily agree with the _supplied_ `degreeStatusId`, e.g.:

```
{
	"degreeStatusId": 222750003
}
```

Moreover, if I don’t supply a `degreeStatusId` it complains with a `400 Bad Request` validation error, saying degree status is required..

But this behaviour seems a bit odd… if `degreeStatusId` is required, why is it ignored? and if it shouldn’t be mandatory, can you clarify what the inference logic should be?

asatwal (approved):
👍 LGTM

PR #4790, opened 2025-05-12
Add guidance about transcripts

emmaachesongray:
### Trello card
https://trello.com/c/WzWSmxg2/7797


gemmadallmandfe:
Looks good to me @emmaachesongray 

One suggestion - is it worth adding something about where we are now creating transcript files - are they all in the same folder?

mariosh-e (approved):
LGTM

PR #4791, opened 2025-05-13
Update accessibility statement following reflow fix on bursaries page

gemmadallmandfe:
### Trello card
https://trello.com/c/qYkIdC6i/7901-update-accessibility-statement-following-reflow-fix-on-bursaries-page

### Context
We have fixed the reflow issue on the bursaries page that was identified in our accessibility audit, so we need to update our accessibility statement.


mariosh-e (approved):
LGTM

PR #4798, opened 2025-05-14
removed box styling around the call back module

Tamara-Lewis:
### Trello

https://trello.com/c/68Nqjyxe/7972

### DIFF
--- a/app/webpacker/styles/registration.scss
+++ b/app/webpacker/styles/registration.scss
@@ -71,12 +71,3 @@ section.registration {
       margin-top: 1em;
     }
   }
-
-    h2 {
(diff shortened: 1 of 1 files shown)


mariosh-e:
@martyn-w can you please check this please, removing the callback module, I think there is more that should be removed. Is it lines 75-82 or everything from 75-87? Thank you

Tamara-Lewis:
iv updated in VSC to include the H2 & p so it should be ok to go ahead @martyn-w

Tamara-Lewis:
merging looks to be blocked until someone else has reviewed it

PR #4838, opened 2025-06-17
Reduce verbosity of link to the DfE in the GIT footer

martyn-w:
### Trello card
[GDS Accessibility Audit: 1 - Footer logo missing destination (Level A)](https://trello.com/c/cAhVmNO9/8093-gds-accessibility-audit-1-footer-logo-missing-destination-level-a)

### Context
The description of the link to the DfE was too verbose

Reduce verbosity

### Guidance to review
Test in a screen reader


mariosh-e:
I checked on staging and using WAVE the aria tag now does display Department for Education homepage (which didn't before). If you can run it pass JL on Thursday to double make sure that would be helpful and then we can push this through. Thank you

PR #4842, opened 2025-06-23
Still time to apply messaging for 24/25 cycle

emmaachesongray:
### Trello card
https://trello.com/c/Pgqrc2tB/8104-add-still-time-to-apply-content-and-messaging-for-24-25-cycle


mariosh-e:
SG happy to be pushed out

PR #4869, opened 2025-07-10
delete unused graphics

aabdi18:
removal of the listed grapics and icons on ticket #8134

### Context
We want to review the graphics / icons in GitHub and delete any old ones that are no longer used and won’t be in the future


martyn-w (approved):
👍 LGTM

PR #4891, opened 2025-08-13
Run applicaton as non-root user

shaheislamdfe:
### Trello card

https://trello.com/c/2zS5P78I/2431-security-aks-containers-run-as-non-root-user-gitapp

### Context

  Security fix to ensure containers run as non-root user instead of root.

  - Updated Dockerfile to create and use a non-root user (UID 10001)
  - Added run_as_non_root = true to Terraform web and worker modules
  - Set proper file ownership in Docker image builds

### Guidance to review

https://github.com/DFE-Digital/get-into-teaching-app/actions/runs/16951496892/job/48045282251

  - Check the app starts without permission errors
  - Verify migrations run successfully
  - Test file uploads and logs still work

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -81,6 +81,9 @@ ENV RAILS_ENV=production \
     RAILS_LOG_TO_STDOUT=true \
     RACK_TIMEOUT_SERVICE_TIMEOUT=60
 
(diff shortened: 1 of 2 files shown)


martyn-w (approved):
👍  LGTM

PR #5095, opened 2026-04-20
GIT Rebrand main PR

martyn-w:
### Trello cards
* [8934 Rebrand: Icon colour](https://trello.com/c/5PasEfgS/8934-rebrand-icon-colour)
* [8935 Rebrand: Heading colours](https://trello.com/c/aHFwdz3i/8935-rebrand-headings-colour)
* [8938 Rebrand: Develop new homepage design](https://trello.com/c/oGKDTWkx/8838-rebrand-develop-new-homepage-design)
* [8950 Rebrand: Block quote colour](https://trello.com/c/boEvpMTB/8950-rebrand-block-quote-colour)
* [8951 Rebrand: Category card hover/click state](https://trello.com/c/ul5aH4Ta/8951-rebrand-category-card-hover-click-state)
* [8952 Rebrand: Non-uk inset box](https://trello.com/c/qwL16Ov8/8952-rebrand-non-uk-inset-box)
* [8953 Rebrand: Feature boxes](https://trello.com/c/vKNZ3aCH/8953-rebrand-feature-boxes)
* [8954 Rebrand: H1 & H2 font](https://trello.com/c/ilQAZ8sr/8954-rebrand-h1-h2-font)
* [8955 Rebrand: Headings colour - inspiration_card_component](https://trello.com/c/NzRAbzGR/8955-rebrand-headings-colour-inspirationcardcomponent)
* [8966 Rebrand: Favicon](https://trello.com/c/q5oH0F9d/8966-rebrand-favicon)
* [8995 Rebrand: Steps to become a teacher](https://trello.com/c/Ef7F5kbL/8995-rebrand-steps-to-become-a-teacher-specific-changes)
* [8996 Rebrand: Icon colour/image change](https://trello.com/c/oiU62qVO/8996-rebrand-icon-colour-image-change)
* [8997 Rebrand: Landing page banners](https://trello.com/c/BciAm4oB/8997-rebrand-landing-page-banners)
* [8998 Rebrand: Grow landing page ](https://trello.com/c/O0qrARCO/8998-rebrand-grow-landing-page-specific-changes-tbc)
* [9023 Rebrand: Cat 1 pages](https://trello.com/c/DObNvbii/9023-rebrand-level-1-category-pages)
* [9024 Rebrand: Cat 2 pages](https://trello.com/c/sMTGFtzU/9024-rebrand-level-2-sub-category-pages)
[...]

PR #5096, opened 2026-04-21
8955 rebrand headings colour inspirationcardcomponent

martyn-w:
### Trello card
[Rebrand headings in inspirational card component](https://trello.com/c/NzRAbzGR/8955-rebrand-headings-colour-inspirationcardcomponent)

### Guidance to review

NB: merges into 8934 https://github.com/DFE-Digital/get-into-teaching-app/pull/5095


app/components/content/inspiration_card_component.html.erb
```diff
@@ -1,7 +1,7 @@
 <% if image? %>
   <!-- This <li> element is dynamically generated within a parent template that includes the surrounding <ul>. -->
   <!-- //NOSONAR --><li class="category__nav-card category__nav-card__with-image category__nav-card__white">
-    <%= link_to(path, class: "link--pink") do %>
+    <%= link_to(path, class: "link--gitlilac") do %>
```
dcyoung-dev, line 4: Is it worth removing the colour name from this class? That way it wouldn't need to change when there's a rebrand?

dcyoung-dev (approved):

app/components/content/inspiration_card_component.html.erb
martyn-w: I take the point - though in this case, these links are different from the standard govuk-blue links used everywhere else, so I think its clearer if explicit here

PR #5146, opened 2026-06-10
Updated the Marketing cookies

hannahvb:
I've updated the table in this section to include Tiktok and last updated date to reflect today's date.

### Trello card
https://trello.com/c/66chMPv1/9087-update-cookie-policy-for-rebrand

### Context
We needed to add TikTok to the 'Marketing cookies we use' table.


Soycoy (approved):
LGTM

PR #5152, opened 2026-06-22
Use sonar.token property in sonar-scanner

martyn-w:
Fix the SonarCloud authentication option in the GitHub Actions workflow by replacing -Dsonar.login with -Dsonar.token in .github/workflows/build.yml. This ensures the scanner uses the SONAR_TOKEN secret correctly and prevents authentication failures during analysis.


hopep-mobilise (approved):
Implemented previously in gitapi

PR #5153, opened 2026-06-23
Training provider events wizard

martyn-w:
### Trello card
[Training provider events: Wizard build - Step 1 Application link](https://trello.com/c/dlnLdOJ3/9124-training-provider-events-wizard-build-step-1-application-link)

### Changes proposed in this pull request
Update copy, tests and URLs on `/events` page

PR #5172, opened 2026-07-17
9116 training provider events wizard build step 9 time date

martyn-w:
### Trello card
[Training Provider Events step 9 - event date times](https://trello.com/c/aSr524CM/9116-training-provider-events-wizard-build-step-9-time-date)

Adds step 9 to capture the event date and start/end times
Upgrades the govuk_design_system_formbuilder to 6.4 to use the time capturing field
Handles multi-param time values manually as they are not supported in ActiveModal (only ActiveRecord)

### Guidance to review
NB: this merges into feature branch https://github.com/DFE-Digital/get-into-teaching-app/pull/5153


app/views/provider_events/steps/_event_date_times.html.erb
```diff
@@ -0,0 +1,17 @@
+<%= tag.h1 t("helpers.label.provider_events_steps_event_date_times.title") %>
+
+<%= f.govuk_text_field(
```
dcyoung-dev: There is a `govuk_time_field` available - not sure if this helps

dcyoung-dev (approved):

app/views/provider_events/steps/_event_date_times.html.erb
martyn-w: ah yes, I've just discovered that and am re-doing this PR to take advantage of it - thanks

app/models/provider_events/steps/event_times.rb
```diff
@@ -0,0 +1,69 @@
+module ProviderEvents
+  module Steps
+    class EventTimes < ::GITWizard::Step
+      include FunnelTitle
+
+      # NB: multi parameter date fields aren't yet supported by ActiveModel so
+      # we are doing things a bit manually here
```
dcyoung-dev, line 7: I've seen this problem before [`ActiveRecord::AttributeAssignment`](https://dcyoung.dev/shorts/using-a-datetime_select-with-a-form-object/) will save the day.

Gemfile.lock
dcyoung-dev, line 1: Should this be updated as part of this PR?

dcyoung-dev (approved):

Gemfile.lock
martyn-w: in this case, yep, as it was necessary to upgrade the GovUK Forms to use the new time widget

app/models/provider_events/steps/event_times.rb
martyn-w: so I did try this approach, but unfortunately the parsing of times here is a bit more subtle than the implementation taken in `ActiveRecord::AttributeAssignment`, so I'll stick with the manual method here

PR #5183, opened 2026-08-04
Update salary figures and effective date

hannahvb:
We have updated these based on figures provided by the pay team.

### Trello card
https://trello.com/c/XSMFQdej/9186-update-pay-salary-values-to-match-new-pay-award-1st-sept

### Context
We need to update the salary amounts in line with the pay award announcement. These figures are subject to consultation so we will be making this clear via banners, as we have done in previous years.


martyn-w (dismissed):
👍 LGTM (for tech review)

martyn-w (approved):
👍 LGTM

PR #5187, opened 2026-08-06
Cache chat availability and add client refresh jitter

martyn-w:
### Trello card
[Investigate and resolve chat outage](https://trello.com/c/COyW7RQK/9249-investigate-and-resolve-chat-outage-on-the-help-and-support-page)

Cache the chat availability API response in Rails.cache (key: "chat_availability_api") for 10 seconds to reduce upstream load and keep the existing Faraday::ConnectionFailed handling. Add a small random jitter (0–3000ms) to the client-side polling interval in the chat controller to smooth out synchronised refresh spikes.

### DIFF
--- a/app/models/chat.rb
+++ b/app/models/chat.rb
@@ -32,8 +32,11 @@ private
 
     @availability ||=
       begin
-        response = Faraday.get(availability_api_uri)
-        JSON.parse(response.body) if response.success?
+        # Cache the response for a few seconds to reduce upstream server traffic
+        Rails.cache.fetch("chat_availability_api", expires_in: 10.seconds) do
+          response = Faraday.get(availability_api_uri)
+          JSON.parse(response.body) if response.success?
+        end
       rescue Faraday::ConnectionFailed
         nil
       end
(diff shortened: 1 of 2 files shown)


app/models/chat.rb
```diff
@@ -32,8 +32,11 @@ def availability
...
       begin
-        response = Faraday.get(availability_api_uri)
-        JSON.parse(response.body) if response.success?
+        # Cache the response for a few seconds to reduce upstream server traffic
+        Rails.cache.fetch("chat_availability_api", expires_in: 10.seconds) do
+          response = Faraday.get(availability_api_uri)
+          JSON.parse(response.body) if response.success?
+        end
       rescue Faraday::ConnectionFailed
         nil
```
dcyoung-dev, line 41: I'm not sure if this would create a `nil` cache entry if it fails.
In the past I've found it easier to manually access the cache - checking the cache first then hitting the API and storing in the cache if successful.

dcyoung-dev (approved):

app/models/chat.rb
martyn-w: Ah perhaps - but then its only going to cache it for 10 seconds, so at the very worst the chat would be delayed by 10 seconds

PR #5215, opened 2026-09-16
Add conditional renderer component

dcyoung-dev:
### Trello card

Part of https://trello.com/c/x8qsr3w4

### Context

We want to be able to render content based on timings. This component aims to set out the ability to render different content depending on the date. 

### Guidance to review

This PR only includes adding the component and documentation for using the component, it does not use the component to render content at this point.


martyn-w (approved):
👍 LGTM

PR #5219, opened 2026-09-17
Improved the Timed content component

dcyoung-dev:
### Trello card

Part of https://trello.com/c/x8qsr3w4/9295-update-fi-scholarship-bursary-information
Fixes implementation of #5215 

### Context

Our content pages are Markdown. Some pages use a `$token$` to drop in a component — like the timed financial content that shows a different branch (2025 / 2026 / default) depending on the date.

The Markdown template handler does not return HTML. It returns a **string of Ruby code**. Rails runs that code later to build the page. So the handler runs at *compile time*, and the code it returns runs at *request time*.

The old handler rendered these `$token$` components **at compile time** and baked the finished HTML into that string. A baked component is frozen. It cannot read the request. So it never saw `?now=`. And it never re-checked the date. It only changed when the Ruby process restarted. That is why the timed content was stuck on one branch.

We now leave the timed component as a **marker** at compile time, and turn it into a live `render` call that runs on **every request**.

The flow:

1. **Mark, don't bake.** When the handler hits a `timed_financial_content` token, it does not render it. It records the component and drops a placeholder into the Markdown: `<div data-dynamic-component="0"></div>`. We chose this exact `<div>` because Kramdown, Rinku and the table-caption step all leave it untouched — they don't escape it or wrap it in a paragraph. So the surrounding markup stays identical to before.
2. **Build the page as normal.** Kramdown turns the Markdown into HTML. The marker survives inside it.
3. **Split and stitch.** `compile_body` splits the finished HTML on those markers. The literal HTML pieces stay as baked strings. Each marker becomes a live `render(...)` call. It joins them all with `safe_join([...])`.
4. **Run per request.** Rails runs that joined code on every request. The `render` calls run inside the real page. So the component can read `params` (`?now=`), and it re-checks the date every time. The branch now switches by itself at the configured date — no redeploy needed.

#### One rule for future maintainers

The marker string (in `runtime_component_marker`) and the split pattern (in `compile_body`) must match each other. They are a pair. Change one, change the other.

> **Note:** only components listed in `RUNTIME_COMPONENT_TYPES` take this per-request path. Every other component type is still baked once at compile time, exactly as before.

#### Reviewing locally: overriding `now` and `branch`

These helpers let you see each dated content branch without changing your clock or the config. **They only work in local environments (development and test) — they are ignored in production** (`Rails.env.local?`).

Use them on any page that renders timed financial content, e.g. `/funding-and-support/scholarships-and-bursaries`.

**Override the date** — pretend it's a given day, and let the normal date logic pick the branch:

```
/funding-and-support/scholarships-and-bursaries?now=2026-11-01   → 2026 branch
/funding-and-support/scholarships-and-bursaries?now=2026-06-01   → 2025 branch
/funding-and-support/scholarships-and-bursaries?now=2026-09-30   → default (the gap)
```

**Force a branch directly** — skip the date logic and render a named branch:

```
/funding-and-support/scholarships-and-bursaries?branch=2026
/funding-and-support/scholarships-and-bursaries?branch=2025
/funding-and-support/scholarships-and-bursaries?branch=default
```

**Precedence:** `?branch=` wins over `?now=`, which wins over the real date. A branch name that isn't configured falls back to `default`. A malformed date (e.g. `?now=broken`) is ignored and falls back to the real date.

**To confirm the production gate:** with `Rails.env.local?` false, all three overrides are ignored and the real date is used. This is covered by the `"only in local environments"` tests in `spec/components/content/timed_financial_content_component_spec.rb`.


app/components/content/timed_financial_content_component.rb
```diff
@@ -13,31 +15,78 @@ def self.content_dates
...
+
+    # An explicit branch override wins over the date-based selection, so we can
+    # force a branch for debugging - e.g. ?branch=2026 - without moving the clock.
+    def selected_branch_key
+      override_branch_key || override_now_key || branch_key(now)
+    end
+
+    # Debug overrides are only honoured in local (development/test) environments.
+    def overrides_enabled?
+      Rails.env.local?
```
martyn-w: I think this will also be handy in review apps, so maybe relax this condition to:
```
!Rails.env.production?
```

martyn-w (dismissed):
👍 LGTM

martyn-w (dismissed):
👍 LGTM

martyn-w (approved):

